PR #26, opened 2017-09-02
Product categories format fix

jancinert:
screen shot 2017-09-02 at 10 00 57 pm


dmitridr:
Thanks for sending us a pull request! We'll patch it and test it ASAP to see if we can merge.

dmitridr:
Tested and repro-ed the problem last week. This is a good fix, we've merged it and will call it out in the next release. Thanks @jancinert !!

PR #86, opened 2017-10-13
Add to cart and ajax add to cart

### LINKED ISSUE #3: Add to cart pixel not fired when item is added to cart from product page
The add to cart pixel is only fired when a user visits the cart page and `woocommerce_after_cart` is triggered. This misses the event on the single product page. It's also possible that a user might go straight to checkout (or be redirected straight to checkout) and so you'd miss the add to cart action all together.

### LINKED ISSUE #4: Allow multiple product purchase on Facebook Page or option to disable add to cart URL
Customer is looking to sell products on their Facebook page but is wanting to not have the redirect to checkout

> Is there any way we can amend this to send the customer to the product page rather than straight to checkout please ?

Would this be possible in a future release to have settings to change the `Check Out on Website` button / link to either "Continue shopping" or "Browse more on our site" or similar?

helgatheviking:
Fires pixel on add to cart hook (instead of viewing cart) and ajax add to cart. closes #3 and closes #4.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -184,7 +184,7 @@ class WC_Facebookcommerce_EventsTracker {
 
     $pixel = ob_get_clean();
 
-    wp_send_json( $pixel );
+    wp_send_json($pixel);
   }
 
   /**
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -163,6 +163,31 @@ class WC_Facebookcommerce_EventsTracker {
   }
 
   /**
+  * Triggered by add_to_cart jquery trigger
+  */
+  public function inject_ajax_add_to_cart_event() {
+    ob_start();
+    
+    echo '<script>';
+    
+    $product_ids = $this->get_content_ids_from_cart(WC()->cart->get_cart());
+
+    echo $this->pixel->build_event(
+      'AddToCart',
+      array(
+        'content_ids' => json_encode($product_ids),
+        'content_type' => 'product',
+        'value' => WC()->cart->total,
+        'currency' => get_woocommerce_currency()
+      ));
+    echo '</script>';
+
+    $pixel = ob_get_clean();
+
+    wp_send_json( $pixel );
+  }
+
+  /**
    * Triggers InitiateCheckout for checkout page
    */
   public function inject_initiate_checkout_event() {
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -38,11 +38,11 @@ fbq('track', 'PageView', %s);
 <!-- Support AJAX add to cart -->
 if(typeof jQuery != 'undefined') {
   jQuery(document).ready(function($){
-    jQuery( 'body' ).on( 'added_to_cart', function( event ) {
+    jQuery('body').on('added_to_cart', function(event) {
 
       // Ajax action.
-      $.get( '?wc-ajax=fb_inject_add_to_cart_event', function( data ) {
-        $('head').append( data );
+      $.get('?wc-ajax=fb_inject_add_to_cart_event', function(data) {
+        $('head').append(data);
       });
       
     });
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -34,6 +34,22 @@ t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
 document,'script','https://connect.facebook.net/en_US/fbevents.js');
 %s
 fbq('track', 'PageView', %s);
+
+<!-- Support AJAX add to cart -->
+if(typeof jQuery != 'undefined') {
+  jQuery(document).ready(function($){
+    jQuery( 'body' ).on( 'added_to_cart', function( event ) {
+
+      // Ajax action.
+      $.get( '?wc-ajax=fb_inject_add_to_cart_event', function( data ) {
+        $('head').append( data );
+      });
+      
+    });
+  });
+}
+<!-- End Support AJAX add to cart -->
+
 </script>
 <!-- DO NOT MODIFY -->
 <!-- End Facebook Pixel Code -->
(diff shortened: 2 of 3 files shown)


dmitridr:
Thanks for the contribution @helgatheviking !!! Our team will review and test this, and if all is ok, we'll merge it.

dmitridr:
@facebook-github-bot import

dmitridr:
Hi @helgatheviking , could you please sign the CLA as specified by the bot? We unfortunately aren't allowed to merge this until that is done.

helgatheviking:
Done.

facebook-commerce-events-tracker.php
```diff
@@ -163,6 +163,31 @@ public function inject_add_to_cart_event() {
   }
 
   /**
+  * Triggered by add_to_cart jquery trigger
+  */
+  public function inject_ajax_add_to_cart_event() {
+    ob_start();
+    
+    echo '<script>';
```
dmitridr: Is it possible to do this without `echo` and `ob_start`, and just to append strings? e.g. `$pixel = $this->pixel_>build_event(...` , we would prefer that, could you make that change?... or is there a reason for not doing so?

facebook-commerce-events-tracker.php
```diff
@@ -163,6 +163,31 @@ public function inject_add_to_cart_event() {
...
+        'content_ids' => json_encode($product_ids),
+        'content_type' => 'product',
+        'value' => WC()->cart->total,
+        'currency' => get_woocommerce_currency()
+      ));
+    echo '</script>';
+
+    $pixel = ob_get_clean();
+
+    wp_send_json( $pixel );
```
dmitridr: Could you remove spaces around variables in this way? This will be consistent with the rest of our codebase and pass our lints.

facebook-commerce-pixel-event.php
```diff
@@ -34,6 +34,22 @@ public function pixel_base_code() {
 document,'script','https://connect.facebook.net/en_US/fbevents.js');
 %s
 fbq('track', 'PageView', %s);
+
+<!-- Support AJAX add to cart -->
+if(typeof jQuery != 'undefined') {
+  jQuery(document).ready(function($){
+    jQuery( 'body' ).on( 'added_to_cart', function( event ) {
```
dmitridr: Same comment here, could you remove the spaces around `'body'` and `event` as well as in front of `'added_to_cart`?

facebook-commerce-pixel-event.php
```diff
@@ -34,6 +34,22 @@ public function pixel_base_code() {
...
+
+      // Ajax action.
+      $.get( '?wc-ajax=fb_inject_add_to_cart_event', function( data ) {
```
dmitridr: Same comment here.

facebook-commerce-pixel-event.php
```diff
@@ -34,6 +34,22 @@ public function pixel_base_code() {
...
+        $('head').append( data );
```
dmitridr: And here.

facebook-commerce.php
```diff
@@ -210,8 +210,10 @@ public function __construct() {
         array($this->events_tracker, 'inject_view_category_event'));
       add_action('pre_get_posts',
         array($this->events_tracker, 'inject_search_event'));
-      add_action('woocommerce_after_cart',
+      add_action('woocommerce_add_to_cart',
         array($this->events_tracker, 'inject_add_to_cart_event'));
+      add_action( 'wc_ajax_fb_inject_add_to_cart_event', 
+        array($this->events_tracker, 'inject_ajax_add_to_cart_event' ) );
```
dmitridr: Same comment about spaces here and on the line above.

facebook-commerce-events-tracker.php
helgatheviking: I can't go back and test extensively, but I don't think it worked. I *think* that `build_event()` just adds the script to the global `$wc_queued_js` which is printed in the footer. So I don't think we can use it b/c we need to dynamically inject code *after* the page is loaded.

helgatheviking:
@dmitridr Unrelated to this PR, but who would be the best person to talk to about how the products are displayed on Facebook? One of my plugins is Name Your Price (allows customers to enter their own price) and the price input doesn't display on Facebook, so when users click "add to cart" they get taken to the site but with an error message. Additionally, I'm the co-author of Mix and Match, but I haven't checked how FB is handling more complex product types. Thanks!

dmitridr:
@helgatheviking Regarding the add-to-cart link showing an error message, one option is to link users to the product page instead of the add-to-cart page, we added this as a setting within the shop settings about a month back.

Our team does not directly manage the shops interface, but there is a team responsible for that. If you have a Facebook representative that you talk to regularly, they may be able to convey your requests as a feature request to them. You can also file an issue on us, and we can pass it along to the shops team as a feature request. 

Realistically, it may be some time before they can address that request.

If there's some compatibility issue, we're happy to review more of your pull requests that support the plugins you've built, especially if they are widely used!

dmitridr:
Thanks for your patience and detailed contributions, this is a big feature for our project! 
We'll merge this and build on it to see if we can make the changes mentioned in the review and include it in the next release.

helgatheviking:
Thanks @dmitridr . I don't have a FB team rep to talk to at all, hence the message here asking who to speak to. I sent a message via a contact form somewhere and never heard back. I'm not personally having an issue b/c I don't have a store. But when people run my Name Your Price plugin and presumably also any complex product type that is added by plugin: Mix and Match, Product Bundles, Composite Products ( all of the ones mentioned in the WooCommerce.com marketplace and the customers expect official WooCommerce-supported products to work together) all those products require some sort of configuration before they can be added to the cart. In the case of Name Your Price, a price must be entered.

But a store may not want to *always* link to the product. Some products don't need configuration (simple products) and so those could be added directly. However, some products need configuration and trying to add them to the cart w/o said configuration produces an error. WooCommerce supports ajax add to cart on it's shop archive, but solves this issue by determining if the product supports "ajax add to cart". If Facebook did something similar on it's side I could make sure I am filtering the right thing and Name Your Price products would link to a product page, while simple products could still be added to cart. 

Should I create a separate issue?

PR #99, opened 2017-10-19
added missing closing tag

pwag42:
Without this tag, `<div id='facebook_options' class='panel woocommerce_options_panel'>` eats other elements within the admin, breaking any plugins that render their `woocommerce_options_panel` elements after it.


dmitridr:
Good catch, and thank you for the pull request!

Issue #57, opened 2017-09-27
Conflict with A2 Optimized Plugin


digirev:
Hi, after I install the plugin, I can see duplicated event in Pixel Helper. 1 of them shows me an error.

I tried with the plugin by PixelYourSite, it works well. Seems like there's something wrong with the plugin.

How can I fix or troubleshoot this further? Definitely not the cache since I cleared it already.

Thanks.

dmitridr:
Hello.

We don't seem to be able to reproduce this on our end.

- Could you tell us which version of the plugin you have?
- Could you share a screenshot or link to a site with the problem?

digirev:
It's version 1.6.1.
You can check it here: https://gizmosified.com/
It's still in a development phase. I turn it on for you to take a look for that Facebook Pixel Plugin issue. Please let me know if you have taken a look. I need to change it back to maintenance mode again.
Thanks.

dmitridr:
I took a look. Seems like a bug with the helper tool because the pixel fires once, correctly, the other duplicate fire is some kind of mis-parsing of your source by the pixel helper tool, interesting that PixelYourSite doesn't have this problem though.

If you [email us](ads_extension_woocommerce@fb.com) the source of your main page with PixelYourSite installed, perhaps we can figure out and fix the bug by looking at the difference. But I would not worry too much in the meantime, since this appears to be a problem with the pixel helper tool in some cases, and not the pixel itself.

digirev:
Hi, is this the best email ads_extension_woocommerce@fb.com ? I think couple days ago I sent an email there, but there's an automated email that I need to check to Facebook resource page which is not telling anything. I can send you the page source if you want so that you can apply the fix. Yeah that PixelYourSite plugin, instead of creating 2 Page Views, it creates 2 different event: PageView & (something else that I forgot).

Anyway, if you're 100% sure that it's only the bug in Pixel Helper, then I am going to use the plugin. Thanks.

digirev:
The 2 events that was created by PixelYourSite plugin is PageView & GeneralEvent, while your plugin created 2 PageView events (that's what it's shown on Pixel Helper).

I sent you the email as requested. Please check. Thanks.

dmitridr:
Hello,

We had to prioritize other issues above this one, but I finally had some time to take a look. Thank you for sending us your source, but unfortunately it seems that pixel your site uses a completely different approach that we won't be able to emulate.

[...]

PR #127, opened 2017-10-31
update block comment when HTML script are inlined

citymont:
when I use Fb-for-wc plugin with minify HTML plugins (eg. Cache Enabler), comment on HTML script block created an error. Just passing inline comment into block comment resolved the issue.


dmitridr:
Thanks for your contribution to our project!

This is a good fix! I think this may potentially solve Issue #57 as well. Give us ~24hrs to test and verify and we'll merge this.

PR #210, opened 2018-01-12
simple fix for bug #201

### LINKED ISSUE #201: W3 validation error : An img element must have an alt attribute.
After validating at w3.org I have the following error: An img element must have an alt attribute

Is there any way to fix it?

citymont:
Closes #201


dmitridr:
Thanks for sending this fix!! We'll test this out on our side then it should be good to merge.

dmitridr:
@citymont Sorry for the delay, we had a bit of trouble with our infrastructure, as you can probably tell from the bot's comments above, but we finally got that sorted out. Thanks for your contribution!!!

PR #252, opened 2018-02-01
Fixes #251

### LINKED ISSUE #251: New "facebook_for_woocommerce_integration_pixel_enabled" filter applied to early
Hello,
I'm PixelYourSite developer and trying to integrate `facebook_for_woocommerce_integration_pixel_enabled` filter. The filter is great itself, but it is applied too early, when other plugins, like ours, can be not ready yet. I'm going to make pull request which fixes the issue.

mister:
Static `apply_filters` method moved from constructor to `wp_head` action which allows 3rd party setup filters before they applied. Priority 9 used to make sure that action will be fired before rest of class actions.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -19,7 +19,31 @@ class WC_Facebookcommerce_EventsTracker {
 
   public function __construct($user_info) {
     $this->pixel = new WC_Facebookcommerce_Pixel($user_info);
+
     add_action('wp_head', array($this, 'apply_filters'));
+
+    // Pixel Tracking Hooks
+    add_action('wp_head',
+	  array($this, 'inject_base_pixel'));
+    add_action('wp_footer',
+	  array($this, 'inject_base_pixel_noscript'));
+    add_action('woocommerce_after_single_product',
+	  array($this, 'inject_view_content_event'));
+    add_action('woocommerce_after_shop_loop',
+	  array($this, 'inject_view_category_event'));
+    add_action('pre_get_posts',
+	  array($this, 'inject_search_event'));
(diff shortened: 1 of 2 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -19,10 +19,10 @@ class WC_Facebookcommerce_EventsTracker {
 
   public function __construct($user_info) {
     $this->pixel = new WC_Facebookcommerce_Pixel($user_info);
-    self::apply_filters();
+    add_action('wp_head', array($this, 'apply_filters'), 9);
```
mengyingdu: Why it's 9?

mister: Because default priority is 10 and it will guarantee that this action will be fired before others.

mengyingdu: 10 should be good enough and we can just skip this param. It's a good idea to move the rest of [actions](https://github.com/facebookincubator/facebook-for-woocommerce/blob/master/facebook-commerce.php#L263-L283) here. Would you like to change this as well?

mengyingdu: Hi @mister ,

If you dont want to change the action setup, we can change it on our side. Can you update the priority then i will accept. Since we are using ShipIt to automatically sync with github, I can't help change your code on my side. Thanks for your pull request!

mister: Hi @mengyingdu ,

I'll try to update it today or tomorrow. Not a problem, was busy few past days.

mengyingdu:
Thanks for your pull request @mister ! If you have more questions when using this filter, feel free to reach out.

PR #268, opened 2018-02-05
Ensure jQuery is loaded before using it

### LINKED ISSUE #267: ReferenceError: jQuery is not defined
When jQuery is not included before this plugin, a JavaScript error is thrown on the console.

This could happen for a number of reasons, especially if you're using a plugin that moves JS code to the bottom, use defer or simply have a custom theme.

The plugin should ideally ensure jQuery is loaded before using it.

glagonikas:
Fixes #267


glagonikas:
Hey @jlkalberer considering this PR overwrites part of your previous commit, it's probably worth reviewing it.

PR #437, opened 2018-04-26
Fix performance issue by reusing existing wp_query object contents

ottok:
Running get_posts() triggers a database query which is unnecessary as
the object wp_query already contains the posts, so rather reuse it
directly.

I hereby contribute this as public domain to Facebook so I don't need to sign extra agreements.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -88,7 +88,7 @@ class WC_Facebookcommerce_EventsTracker {
     $products = array_values(array_map(function($item) {
         return wc_get_product($item->ID);
       },
-      $wp_query->get_posts()));
+      $wp_query->posts));
 
     // if any product is a variant, fire the pixel with
     // content_type: product_group


ottok:
Before:
rekki facebook optimization before

After:
rekki facebook optimization after

mengyingdu:
Hi @ottok ,

Thanks for your pull request! 
Since you didn't sign the Contributor License Agreement, I need to clone/copy your change and then land internally. Is that OK?

Thanks!

ottok:
Sure, go ahead as you see fit. The mention of public domain is inside the commit itself, so you should be able to merge that as well without problems.

dmitridr:
Thanks for this!! Out of curiosity what plugin/software did you use to generate those charts?

@mengyingdu closed this in https://github.com/facebookincubator/facebook-for-woocommerce/commit/c7d75756fc7cf9853d1d245c9ce336a23c805d39 it should be live in the next version.

ottok:
@dmitridr I used Tideways. See details at https://seravo.com/use-the-new-tideways-integration-to-debug-any-php-performance-bottle-necks-on-your-wordpress-site/

PR #525, opened 2018-06-04
WP_DEBUG: Make the error message more precise

Orlandster:
I've updated the error message to match the `condition` specified here: https://github.com/facebookincubator/facebook-for-woocommerce/blob/78afec714689242dabbd32216c195c77103ada73/facebook-for-woocommerce.php#L44

I think it can be misleading right now.

### DIFF
--- a/facebook-for-woocommerce.php
+++ b/facebook-for-woocommerce.php
@@ -71,7 +71,7 @@ class WC_Facebookcommerce {
       <p>
       <?php
         printf(__('To use Facebook for WooCommerce,
-          please disable WP_DEBUG_DISPLAY in your wp-config.php file.
+          please disable WP_DEBUG and WP_DEBUG_DISPLAY in your wp-config.php file.
           Contact your server administrator for more assistance.',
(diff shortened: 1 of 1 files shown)


facebook-for-woocommerce.php
```diff
@@ -71,7 +71,7 @@ public function wp_debug_display_error() {
       <p>
       <?php
         printf(__('To use Facebook for WooCommerce,
-          please disable WP_DEBUG_DISPLAY in your wp-config.php file.
+          please disable WP_DEBUG and WP_DEBUG_DISPLAY in your wp-config.php file.
```
dmitridr: The notice, as you rightly identified in your comment, only occurs when both are on. 
So it's enough if one of them is disabled, so could you change this to `or` instead of `and` ?

dmitridr:
Hey thanks for your pull request! Appreciate you taking the time, I have a comment inline about making the message even more precise. Once that's addressed, we're happy to merge this.

Issue #610, opened 2018-08-03
fbq('consent', 'revoke')


addedlovely:
Hello.

We'd like to use the revoke consent feature listed here:
https://developers.facebook.com/docs/facebook-pixel/events-advanced-use-cases/v3.0

But you haven't provided any hooks or filters to be able to inject that after the tracking script is initially loaded. Adding a filter here: https://github.com/facebookincubator/facebook-for-woocommerce/blob/3c858c9ba535a1c9cf7603ee418e3adc80117c26/facebook-commerce-pixel-event.php#L21 would solve the issue.

Thanks.

Ian.

jieminz22:
## Automated Support Message

Hello,

Thanks for your interest in our plugin!

If your issue is a feature request or feedback for the developers, reply with `Feature Request` or `Feedback`.
If you are a developer or have a question for the developers related to code, reply with `Developer Question`.
If you are running ads with us, include `Advertiser` in your reply.

If you have encountered a bug or a problem with the plugin, the steps below contain information to help you resolve the issue yourself and provide us information we need to help you with your problem.

## Search Existing Issues

**First, please [search our issues](https://github.com/facebookincubator/facebook-for-woocommerce/issues?q=) to see if your problem has been reported before.**
There may already be a workaround or a solution to your problem. You can search our issues using [the search bar near the top of our issues page](https://github.com/facebookincubator/facebook-for-woocommerce/issues?q=); be sure to search both _open_ and _closed_ issues.
If you do find a similar issue, leave a comment on that issue and close this one. Your comment helps us prioritize which existing issues are important and urgent.

## Common Resolution Methods

**If you are experiencing a problem or bug and have not found any similar issue, please try the following:**

* Upgrade to the latest version of the plugin.
* Try to `Resync` products using the button on the plugin config page
* Reset the plugin by selecting `Reconfigure > Advanced Options > Delete`, and setting up this plugin again.
* Try to deactivate other plugins that you think may be interfering, then reset and set up again.

## Necessary Information

**If the above steps did not help, please provide the following information in your next comment in order for us to debug your problem:**

* WooCommerce version
* WordPress version
* Plugin version (please note: we only support up-to-date versions of the plugin)
* A screenshot of where you see the problem (for example: if it is on Facebook, a screenshot of Facebook, if it is on WooCommerce, a screenshot of WooCommerce. If it is in multiple places, multiple screenshots.)
* The URL of your website
* The URL of your page
[...]

PR #613, opened 2018-08-07
Adding filter to Pixel init

addedlovely:
To allow for injection of fbq('consent', 'revoke') (GDPR) - #610 -https://developers.facebook.com/docs/facebook-pixel/events-advanced-use-cases/v3.0

### DIFF
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -275,11 +275,11 @@ src=\"https://www.facebook.com/tr?id=%s&ev=PageView&noscript=1\"/>
     $params = array(
       'agent' => $agent_string);
 
-    return sprintf(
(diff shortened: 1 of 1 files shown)


facebook-commerce-pixel-event.php
```diff
@@ -275,11 +275,11 @@ private function pixel_init_code() {
     $params = array(
       'agent' => $agent_string);
 
-    return sprintf(
+    return apply_filter('facebook_woocommerce_pixel_init', sprintf(
```
dmitridr: Thanks for the pull request, but we found a small problem : this should be `apply_filters` instead of `apply_filter`. If you make this change we can merge it.

dmitridr: Nevermind, It's a small change so I fixed it inline.

PR #759, opened 2018-11-07
Added filter for product ID on purchase event

ahegyes:
This filter would provide compatibility with plugins which make use of the WC checkout process and order management, but are selling something else, not WC Products directly.

For example, we've encountered this issue with the plugin LearnPress in conjunction with the extension LearnPress - WooCommerce Payments which enables the selling of access to courses (LearnPress) through the WooCommerce checkout process. In this case, the product ID is not stored inside $item['product_id'] but inside $item['course_id']. Since the value of $item['product_id'] is 0 in this case, the code breaks the checkout confirmation page.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -310,7 +310,7 @@ class WC_Facebookcommerce_EventsTracker {
     $content_type = 'product';
     $product_ids = array();
     foreach ($order->get_items() as $item) {
-      $product = wc_get_product($item['product_id']);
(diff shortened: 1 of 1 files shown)


dmitridr:
Hi @ahegyes thanks for the pull request!

Interesting case with LearnPress, if you encounter more such cases please feel free to add more filters into our plugin, we welcome it. We will just double check things on our side and run through our tests. Will aim to land it next week.

Cheers,
Dmitri

dmitridr:
Thanks for your patience, we are merging this today and will include it in the next version v1.9.9. 

If you need similar filters for ViewContent and AddToCart for the same reasons you outlined, please don't hesitate to send us additional pull requests, we're more than happy to review.

PR #892, opened 2019-02-23
Update the plugin version here

carazo:
Good afternoon,

In the readme file you have the version marked as:
Stable tag: 1.9.5

But in the plugin file it is:
 * Version: 1.9.10

Thanks!


diwangcmu:
Thanks for the effort! This is no longer applicable

Issue #493, opened 2018-05-19
Visual Composer Shortcode in product description


pixolab:
Hi, is there a solution to remove/strip the Visual Composer shortcodes from the product descriptions in order to have clear descriptions?

Exemple of shortcodes to strip: [vc_row][vc_column width="5/6"][vc_column_text] 

Thanks!

Plugin: 1.8.7
WC: 3.3.5
WP: 4.9.6
PHP: 7.0.0

![2018-05-20_09-13-28](https://user-images.githubusercontent.com/14137754/40276585-395e1e5a-5c0e-11e8-8b7a-99762874c414.png)

joe6ar:
Same problem, needs fixing, seems impossible to edit the description in facebook

mengyingdu:
[Note : This is an Automated Message]

Hello,

Thanks for your interest in our plugin!
Due to a high volume of issues we may not be able to get to yours immediately.

If this is a feature request, or feedback, please reply with
`Feature Request` or `Feedback` and it will be categorized accordingly, and
we will read it and reply when we can.

If you are experiencing problems, read on.

First. Please check to see if your issue is not a duplicate of another one.
To search our issues, use the search feature at the top of the GitHub issues page.
We have nearly 500 existing and past issues, so it is fairly likely someone else has already reported your problem.
If you do find a relevant issue, please comment in that issue, and close this one.
It helps us prioritize issues to tackle by looking at the number of people with the same problem.

If your issue is unique, and you are sure, then read on.
We require the following information in order to help you:
- WooCommerce Version
- Wordpress Version
- Plugin Version. Please Note that we do not support outdated versions of the plugin.
- A Screenshot of where you see the problem. If it is on Facebook, a screenshot of Facebook,
if it is on WooCommerce, a screenshot of WooCommerce. If it is in multiple places, multiple screenshots.
- The URL of your website.
- The URL of your page.
- The URL(s) where you took your screenshot(s).
- (If applicable) Detailed steps you took in order to see the problem.

Please be advised, it may take us some time to get to your issue.
In the meantime, below are generic approaches you can take that may solve the issue :
- Upgrade to the latest version.
- Force Resync
- Reset the plugin using `Reconfigure > Advanced Options > Delete` and setup again.
- Try to deactivate other plugins which you think could be interfering and repeating the above.

When you have the above data for us, please include
`Debug Ready` in your reply. We will get to it and debug it as soon as we can.

Thanks for your patience,
[...]

PR #1102, opened 2019-08-10
Remove leftover shortcodes

robwent:
This is related to issue #493 

The clean_string function seems to be used before some plugins have registered their shortcodes, so they don't get rendered by do_shortcode and are left in the description.

This removes anything within square brackets that are left in the description without removing the inner content (Using strip_shortcodes removes inner content).


robwent:
Also related to #977

rawdreeg:
Thanks for contributing.

The issue is addressed by #2254, so I will close this PR

PR #1175, opened 2019-11-28
Check if fb_api_key is set before calling it

legoblokje:
When PHP notices are on (or when WP_DEBUG is on) the following line produces a warning (Undefined index: fb_api_key) when the plug-in isn't setup yet. This in turn generates a JavaScript error which leaves the user unable to  start initial setup.


rawdreeg:
I am closing this PR as stale.

PR #1206, opened 2020-02-20
Added qTranslate-X support

Tusko:
Added qTranslate-X and -XT support for strings in `public function clean_string` and site name.
During synchronization language used is the same as the current language of admin area.


rawdreeg:
I will close as outdated.

PR #1281, opened 2020-05-12
Define the connection scopes

wvega:
# Summary

This PR updates the `Connection::get_scopes()` method to return the API scopes we are going to request.

### Tests

### DIFF
--- a/tests/integration/Handlers/ConnectionTest.php
+++ b/tests/integration/Handlers/ConnectionTest.php
@@ -87,7 +87,6 @@ class ConnectionTest extends \Codeception\TestCase\WPTestCase {
 
 		$scopes = $this->get_connection()->get_scopes();
 
-		$this->assertIsArray( $scopes );
 		$this->assertContains( $scope, $scopes );
 	}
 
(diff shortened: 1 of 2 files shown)


tests/integration/Handlers/ConnectionTest.php
```diff
@@ -79,7 +79,12 @@ public function test_get_disconnect_url() {
 	/** @see Connection::get_scopes() */
 	public function test_get_scopes() {
 
-		$this->assertIsArray( $this->get_connection()->get_scopes() );
+		$scopes = $this->get_connection()->get_scopes();
+
+		$this->assertIsArray( $scopes );
+		$this->assertContains( 'manage_business_extension', $scopes );
```
ChaseWiseman, line 85: Can we break these out into a data provider? Whenever possible I think we want to limit each `test_` method to one assert. That way if one fails the others still run and we can know exactly what needs fixing more quickly.

wvega: Sounds good. Thanks for the suggestion! I updated the tests in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1281/commits/6d1b8671ab073485b80a3081c164de93992f7245.

I also removed `assertIsArray()` as the version of PHPUnit we are using currently throws a warning if `assertContains()` is called with a string haystack:

```
[PHPUnit\Framework\Warning] Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.
```

wvega: Also added a filter, which I forgot to include on my first PRs: 8c34ef8.

PR #1282, opened 2020-05-12
Build the connection redirect URL

wvega:
# Summary

This PR updates the `Connection::get_redirect_url()` method to build the connection redirect URL.

### Tests

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -172,7 +172,9 @@ class Connection {
 	 */
 	public function get_redirect_url() {
 
-		return '';
+		$redirect_url = add_query_arg( 'wc-api', self::ACTION_CONNECT, home_url() );
+
+		return wp_nonce_url( $redirect_url, self::ACTION_CONNECT, 'nonce' );
 	}
 
 
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -20,6 +20,10 @@ defined( 'ABSPATH' ) or exit;
 class Connection {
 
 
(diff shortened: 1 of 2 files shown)


includes/Handlers/Connection.php
```diff
@@ -168,7 +172,17 @@ public function get_business_manager_id() {
...
+
+		/**
+		 * Filters the redirect URL where the user will return to after OAuth.
+		 *
+		 * @since 2.0.0-dev.1
+		 *
+		 * @param string $redirect_url redirect URL
+		 * @param Connection $connection connection handler instance
+		 */
+		return (string) apply_filters( 'wc_facebook_redirect_url', $redirect_url, $this );
```
ChaseWiseman, line 185: This wasn't really defined in the implementation plan, but how about we have all hook names in this class begin with `wc_facebook_connection_`?

wvega: @ChaseWiseman sounds good to me. I used `wc_facebook_connection_` for the scopes filter, because it seemed too generic without it, while the others are a bit more descriptive already, but it also makes sense to me to give them the same prefix if they all live in the same class.

Fixed in 3f3a505.

>This wasn't really defined in the implementation plan

Do you think we should follow the plan and only add filters if we see a specific use case? or did you mean that we didn't define a convention in the plan?

ChaseWiseman: >Do you think we should follow the plan and only add filters if we see a specific use case? or did you mean that we didn't define a convention in the plan?

Yeah I just meant that I didn't make any sort of filter naming convention clear in the plan.

The new filter looks good!

PR #1287, opened 2020-05-13
Build the connection parameters

unfulvio:
# Summary

This PR updates the `Connection::get_connection_parameters()` method to build the connection parameters.

### Tests

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -261,7 +261,42 @@ class Connection {
 	 */
 	public function get_connect_parameters() {
 
-		return [];
+		return [
+			'client_id'     => self::CLIENT_ID,
+			'redirect_uri'  => self::PROXY_URL,
+			'display'       => 'page',
+			'response_type' => 'token',
+			'scope'         => $this->get_scopes(),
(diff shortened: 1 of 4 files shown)


includes/Handlers/Connection.php
```diff
@@ -255,7 +261,42 @@ public function get_redirect_url() {
 	 */
 	public function get_connect_parameters() {
 
-		return [];
+		return [
+			'client_id'     => self::CLIENT_ID,
+			'redirect_uri'  => self::PROXY_URL,
+			'display'       => 'page',
+			'response_type' => 'token',
+			'scope'         => $this->get_scopes(),
```
unfulvio, line 269: by looking a the Authentication doc from FB I'm not sure if this can be an array of strings - the doc only mentions "string", does it mean comma separated strings? json encoded?

unfulvio: never mind it looks like it has to be comma separated

unfulvio: see https://github.com/facebookincubator/facebook-for-woocommerce/pull/1287/commits/e561825e436c75730bfe2ac22876f0b7bdca576e

PR #1311, opened 2020-05-19
Create the API class

deivamagalhaes:
# Summary
Creates the API class.

**Main Story:** [CH 54709](https://app.clubhouse.io/skyverge/story/54709/create-the-api-class)

## Tasks
- [x] Define class `Facebook\API` that extends `Framework\SV_WC_API_Base` in `includes/API.php`
- [x] Define the `acess_token` property
- [x] Add stub versions for the methods described in [the document](https://docs.google.com/document/d/1zfIwNjbnXd4Dfj1yPxsycNy4GvJnJj4mLhOeRC0cGWs/edit#heading=h.1uc9oyql4xrr)

## QA

N/A

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -215,7 +215,7 @@ class API extends SV_WC_API_Base {
 	 *
 	 * @since 2.0.0-dev.1
 	 *
-	 * @return SV_WC_Plugin
+	 * @return \WC_Facebookcommerce
 	 */
 	protected function get_plugin() {
 
--- a/includes/API.php
+++ b/includes/API.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package FacebookCommerce
+ */
+
+namespace SkyVerge\WooCommerce\Facebook;
+
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_API_Base;
+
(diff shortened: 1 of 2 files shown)


includes/API.php
```diff
@@ -0,0 +1,226 @@
...
+	/** @var string the configured access token */
+	protected $access_token;
+
+
+	/**
+	 * Gets a Page object from Facebook.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param $page_id page ID
```
unfulvio, line 36: doesn't specify what type $page_id is (which I believe is string), but I suppose we can update this later

includes/API.php
```diff
@@ -0,0 +1,226 @@
...
+
+
+	/**
+	 * Uses the Catalog Batch API to update or remove items from catalog.
+	 *
+	 * @see Sync::create_or_update_products()
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param array $requests array of prefixed product IDs to create, update or remove
```
unfulvio, line 52:
```suggestion
	 * @param string[] $requests array of prefixed product IDs to create, update or 
```

then it's an array of strings if this is not an associative array (can update later)

includes/API.php
```diff
@@ -0,0 +1,226 @@
...
+		// TODO: Implement get_new_request() method.
+	}
+
+
+	/**
+	 * Returns the plugin class instance associated with this API.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @return SV_WC_Plugin
```
unfulvio, line 218:
```suggestion
	 * @return \WC_Facebookcommerce
```

likely copypasta? :)

PR #1314, opened 2020-05-20
Schedule deletes for variable products

unfulvio:
# Summary

Updates `WC_Facebookcommerce_Integration::on_product_delete()` to schedule variations deletion.

**Main Story:** [CH 54692](https://app.clubhouse.io/skyverge/story/54692/schedule-deletes-for-variable-products)

## Tasks

- [x] Update `WC_Facebookcommerce_Integration::on_product_delete()` to not bail if the product does not have Facebook Product ID or Facebook Product Group ID defined
- [x] If the product is not variable and the product has a Facebook Product ID, call `::delete_product_item()`
- [x] If the product is variable call `Facebook\Products\Sync::delete_products()` with the ID of each variation
- [x] If the product has a Facebook Product Group ID, call `::delete_product_group()`

## QA

N/A

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -378,12 +378,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 				add_action( 'trashed_post', [ $this, 'on_product_trash' ] );
 
-				add_action(
-					'before_delete_post',
-					array( $this, 'on_product_delete' ),
-					10,
-					1
-				);
+				add_action( 'before_delete_post', [ $this, 'on_product_delete' ] );
 
 				  add_action( 'add_meta_boxes', array( $this, 'fb_product_metabox' ), 10, 1 );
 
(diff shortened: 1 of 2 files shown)


facebook-commerce.php
```diff
@@ -938,47 +933,45 @@ public function on_product_trash( $post_id ) {
...
 	 * Deletes a product from Facebook.
 	 *
-	 * @param int $wp_id product ID
+	 * @param int $product_id product ID
 	 */
-	public function on_product_delete( $wp_id ) {
+	public function on_product_delete( $product_id ) {
 
-		$woo_product = new WC_Facebook_Product( $wp_id );
+		$woo_product = new \WC_Facebook_Product( $product_id );
```
unfulvio, line 940: can we drop instantiating this object? doesn't seem to help us in any way in the context of this method and could be replaced by the result of the more standard `wc_get_product()`

unfulvio: similar thing in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1315

wvega: Yeah, I think that makes sense. We pass that to `get_product_fbid()` but is optional. I'll quickly change that before I merge.

wvega: I made the updates in 6fc8e49 and b7f4660.

PR #1315, opened 2020-05-20
Schedule visibility updates for variable products

unfulvio:
# Summary

Updates `WC_Facebookcommerce_Integration::update_fb_visibility()` to schedule visibility updates for variable products.

**Main Story:** [CH 54701](https://app.clubhouse.io/skyverge/story/54701/schedule-visibility-updates-for-variable-products)

## Tasks

- [x] Add an if statement before the foreach loop in `WC_Facebookcommerce_Integration::update_fb_visibility()` to check whether the product is variable
- [x] If the product is NOT variable, use `Products::set_product_visibility()` to set the visibility of the product
- [x] If the product is NOT variable, use `$this->fbgraph->update_product_item()`, as its currently done, to update the visibility of the product
- [x] If the product is variable, loop over all the product variations and use `Products::set_product_visibility()` to set the visibility of each variation
- [x] If the product is variable, call `Facebook\Products\Sync::create_or_update_products()` to enqueue the variations to be synced

## QA

N/A

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -4238,6 +4238,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		return $to_delete;
 	}
 
+
 	/**
 	 * Helper function to update FB visibility.
 	 *
@@ -4251,16 +4252,16 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 			return;
 		}
 
-		$woo_product = new \WC_Facebook_Product( $product_id );
+		$product = wc_get_product( $product_id );
 
 		// bail if product isn't found
-		if ( ! $woo_product->woo_product instanceof \WC_Product || ! $woo_product->exists() ) {
+		if ( ! $product instanceof \WC_Product ) {
 			return;
 		}
 
 		$should_set_visible = $visibility === self::FB_SHOP_PRODUCT_VISIBLE;
 
-		if ( ! $woo_product->woo_product->is_type( 'variable' ) ) {
+		if ( ! $product->is_type( 'variable' ) ) {
 
 			$fb_product_item_id = $this->get_product_fbid( self::FB_PRODUCT_ITEM_ID, $product_id );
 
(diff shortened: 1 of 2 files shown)


facebook-commerce.php
```diff
@@ -4240,49 +4240,57 @@ function fb_duplicate_product_reset_meta( $to_delete ) {
...
-	function update_fb_visibility( $wp_id, $visibility ) {
+	function update_fb_visibility( $product_id, $visibility ) {
 
 		// bail if the plugin is not configured properly
 		if ( ! $this->is_configured() || ! $this->get_product_catalog_id() ) {
 			return;
 		}
 
-		$woo_product = new WC_Facebook_Product( $wp_id );
+		$woo_product = new \WC_Facebook_Product( $product_id );
```
unfulvio, line 4254: can we drop instantiating this object? doesn't seem to help us in any way in the context of this method

I'd just replace it with `wc_get_product()`

unfulvio: similar thing in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1314

deivamagalhaes: I'm okay with replacing with `wc_get_product` and just checking `$woo_product instanceof \WC_Product`

facebook-commerce.php
```diff
@@ -4240,49 +4240,57 @@ function fb_duplicate_product_reset_meta( $to_delete ) {
...
-				$is_visible = $visibility === self::FB_SHOP_PRODUCT_VISIBLE;
+			$set_visibility = $this->fbgraph->update_product_item( $fb_product_item_id, [ 'visibility' => $visibility ] );
+
+			if ( $this->check_api_result( $set_visibility ) ) {
+				Products::set_product_visibility( $woo_product->woo_product, $should_set_visible );
+			}
+
+		} else {
+
+			$product_ids = array_merge( [ $product_id ], $woo_product->get_children() );
```
deivamagalhaes, line 4280: If I'm not mistaken this would throw an exception now, as `$woo_product` is currently a `WC_Facebook_Product`.

unfulvio: yeah one reason why I'd like to just use a plain \WC_Product; the facebook product does have a __call override so the methods would still work https://github.com/facebookincubator/facebook-for-woocommerce/blob/d81f04033cd0ef4db5cfc7c14a1a51cd8f2a58c8/includes/fbproduct.php#L79

you can see this is currently used when passing the facebook product  to this helper method: https://github.com/facebookincubator/facebook-for-woocommerce/blob/a31edf0742b9d13d719f3ac4f98bcda9c0f6bb7c/facebook-commerce.php#L4298

which in turn also calls get_children() here https://github.com/facebookincubator/facebook-for-woocommerce/blob/d81f04033cd0ef4db5cfc7c14a1a51cd8f2a58c8/includes/fbutils.php#L170

I will update the PR since we agreed so far that we can easily switch to the standard WC_Product for the context of these methods

thanks for double checking!

facebook-commerce.php
```diff
@@ -4238,51 +4238,67 @@ function fb_duplicate_product_reset_meta( $to_delete ) {
...
+			}
+
+		} else {
+
+		    // parent product
+			Products::set_product_visibility( $product, $should_set_visible );
 
-				update_post_meta( $item_id, Products::VISIBILITY_META_KEY, wc_bool_to_string( $is_visible ) );
-				update_post_meta( $wp_id, Products::VISIBILITY_META_KEY, wc_bool_to_string( $is_visible ) );
+			$product_ids = [ $product_id ];
```
unfulvio, line 4284: should we still set visibility for variable parent product and pass that to the array of IDs to sync?

I believe this is what was happening before this PR, in this method's context

unfulvio:
hey @deivamagalhaes thanks for checking this, I've updated the PR to use only \WC_Product. I have a last question in PR comments.

facebook-commerce.php
deivamagalhaes: @wvega, I'd like your input on this one. When updating visibility for a variable product, we schedule the update for the variations. What about the parent product? Should we also schedule it or run it immediately as we do for simple products?

wvega: @unfulvio @deivamagalhaes based on what I've seen while working with the API, product groups, which are used to represent the parent product for variable product, don't have visibility property on Facebook. I think we should **not** add the parent product's ID to the list of IDs to update or create.

The previous version of the method was using [`WC_Facebookcommerce_Utils::get_product_array( $woo_product )`](https://github.com/facebookincubator/facebook-for-woocommerce/blob/2d0679bb3a6246698d1813a816e6e74c3e1d30d6/includes/fbutils.php#L159-L177) to get the IDs, which doesn't include the ID of the parent in the case of a variable product.

deivamagalhaes: Perfect, thanks, @wvega, for clarifying! 🙏

PR #1320, opened 2020-05-20
Implement the method to process a job item

wvega:
# Summary

This PR implements the `Sync\Background::process_item()` method.

## Details

I also added a couple of private helper methods to separate the code that processes UPDATE and DELETE requests, and the code that processes the data returned from `WC_Facebook_Product::prepare_product()`.

## QA

- [x] `vendor codecept run integration tests/integration/Products/Sync/BackgroundTest.php` pass

### DIFF
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -143,6 +143,20 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 	public function process_item( $item, $job ) {
 
 		list( $product_id, $method ) = $item;
+
+		$product = wc_get_product( $product_id );
+
+		if ( ! $product instanceof \WC_Product ) {
+			// TODO: throw an exception and add a test
+			return;
+		}
+
+		if ( ! in_array( $method, [ Sync::ACTION_UPDATE, Sync::ACTION_DELETE ], true ) ) {
+			// TODO: throw an exception and add a test
+			return;
+		}
+
+		return $this->process_item_update( $product );
 	}
 
 
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -182,7 +182,11 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 	 */
 	private function process_item_update( $product ) {
 
-		$product_data = $this->prepare_product_data( $product );
+		if ( $product->is_type( 'variation' ) ) {
+			$product_data = $this->prepare_product_variation_data( $product );
+		} else {
+			$product_data = $this->prepare_product_data( $product );
+		}
 
 		$request = [
 			'retailer_id' => $product_data['retailer_id'],
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -264,47 +264,17 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 	 *
 	 * @param \WC_Product $product product object
 	 * @return array
-	 * @throws Framework\SV_WC_Plugin_Exception
 	 */
 	private function prepare_product_data( $product ) {
 
-		if ( $product->is_type( 'variation' ) ) {
-
-			$parent_product = wc_get_product( $product->get_parent_id() );
-
-			if ( ! $parent_product instanceof \WC_Product ) {
-				throw new Framework\SV_WC_Plugin_Exception( "No parent product found with ID equal to {$product->get_parent_id()}." );
-			}
-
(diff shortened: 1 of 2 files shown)


includes/Products/Sync/Background.php
```diff
@@ -139,9 +145,115 @@ public function process_items( $job, $data, $items_per_batch = null ) {
...
+		return $request;
+	}
+
+
+	/**
+	 * Processes an UPDATE sync request for the given product.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param \WC_Product $product product object
```
jdeeburke, line 179: Missing `@return` and `@throws` doc params

includes/Products/Sync/Background.php
```diff
@@ -139,9 +145,115 @@ public function process_items( $job, $data, $items_per_batch = null ) {
...
+		$data['retailer_product_group_id'] = $retailer_product_group_id ?: $data['retailer_id'];
+
+		// attributes other than size, color, pattern, or gender need to be included in the additional_variant_attributes field
+		if ( isset( $data['custom_data'] ) && is_array( $data['custom_data'] ) ) {
+
+			$data['additional_variant_attributes'] = $data['custom_data'];
+			unset( $data['custom_data'] );
+		}
+
+		return $data;
```
jdeeburke, line 240: May be a good spot to add a filter. Especially with the harcoded `condition`, even if it has always been done that way it wouldn't hurt to give shops an option to change it and other data if needed

jdeeburke: Actually on second thought, it may make more sense to put the filter in `process_item_update` and another matching filter in `process_item_delete` instead.

includes/Products/Sync/Background.php
```diff
@@ -139,9 +145,115 @@ public function process_items( $job, $data, $items_per_batch = null ) {
...
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param \WC_Product $product product object
+	 * @return array
+	 * @throws Framework\SV_WC_Plugin_Exception
+	 */
+	private function prepare_product_data( $product ) {
+
+		if ( $product->is_type( 'variation' ) ) {
```
jdeeburke, line 204: Optional suggestion: extract the variation logic into its own method for clarity / cleaner code

tests/integration/Products/Sync/BackgroundTest.php
```diff
@@ -79,9 +80,253 @@ public function test_process_items() {
...
+
+		// compare results with specific values from the test case
+		if ( isset( $request['retailer_id'] ) ) {
+			$this->assertEquals( $request['retailer_id'], $result['retailer_id'] );
+		}
+
+		if ( isset( $request['data'] ) ) {
+
+			foreach ( $request['data'] as $key => $value ) {
+				$this->assertEquals( $request['data'][ $key ], $result['data'][ $key ] );
```
jdeeburke, line 149: Can't you perform [`assertEquals`](https://phpunit.readthedocs.io/en/9.0/assertions.html#assertequals) with an array inputs?
Or, if you're concerned that the keys may be out of order and cause it to fail, can use [`assertEqualsCanonicalizing`](https://phpunit.readthedocs.io/en/9.0/assertions.html#assertequalscanonicalizing)

includes/Products/Sync/Background.php
wvega: That's good idea. I added the filters in d30612d, 8a4ae86, and tests in e3fabe7.

includes/Products/Sync/Background.php
wvega: Sounds good. I added a couple of updates for that., starting from https://github.com/facebookincubator/facebook-for-woocommerce/pull/1320/commits/32b0e031a6de18f23b169359dbf88cf1e995f108. What do you think?

tests/integration/Products/Sync/BackgroundTest.php
wvega: The idea is that `$request['data']` has an incomplete version of all the data that can be returned. Users of `check_process_item_update_request()` would only include fields that they are interested in checking. So the two arrays don't need to be equal, but all items that appear in `$request['data']` must appear and match items in `$result['data]`.

jdeeburke: Ah cool. Makes sense then. 👍

includes/Products/Sync/Background.php
jdeeburke: Looks great! Thanks for changing that!

PR #1321, opened 2020-05-20
Create a request to Send Item Updates

wvega:
# Summary

This PR implements the `API\Catalog\Send_Item_Updates\Request` class, the constructor and new setter methods.

## Details

I also added the implementation for the constructor and `set_data()` methods in the parent class.

## QA

- [ ] `vendor codecept run integration tests/tests/integration/API/RequestTest.php` pass
- [ ] `vendor codecept run integration tests/tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php` pass

### DIFF
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -42,4 +42,17 @@ class Request extends API\Request  {
 	}
 
 
+	/**
+	 * Sets the array of item update requests.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param array $requests item update requests
+	 */
+	public function set_requests( array $requests ) {
+
+		$this->requests = $requests;
+	}
+
+
 }
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -30,6 +30,19 @@ class Request extends API\Request  {
 
 
 	/**
+	 * Gets the ID of this request for rate limiting purposes.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @return string
+	 */
+	public static function get_rate_limit_id() {
+
+		return 'ads_management_api_request';
+	}
+
+
+	/**
 	 * API request constructor.
 	 *
 	 * @since 2.0.0-dev.1
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -21,4 +21,12 @@ use SkyVerge\WooCommerce\Facebook\API;
  */
 class Request extends API\Request  {
 
+
+	/** @var array an arary of item update requests */
+	protected $requests = [];
+
+	/** @var bool determines whether updates for products that are not currently in the catalog should create new items */
+	protected $allow_upsert = true;
+
+
 }
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -73,7 +73,7 @@ class Request extends API\Request  {
 	 *
 	 * @since 2.0.0-dev.1
 	 *
-	 * @param bool $allow_upsert wehther to allow updates to insert new items
+	 * @param bool $allow_upsert whether to allow updates to insert new items
 	 */
 	public function set_allow_upsert( $allow_upsert ) {
 
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -22,7 +22,7 @@ use SkyVerge\WooCommerce\Facebook\API;
 class Request extends API\Request  {
 
 
-	/** @var array an arary of item update requests */
+	/** @var array an array of item update requests */
 	protected $requests = [];
 
 	/** @var bool determines whether updates for products that are not currently in the catalog should create new items */
--- a/includes/API/Catalog/Send_Item_Updates/Request.php
+++ b/includes/API/Catalog/Send_Item_Updates/Request.php
@@ -29,4 +29,17 @@ class Request extends API\Request  {
 	protected $allow_upsert = true;
 
 
+	/**
+	 * API request constructor.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param string $catalog_id catalog ID
+	 */
+	public function __construct( $catalog_id ) {
+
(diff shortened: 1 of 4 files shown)


includes/API/Catalog/Send_Item_Updates/Request.php
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Base API request object.
+ *
+ * @since 2.0.0-dev.1
+ */
+class Request extends API\Request  {
+
+
+	/** @var array an arary of item update requests */
```
deivamagalhaes, line 25:
```suggestion
	/** @var array an array of item update requests */
```

includes/API/Catalog/Send_Item_Updates/Request.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->requests = $requests;
+	}
+
+
+	/**
+	 * Sets whether updates for products that are not currently in the catalog should create new items.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param array $data
```
deivamagalhaes, line 76: Incorrect `@param`

includes/API/Request.php
```diff
@@ -33,6 +33,8 @@ class Request extends Framework\SV_WC_API_JSON_Request {
 	 */
 	public function __construct( $object_id, $path, $method ) {
 
+		$this->method = $method;
+		$this->path   = $path ? sprintf( '/%s/%s', $object_id, trim( $path, '/' ) ) : "/{$object_id}";
```
deivamagalhaes, line 37: Should this be decided by the parent class? For the Find request, if I understood correctly, there is no `object_id`. I was planning on sending the `path` and have the parent just assign it.

tests/integration/API/RequestTest.php
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * @see Request::__construct()
+	 *
+	 * @param string $object_id object ID
+	 * @param string $path endpoint route
+	 * @param string $method HTTP method
+	 *
+	 * @dataProvider provider_constructor
+	 */
+	public function test_constructor( $object_id, $path, $method, $expect_path ) {
```
deivamagalhaes, line 37:
```suggestion
	public function test_constructor( $object_id, $path, $method, $expected_path ) {
```

tests/integration/API/RequestTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+	/** Test methods **************************************************************************************************/
+
+
+	/**
+	 * @see Request::__construct()
+	 *
+	 * @param string $object_id object ID
+	 * @param string $path endpoint route
+	 * @param string $method HTTP method
```
deivamagalhaes, line 33: Missing the last param.

tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+
+	/** @see Request::set_requests() */
+	public function test_set_requests() {
+
+		$request = new Request( '1234 ');
+
+		$request->set_requests( [] );
+
+		// TODO: assert that the array of requests is included in get_data()
```
deivamagalhaes, line 46: What is missing so that this can be added?

includes/API/Catalog/Send_Item_Updates/Request.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+namespace SkyVerge\WooCommerce\Facebook\API\Catalog\Send_Item_Updates;
+
+defined( 'ABSPATH' ) or exit;
+
+use SkyVerge\WooCommerce\Facebook\API;
+
+/**
+ * Base API request object.
```
deivamagalhaes, line 18: Needs to be updated.

includes/API/Request.php
deivamagalhaes: I just saw that this is in the implementation doc, so it should be good.

tests/integration/API/RequestTest.php
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @dataProvider provider_constructor
+	 */
+	public function test_constructor( $object_id, $path, $method, $expect_path ) {
+
+		$request = new Request( $object_id, $path, $method );
+
+		$this->assertEquals( $expect_path, $request->get_path() );
```
deivamagalhaes, line 41:
```suggestion
		$this->assertEquals( $expected_path, $request->get_path() );
```

includes/API/Request.php
wvega: That's a good point. With this implementation we abuse the `$object_id` parameter a little, for requests that don't use an ID. Using the `path` alone as you suggest could have a been a better approach here to support more scenarios.

I think we can leave it as is here, but we can consider tweaking it before we are done with these updates.

includes/API/Catalog/Send_Item_Updates/Request.php
wvega: Fixed in e2ce569

includes/API/Catalog/Send_Item_Updates/Request.php
wvega: Updated in 7384e5b

tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php
wvega: We have another [story](https://app.clubhouse.io/skyverge/story/54734/prepare-the-data-for-the-send-item-updates-request) to overwrite the `get_data()` method in this class and return a value that would include the values from the requests and the allow user properties.

I thought I could use get_data() for testing instead of accessing the protected property with reflection.

tests/integration/API/RequestTest.php
wvega: Fixed in a6d0ab3. Thanks for catching those oversights.

includes/API/Catalog/Send_Item_Updates/Request.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->requests = $requests;
+	}
+
+
+	/**
+	 * Sets whether updates for products that are not currently in the catalog should create new items.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param bool $allow_upsert wehther to allow updates to insert new items
```
deivamagalhaes, line 76:
```suggestion
	 * @param bool $allow_upsert whether to allow updates to insert new items
```

wvega: Thanks! I committed your suggestion.

PR #1325, opened 2020-05-20
Disable Facebook sync options for Virtual products

wvega:
This PR brings the following changes from the private repo:

```
 * Tweak - Hide Facebook options for virtual products and virtual variations
 * Tweak - Do not allow merchant to bulk enable sync for virtual products
 * Tweak - On upgrade, automatically disable sync for virtual products and virtual variations
```

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -112,7 +112,6 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 				$this->product_feed            = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
 				$this->products_sync_handler   = new \SkyVerge\WooCommerce\Facebook\Products\Sync();
 				$this->sync_background_handler = new \SkyVerge\WooCommerce\Facebook\Products\Sync\Background();
-				$this->product_feed = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
 
 				if ( is_ajax() ) {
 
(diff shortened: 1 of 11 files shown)


class-wc-facebookcommerce.php
```diff
@@ -108,6 +112,7 @@ public function init() {
 				$this->product_feed            = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
 				$this->products_sync_handler   = new \SkyVerge\WooCommerce\Facebook\Products\Sync();
 				$this->sync_background_handler = new \SkyVerge\WooCommerce\Facebook\Products\Sync\Background();
+				$this->product_feed = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
```
deivamagalhaes, line 115: Duplicated.

includes/Lifecycle.php
```diff
@@ -39,6 +39,7 @@ public function __construct( $plugin ) {
 			'1.10.0',
 			'1.10.1',
 			'1.11.0',
+			'2.0.0',
```
deivamagalhaes, line 42: I'd prefer to release this before `2.0.0`.

class-wc-facebookcommerce.php
wvega: Thank for catching that one. I removed the duplicated line in 6664e0d.

includes/Lifecycle.php
wvega: I reverted the commits that change the version number with f6a1b29. We will need to revert others from the changes I merged yesterday, but we can do that in another PR after 1.11.3 is out.

wvega:
Closing in favor of #1335. Everything change included in this PR is also there.

PR #1326, opened 2020-05-20
Add an API method to create a Product Group

deivamagalhaes:
# Summary

This PR implements the `API::create_product_group()` method.

## QA

- [x] `vendor codecept run tests/integration/APITest.php` pass

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -73,7 +73,7 @@ class API extends Framework\SV_WC_API_Base {
 	 */
 	public function create_product_group( $catalog_id, $data ) {
 
-		$request = new Request( $catalog_id, '/product_groups', 'POST' );
+		$request = $this->get_new_request( [ $catalog_id, '/product_groups', 'POST' ] );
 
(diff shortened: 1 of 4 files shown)


includes/API.php
```diff
@@ -62,12 +66,20 @@ public function send_item_updates( $requests, $allow_upsert ) {
...
-	 * @param array $data
+	 * @param string $catalog_id catalog ID
+	 * @param array $data product group data
+	 * @return Response
+	 * @throws Framework\SV_WC_API_Exception
 	 */
 	public function create_product_group( $catalog_id, $data ) {
 
-		// TODO: Implement create_product_group() method.
+		$request = $this->get_new_request( [ $catalog_id, '/product_groups', 'POST' ] );
```
wvega, line 76: I think the approach you mentioned earlier, to have the `API\Request` class receive the path and let users of the class build it however they need is much better. I'll create a story to make that update. Thank you for bringing that up!

PR #1329, opened 2020-05-20
Add an API method to create a Product Item

deivamagalhaes:
# Summary

This PR implements the `API::create_product_item()` method.

## QA

- [ ] `vendor codecept run tests/integration/APITest.php` pass

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -129,12 +129,20 @@ class API extends Framework\SV_WC_API_Base {
 	 *
 	 * @since 2.0.0-dev.1
 	 *
-	 * @param string $product_group_id
-	 * @param array $data
+	 * @param string $product_group_id parent product ID
+	 * @param array $data product data
+	 * @return Response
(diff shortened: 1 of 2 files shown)


includes/API.php
```diff
@@ -129,12 +129,20 @@ public function find_product_item( $catalog_id, $retailer_id ) {
...
-	 * @param array $data
+	 * @param string $product_group_id parent product ID
+	 * @param array $data product data
+	 * @return Response
+	 * @throws Framework\SV_WC_API_Exception
 	 */
 	public function create_product_item( $product_group_id, $data ) {
 
-		// TODO: Implement create_product_item() method.
+		$request = $this->get_new_request( [ $product_group_id, '/products', 'POST' ] );
```
jdeeburke, line 139: I'm not a fan of how `get_new_request()` is set up with the arguments array. At the very least it should be a keyed array so there's some clarity IMO. I know it's not something about this PR though :) I left a comment on another story, hopefully we can fix it.

PR #1333, opened 2020-05-21
Prepare the data for the Send Item Updates request

wvega:
# Summary

This PR implements the `API\Catalog\Send_Item_Updates\Request::get_data()` method.

## Details

Also added getter methods for `$requests` and `$allow_upsert` for consistency with the other properties.

## QA

- [x] `vendor codecept run integration tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php` pass


tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php
```diff
@@ -39,22 +40,54 @@ public function test_constructor() {
 	/** @see Request::set_requests() */
 	public function test_set_requests() {
 
-		$request = new Request( '1234 ');
+		$request  = new Request( '1234 ');
+		$requests = [ [ 'method' => Sync::ACTION_UPDATE ] ];
```
deivamagalhaes, line 44: Is this not supposed to be indexed by product ID?

```suggestion
		$requests = [ [ '123456' => Sync::ACTION_UPDATE ] ];
```

deivamagalhaes: I know it does not affect the test result, but just to avoid any confusion.

tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php
```diff
@@ -39,22 +40,54 @@ public function test_constructor() {
...
+	public function provider_set_allow_upsert() {
+
+		return [ [ true ], [ false ] ];
+	}
+
+
+	/** @see Request::get_data() */
+	public function test_get_data() {
+
+		$requests     = [ [ 'method' => Sync::ACTION_UPDATE ] ];
```
deivamagalhaes, line 79:
```suggestion
		$requests     = [ [ '123456' => Sync::ACTION_UPDATE ] ];
```

tests/integration/API/Catalog/Send_Item_Updates/RequestTest.php
wvega: The `$requests` array here is the one we send to Facebook and will normally include one or more arrays, each with [`retailer_id`, `method`, and `data` entries](https://developers.facebook.com/docs/marketing-api/catalog-batch/guides/send-item-updates#parameters).

Maybe you're referring to the array of requests that we pass to the background job?

I agree that the current approach is confusing though, so I should probably update the arrays in the test to include `retailer_id` and `data` entries too. What do you think?

deivamagalhaes: I think that's a good idea, to avoid the same confusion I made, lol!

wvega:
@deivamagalhaes fixed in f411e63. I updated it to make it more similar to a real requests object.

Thank you for merging the changes from master!

PR #1335, opened 2020-05-21
Merge from v1.11.3 

deivamagalhaes:
# Summary

Merging the remaining changes from release 1.11.3 from SkyVerge's repo:

```
2020.05.20 - version 1.11.3
 * Tweak - Write product feed to a temporary file and rename it when done, to prevent Facebook from downloading an incomplete feed file
 * Tweak - Hide Facebook options for virtual products and virtual variations
 * Tweak - Do not allow merchant to bulk enable sync for virtual products
 * Tweak - On upgrade, automatically disable sync for virtual products and virtual variations
 * Tweak - When using checkboxes for tags, make sure the modal is displayed when trying to enable sync for a product with an excluded tag
```

## Details

We are intentionally merging the commits to preserve history, but most of them are not really introducing code changes since the code changes were already previously merged via a patch.

## QA

- [x] Integration tests pass
- [x] Acceptance tests pass


tests/integration/Handlers/ConnectionTest.php
```diff
@@ -15,8 +15,6 @@ class ConnectionTest extends \Codeception\TestCase\WPTestCase {
 	public function _before() {
 
 		parent::_before();
-
-		require_once 'includes/Handlers/Connection.php';
```
deivamagalhaes, line 19: Removed unnecessary require that was causing the integration tests to fail with this error:

```
Fatal error: Cannot declare class SkyVerge\WooCommerce\Facebook\Handlers\Connection, because the name is already in use in /Users/deivamagalhaes/Code/facebook-for-woocommerce/includes/Handlers/Connection.php on line 22
```

PR #1337, opened 2020-05-21
Create a response for the Page object

wvega:
# Summary

This PR implements the `API\Pages\Read\Response` class.

## QA

- [x] `codecept run integration tests/integration/API/Pages/Read/ResponseTest.php` pass

### DIFF
--- a/includes/API/Pages/Read/Response.php
+++ b/includes/API/Pages/Read/Response.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package FacebookCommerce
+ */
+
+namespace SkyVerge\WooCommerce\Facebook\API\Pages\Read;
+
+defined( 'ABSPATH' ) or exit;
+
+use SkyVerge\WooCommerce\Facebook\API;
+
+/**
+ * Page API response object.
+ *
+ * @since 2.0.0-dev.1
+ */
+class Response extends API\Response  {
+
+
+}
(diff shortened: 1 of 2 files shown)


includes/API/Pages/Read/Response.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Gets the page name.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @return string|null
+	 */
+	public function get_name() {
+
+		return $this->name;
```
deivamagalhaes, line 34: Will the API handle the exception properly if the property does not exist?

tests/integration/API/Pages/Read/ResponseTest.php
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+
+	/** Test methods **************************************************************************************************/
+
+
+	/** @see Response::get_name() */
+	public function test_get_name() {
+
+		$raw_response = json_encode( [ 'name' => 'Test Page', 'link' => 'https://example.org' ] );
```
deivamagalhaes, line 32: We should probably test with invalid response data as well.

includes/API/Pages/Read/Response.php
wvega: It would return `null` if the property doesn't exist. 

I recently added a [story](https://app.clubhouse.io/skyverge/story/55072/handle-graph-api-errors) to handle API exceptions. I believe that if we don't get an API error, we should receive all the field that requested. If that's the case, if `perform_request()` returns a response object, it should have the fields we need.

tests/integration/API/Pages/Read/ResponseTest.php
wvega: Yes, that sounds good. Even in the case of an API error, the response object is created with a JSON object like the one [here](https://developers.facebook.com/docs/graph-api/using-graph-api/error-handling#handling-errors), so we could test with that and confirm that the method return `null` (instead of a notice).

PR #1338, opened 2020-05-21
Add a method to get the API instance

wvega:
# Summary

This PR adds a getter for the API instance, defines the `request_uri` and `request_headers` of the API class, and adds the implementation for the `API\Response::get_id()` method.

## Details

The items above are missing pieces that I failed to define as tasks on the associated stories.

## QA

- [x] `codecept run integration tests/integration/WC_Facebookcommerce_Test.php` pass
- [x] `codecept run integration tests/integration/API/ResponseTest.php` pass

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -209,11 +209,16 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 		 *
 		 * @since 2.0.0-dev.1
 		 *
-		 * @return SkyVerge\WooCommerce\Facebook\API
+		 * @return \SkyVerge\WooCommerce\Facebook\API
+		 * @throws Framework\SV_WC_API_Exception
 		 */
 		public function get_api() {
 
 			if ( ! is_object( $this->api ) ) {
+
+				if ( ! $this->get_connection_handler()->get_access_token() ) {
+					throw new Framework\SV_WC_API_Exception( __( 'Cannot create the API instance because the access token is missing.', 'facebook-for-woocommerce' ) );
+				}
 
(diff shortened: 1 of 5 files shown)


class-wc-facebookcommerce.php
```diff
@@ -201,6 +204,28 @@ public function add_wordpress_integration() {
...
+		 */
+		public function get_api() {
+
+			if ( ! is_object( $this->api ) ) {
+
+				require_once __DIR__ . '/includes/API.php';
+				require_once __DIR__ . '/includes/API/Request.php';
+				require_once __DIR__ . '/includes/API/Response.php';
+
+				$this->api = new SkyVerge\WooCommerce\Facebook\API( $this->get_connection_handler()->get_access_token() );
```
deivamagalhaes, line 222: Should we verify if the token is available before trying to instantiate the API?

wvega: I'm not against it. I didn't add a verification because we a lot of checks to bail if we don't have an access token, so is less likely that we try to use the API with a token.

Also, if the access token is not set, trying to use any of the API methods should throw a `SV_WC_API_Exception`, that we will likely try to catching whenever we use the API, so I thought that was good enough.

Do you think it would be better to check whether the plugin is connected here and throw `SV_WC_API_Exception` if it isn't ?

deivamagalhaes: I believe so, there is no reason to instantiate an object if nobody will be able to use it, IMO. It's probably better to throw the exception first and save the resources.

wvega: Sounds good. I added the exception in 515ed07 and a new test in d933277.

PR #1343, opened 2020-05-21
Handle Graph API errors

wvega:
# Summary

This PR implements the `API::do_post_parse_response_validation()` method to thrown an exception if a rate limiting or general API error is included in the response.

## QA

- [ ] `vendor codecept run integration tests/integration/APITest.php` pass

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -97,11 +97,15 @@ class API extends Framework\SV_WC_API_Base {
 	 * @since 2.0.0-dev.1
 	 *
 	 * @param $page_id page ID
-	 * @return
+	 * @return API\Pages\Read\Response
 	 */
 	public function get_page( $page_id ) {
 
-		// TODO: Implement get_page() method.
+		$request = new API\Pages\Read\Request( $page_id );
+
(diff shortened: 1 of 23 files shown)


includes/API.php
```diff
@@ -49,6 +48,52 @@ public function __construct( $access_token ) {
...
+			 *
+			 * @link https://developers.facebook.com/docs/graph-api/using-graph-api/error-handling#errorcodes
+			 * @link https://developers.facebook.com/docs/graph-api/using-graph-api/error-handling#rate-limiting-error-codes
+			 * @link https://developers.facebook.com/docs/marketing-api/reference/product-catalog/batch/#validation-rules
+			 */
+			if ( in_array( $response->get_api_error_code(), [ 4, 17, 32, 613, 80004 ], true ) ) {
+
+				throw new API\Exceptions\Request_Limit_Reached( $message, $response->get_api_error_code() );
+
+			} else {
```
unfulvio, line 89: `else` not needed since the first throw would stop the train and the latter throw would be executed if the first if condition is false

wvega: Good point. I removed the `else` in e3b2472.

PR #1346, opened 2020-05-22
Add an API method to retrieve page info

wvega:
# Summary

This PR implements the `API::get_oage()` method.

## QA

- [x] `vendor codecept run integration tests/integration/APITest.php` pass

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -59,7 +59,11 @@ class API extends Framework\SV_WC_API_Base {
 	 */
 	public function get_page( $page_id ) {
 
-		// TODO: Implement get_page() method.
+		$request = new API\Pages\Read\Request( $page_id );
+
+		$this->set_response_handler( API\Pages\Read\Response::class );
+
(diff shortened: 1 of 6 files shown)


includes/API.php
```diff
@@ -55,11 +55,15 @@ public function __construct( $access_token ) {
 	 * @since 2.0.0-dev.1
 	 *
 	 * @param $page_id page ID
-	 * @return
+	 * @return API\Pages\Read\Response
```
unfulvio, line 58:
```suggestion
	 * @return API\Pages\Read\Response
	 * @throws Framework\SV_WC_API_Exception
```

seems `perform_request()` is intended to throw an exception in framework - we might not implement throws, but if the response is a WP_Error it still might?   just double checking

wvega: Good catch @unfulvio! You're right, the framework throws an exception if we get `WP_Error` and we know throw exceptions too, so your suggestion is spot on. Fixed in 228a592.

PR #1347, opened 2020-05-22
Get page name and URL in a single request

wvega:
# Summary

This PR updates the `WC_Facebookcommerce_Integration::get_page_name()` and `::get_page_url()` to retrieve the name and the URL of the configured page with a single API request.

## QA

- [x] `vendor codecept run integration` pass

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -4106,13 +4106,9 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	 */
 	public function get_page_url() {
 
-		if ( $this->is_configured() ) {
-			$page_url = $this->fbgraph->get_page_url( $this->get_facebook_page_id() );
-		} else {
-			$page_url = '';
-		}
+		$page = $this->get_page();
 
-		return is_string( $page_url ) ? $page_url : '';
+		return isset( $page['url'] ) ? $page['url'] : '';
 	}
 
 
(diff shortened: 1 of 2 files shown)


facebook-commerce.php
```diff
@@ -123,6 +123,9 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var bool|null whether the feed has been migrated from FBE 1 to FBE 1.5 */
 	private $feed_migrated;
 
+	/** @var array the page name and url */
+	private $page;
```
unfulvio, line 127: I have a question. The `get_page()` function will try to query API if `$page` is not an array, then, if it fails, it will set this prop as an empty array. Instead of checking `! is_array()` to make an API call, would it make more sense to check for empty so there could be a retry? Or is this intended so in a single thread there's only a single try? It sounds ok the way it is now, but I wanted to double check if all this is intentional.

wvega: Good question. I wanted to avoid additional API requests in the same thread/web request, even if the first attempt failed.

I believe that if it fails, is less likely to recover from the error with more attempts, assuming that what causes the request to fail is an expired token, incorrect page ID, or a rate limit error.

Maybe we can add a comment to make that intention more explicit?

unfulvio: sure thing, after that I believe you could merge this -- thanks!

wvega: Thank you for taking a look! I added a note in 60eedbe.

PR #1350, opened 2020-05-22
Send item updates using the Catalog Batch API

wvega:
# Summary

This PR adds the `send_item_updates()` method to the sync background handler.

## Details

If you have a connected plugin, you should now be able to sync product variations using the sync background handler.

## Setup

1. You can also use a local version of the proxy with your own Facebook App, as described in https://github.com/skyverge/wc-plugins/wiki/Facebook-for-WooCommerce

###  Sync a variable product

1. Edit a variable product that is alredy synced or create a new one
1. Click Update
	- [x] A background job is created and executed (options that start with `wc_facebook_background_product_sync_job_`)
	- [x] The product is updated in the Catalog

### Tests

- [x] `vendor codecept run integration` pass


includes/Products/Sync/Background.php
```diff
@@ -313,10 +339,17 @@ private function process_item_delete( $product ) {
 	 *
 	 * @since 2.0.0-dev.1
 	 *
-	 * @param $requests
+	 * @param \stdClass|object $job
```
deivamagalhaes, line 342:
```suggestion
```

PR #1353, opened 2020-05-23
Add a notice for invalid connections

ChaseWiseman:
# Summary

Adds a notice whenever the Facebook API returns an error code that determines the access token is invalid.

## Details

Whenever an API request returns an error with a code that [requires a new access token](https://developers.facebook.com/docs/graph-api/using-graph-api/error-handling#errorcodes), we set a transient to use as a flag.

When the transient is present, display a notice to prompt reconnection.

Whenever the API returns an error code that _isn't_ OAuth related, the transient is deleted since the connection isn't the problem. Likewise, whenever an API request succeeds, the transient is deleted.

## UI Changes

- New notice: https://cloud.skyver.ge/04uPXQlx

## QA

- [ ] Acceptance tests pass

## Before merge

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -188,6 +188,7 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 			if ( get_transient( 'wc_facebook_connection_invalid' ) && $this->get_connection_handler()->is_connected() ) {
 
 				$message = sprintf(
+					/* translators: Placeholders: %1$s - <strong> tag, %2$s - </strong> tag, %3$s - <a> tag, %4$s - </a> tag */
 					__( '%1$sHeads up!%2$s Your connection to Facebook is no longer valid. Please %3$sclick here%4$s to securely reconnect your account and continue syncing products.', 'facebook-for-woocommerce' ),
 					'<strong>', '</strong>',
 					'<a href="' . esc_url( $this->get_connection_handler()->get_connect_url() ) . '">', '</a>'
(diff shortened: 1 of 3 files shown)


class-wc-facebookcommerce.php
```diff
@@ -183,6 +183,20 @@ public function add_admin_notices() {
 					'notice_class' => 'notice-info wc-facebook-migrate-notice',
 				] );
 			}
+
+			// if the connection is otherwise invalid, but there is an access token
+			if ( get_transient( 'wc_facebook_connection_invalid' ) && $this->get_connection_handler()->is_connected() ) {
+
+				$message = sprintf(
+					__( '%1$sHeads up!%2$s Your connection to Facebook is no longer valid. Please %3$sclick here%4$s to securely reconnect your account and continue syncing products.', 'facebook-for-woocommerce' ),
```
deivamagalhaes, line 191: Should we be adding translator comments?

includes/API.php
```diff
@@ -86,8 +86,23 @@ protected function do_post_parse_response_validation() {
 				throw new API\Exceptions\Request_Limit_Reached( $message, $response->get_api_error_code() );
 			}
 
+			/**
+			 * Handle invalid token errors
+			 *
+			 * @link https://developers.facebook.com/docs/graph-api/using-graph-api/error-handling#errorcodes
+			 */
+			if ( in_array( $response->get_api_error_code(), [ 10, 102, 190 ], false ) ) {
```
deivamagalhaes, line 94: From that table, 10 and 200-299 look like they may mean the same thing. Should we included 200-299 here?

ChaseWiseman: Probably a good idea! Added 62b2a56face6a4382942192e50e874e05e829ad4

class-wc-facebookcommerce.php
ChaseWiseman: Good catch, added 27f90385076d771a97667df8129c96f9d5762e50

PR #1362, opened 2020-05-25
Do not log an error when product is null on add to cart redirect

unfulvio:
## Summary

Removes logging an error when a `WC_Product` is `null` while adding to cart upon redirect.

## Story:

- [ch55122](https://app.clubhouse.io/skyverge/story/55122/woocommerce-filter-is-causing-huge-amount-of-logging-6)

## Additional details

 [This filter](https://github.com/woocommerce/woocommerce/blob/f2cce7f03780b2843f6b1b0e798dd952d51afb53/includes/class-wc-frontend-scripts.php#L559-L559) may pass `null` in some circumstances. As fallback we can look for the `$_GET` value `add-to-cart` which carries the product ID. We can safely remove logging for now.

Targeting `master` as not part of v2.0 release.

## QA

- n/a

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -386,8 +386,8 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 			if ( $product instanceof \WC_Product ) {
 				WC()->session->set( 'facebook_for_woocommerce_last_product_added_to_cart', $product->get_id() );
-			} else {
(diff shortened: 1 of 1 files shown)


wvega:
@unfulvio leaving this one unmarked in case we want to deploy 1.11.4 while we wait for 2.0 to be approved.

If we decide to release, we may need to cherry pick these changes back into the private repo as I'm not sure we have a 1.11.3 point to branch from here.

PR #1363, opened 2020-05-25
Add a method to initiate full catalog sync

wvega:
# Summary

This PRs adds the `Sync::create_or_update_all_products()` method to add all products with sync enabled to the queue of products to sync.

## Setup

1. You can also use a local version of the proxy with a personal Facebook App, as described in https://github.com/skyverge/wc-plugins/wiki/Facebook-for-WooCommerce

## Steps

1. Connect your store to facebook using version 1.11.x or older
1. Upgrade to this branch
1. Use the link in the reconnect notice to connect your store using the OAuth flow
    - [x] A background job is created to create or update all products in the catalog

### Tests


includes/Products/Sync.php
```diff
@@ -58,6 +57,66 @@ public function add_hooks() {
...
+			if ( 'product_variation' === get_post_type( $post_id ) ) {
+				$parent_product_ids[] = $post_parent;
+			}
+		}
+
+		// remove parent products because those can't be represented as Product Items
+		$product_ids = array_diff( $product_ids, $parent_product_ids );
+
+		// make sure the product should be synced and add it to the sync queue
+		foreach ( $product_ids as $product_id ) {
```
deivamagalhaes, line 98: This should also account for virtual products and variations.

ChaseWiseman: Added 3e42528281ecc50e9be67eb8b57aecbdfc21004c to account for virtual, confirmed no virtual products or variations are sent.

PR #1364, opened 2020-05-25
Fix background sync delete

wvega:
# Summary

This PRs updates the Sync Background handler to expect the retailer ID instead of the product ID for product delete requests.

## Details

Additionally, this PR updates `::on_product_delete` to give specal treatment to product variations. If a variable product is deleted, the method is called both for the product variations and the variable product. We were previously treating each variation as a simple product: trying to delete then while the page was being processed, using an API request for each variation.

## Setup

1. You can also use a local version of the proxy with a personal Facebook App, as described in https://github.com/skyverge/wc-plugins/wiki/Facebook-for-WooCommerce

### Trash a variable product

1. Move to Trash a variable product that is synced
    - [x] One or more batch requests are made to update the products setting the visibility to `staging`
1. Delete the variable product permanently
    - [x] One or more batch requests are made to delete the products

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -4276,6 +4276,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		if ( $product_group_id ) {
 
+			// TODO: replace with a call to API::delete_product_group() {WV 2020-05-26}
 			$pg_result = $this->fbgraph->delete_product_group( $product_group_id );
 
 			\WC_Facebookcommerce_Utils::log( $pg_result );
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -959,9 +959,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		} elseif ( $product->is_type( 'variable' ) ) {
 
-			$product_ids = array_merge( [ $product_id ], $product->get_children() );
-
-			facebook_for_woocommerce()->get_products_sync_handler()->delete_products( $product_ids );
+			facebook_for_woocommerce()->get_products_sync_handler()->delete_products( $product->get_children() );
 
 		} else {
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -954,12 +954,25 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		if ( $product->is_type( 'variation' ) ) {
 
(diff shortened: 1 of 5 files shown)


facebook-commerce.php
```diff
@@ -952,29 +952,34 @@ public function on_product_delete( $product_id ) {
...
+				$variation = wc_get_product( $variation_id );
 
-		$fb_product_group_id = $this->get_product_fbid( self::FB_PRODUCT_GROUP_ID, $product_id );
+				if ( $variation instanceof \WC_Product ) {
+					$retailer_ids[] = \WC_Facebookcommerce_Utils::get_fb_retailer_id( $variation );
+				}
+			}
 
-		if ( $fb_product_group_id ) {
+			facebook_for_woocommerce()->get_products_sync_handler()->delete_products( $retailer_ids );
```
deivamagalhaes, line 975:
```suggestion
                         // enqueue variations to be deleted in the background
			facebook_for_woocommerce()->get_products_sync_handler()->delete_products( $retailer_ids );
```

deivamagalhaes: Spacing may be off on my suggestion...

facebook-commerce.php
```diff
@@ -4256,6 +4261,27 @@ function delete_product_item( $wp_id ) {
...
+	 *
+	 * @param int $product_id product ID
+	 */
+	private function delete_product_group( $product_id ) {
+
+		$product_group_id = $this->get_product_fbid( self::FB_PRODUCT_GROUP_ID, $product_id );
+
+		if ( $product_group_id ) {
+
+			$pg_result = $this->fbgraph->delete_product_group( $product_group_id );
```
deivamagalhaes, line 4278: Should we not be using our API to delete the product group?

facebook-commerce.php
wvega: The suggestion looks good. I tweaked the spacing and committed it.

facebook-commerce.php
wvega: I think can do that now, but I never added stories for replacing those calls due to the limited. time we have and the things we still have left to do, so I've been replacing only what is currently causing other issues.

I added a TODO note in af3b7c1. There are other usages of the old API methods, so is probably better to clean those up at the same time later.

What do you think?

PR #1375, opened 2020-05-27
S2S: Update the build_event method

deivamagalhaes:
# Summary

This PRs updates the build_event method.

## Details

Updates the build_event method to:
- unset the `$params['event_name']` if needed
- send the `$params['custom_data']` if provided
- send the `$params['event_id']` if provided

## Open questions

- [ ] Can the `event_name` and `event_id` be sent inside the `custom_data` element as well? Right now the code only accounts for it outside of the `custom_data` element.

## QA

- [ ] Integration tests pass

### DIFF
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -398,8 +398,14 @@ if ( ! class_exists( 'WC_Facebookcommerce_Pixel' ) ) :
 		 */
 		public static function build_event( $event_name, $params, $method = 'track' ) {
 
+			// do not send the event name in the params
 			if ( isset( $params['event_name'] ) ) {
 				unset( $params['event_name'] );
+			}
+
+			// if custom data is set, send only the custom data
+			if ( isset( $params['custom_data'] ) ) {
+				$params = $params['custom_data'];
 			}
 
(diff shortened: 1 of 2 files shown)


facebook-commerce-pixel-event.php
```diff
@@ -398,14 +398,52 @@ function handle<?php echo $event_name; ?>Event() {
...
-				"fbq('%s', '%s', %s);",
-				WC_Facebookcommerce_Utils::getIntegrationName(),
-				esc_js( $method ),
-				esc_js( $event_name ),
-				json_encode( self::build_params( $params, $event_name ), JSON_PRETTY_PRINT | JSON_FORCE_OBJECT )
-			);
+			// do not send the event name in the params
+			if ( isset( $params['event_name'] ) ) {
+
+				$event_name = $params['event_name'];
```
ChaseWiseman, line 404: Should we be overriding the `$event_name` value that's passed to the method here? This causes the "event_name" test case to fail.

deivamagalhaes: You're right, we always pass the event_name as the first param to the inject methods, so we really don't need to override it here.

PR #1383, opened 2020-05-28
Hook up the sync button to the new batch sync method

ChaseWiseman:
# Summary

Updates the "Sync products" button to use the new batch full-catalog sync method.

## UI Changes

N/A - this is the same display and behavior as before, just with different background processing.

## QA

- Connected to a catalog
- Debug logging on

1. Click "Sync products"
    - [x] The sync is initiated
    - [x] The settings UI is blocked
    - [x] Batch requests are made
1. Wait for the sync to finish
    - [x] The settings are unblocked

## Before merge


assets/js/admin/facebook-for-woocommerce-settings-sync.js
```diff
@@ -147,4 +147,137 @@ jQuery( document ).ready( function( $ ) {
...
+		if ( confirm( facebook_for_woocommerce_settings_sync.i18n.confirm_sync ) ) {
+
+			syncInProgress();
+
+			$.post( facebook_for_woocommerce_settings_sync.ajax_url, {
+				action: 'wc_facebook_sync_products',
+				nonce:  facebook_for_woocommerce_settings_sync.sync_products_nonce,
+			}, function ( response ) {
+
+				console.log( response );
```
deivamagalhaes, line 169: Do we want to always log the response?

assets/js/admin/facebook-for-woocommerce-settings-sync.js
```diff
@@ -147,4 +147,137 @@ jQuery( document ).ready( function( $ ) {
...
+		if ( ! facebook_for_woocommerce_settings_sync.sync_in_progress ) {
+			return;
+		}
+
+		$.post( facebook_for_woocommerce_settings_sync.ajax_url, {
+			action: 'wc_facebook_get_sync_status',
+			nonce:  facebook_for_woocommerce_settings_sync.sync_status_nonce,
+		}, function ( response ) {
+
+			console.log( response );
```
deivamagalhaes, line 269: Do we want to always log the response?

assets/js/admin/facebook-for-woocommerce-settings-sync.js
```diff
@@ -147,4 +147,137 @@ jQuery( document ).ready( function( $ ) {
...
+
+			if ( response.success ) {
+
+				if ( response.data > 0 ) {
```
deivamagalhaes, line 273: Since the action is not very clear about it, can we add a comment saying the response contains the number of remaining items?

includes/AJAX.php
```diff
@@ -33,6 +35,56 @@ public function __construct() {
...
+	 * @internal
+	 *
+	 * @since 2.0.0-dev.1
+	 */
+	public function get_sync_status() {
+
+		check_admin_referer( Product_Sync::ACTION_GET_SYNC_STATUS, 'nonce' );
+
+		$remaining_products = 0;
+		$jobs               = facebook_for_woocommerce()->get_products_sync_background_handler()->get_jobs( [
```
deivamagalhaes, line 76:
```suggestion
		$remaining_products = 0;
		
		$jobs = facebook_for_woocommerce()->get_products_sync_background_handler()->get_jobs( [
```

includes/AJAX.php
```diff
@@ -33,6 +35,56 @@ public function __construct() {
...
+			'status' => 'processing',
+		] );
+
+		if ( ! empty( $jobs ) ) {
+
+			foreach ( $jobs as $job ) {
+				$remaining_products = $job->total - $job->progress;
```
deivamagalhaes, line 83: Should we add to the `$remaining_products` or are we intentionally only getting the latest? If that is the case, it would be better to just get the latest instead of iterating all processing jobs.

includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -59,19 +66,22 @@ public function enqueue_assets() {
...
 			'excluded_tag_ids'                => facebook_for_woocommerce()->get_integration()->get_excluded_product_tag_ids(),
 			'i18n'                            => [
 				/* translators: Placeholders %s - html code for a spinner icon */
 				'confirm_resync'                => esc_html__( 'Your products will now be resynced to Facebook, this may take some time.', 'facebook-for-woocommerce' ),
 				'confirm_sync'                  => esc_html__( "Facebook for WooCommerce automatically syncs your products on create/update. Are you sure you want to force product resync?\n\nThis will query all published products and may take some time. You only need to do this if your products are out of sync or some of your products did not sync.", 'facebook-for-woocommerce' ),
-				'sync_in_progress'              => sprintf( esc_html__( 'Syncing... Keep this browser open until sync is complete. %s', 'facebook-for-woocommerce' ), '<span class="spinner is-active"></span>' ),
+				'sync_in_progress'              => sprintf( esc_html__( 'Syncing... %s', 'facebook-for-woocommerce' ), '<span class="spinner is-active"></span>' ),
```
deivamagalhaes, line 84: Maybe we could add a tooltip or something so that the users know they can close the page now. Just a thought.

assets/js/admin/facebook-for-woocommerce-settings-sync.js
ChaseWiseman: yeah I think that's fine since we're in the admin context

deivamagalhaes:
The progress is lovely! 
https://cloud.skyver.ge/9ZuE5KZZ

assets/js/admin/facebook-for-woocommerce-settings-sync.js
ChaseWiseman: sure! good idea

includes/AJAX.php
ChaseWiseman: good call - in fact there is also a case where the total & progress may not be set yet. I tweaked things here to be a little more solid: dfc7f918063ea62271ac20b711768aa01d66ba4b

includes/Admin/Settings_Screens/Product_Sync.php
ChaseWiseman: Yeah that may be a nice enhancement - opened a story for the future https://app.clubhouse.io/skyverge/story/55573/let-folks-know-they-can-leave-the-sync-page-now

includes/AJAX.php
```diff
@@ -80,9 +80,14 @@ public function get_sync_status() {
 
 		if ( ! empty( $jobs ) ) {
 
-			foreach ( $jobs as $job ) {
-				$remaining_products = $job->total - $job->progress;
+			$job = $jobs[0];
```
deivamagalhaes, line 83: Maybe add a comment saying we chose to display the first job?

PR #1392, opened 2020-05-29
S2S: Adjust the contents params for multi product events

ChaseWiseman:
# Summary

Adds the `contents` param to `Search`, `ViewContent`, and `ViewCategory`.

Adds the `content_category` param to `InitiateCheckout` when there is only one item in the cart.

## Details

See [this discussion](https://skyverge.slack.com/archives/CR8VDB4G7/p1590772355255300) for details on approach.

### Setup

- Enable the Facebook Pixel Helper extension
- Click Troubleshoot Pixel and start listening to Pixel events
- Enable logging for the plugin

### Steps

1. Search for a term that will return products
    - [ ] The `Search` event contains the `contents` param with the product results
1. Click on one of the products
    - [ ] The `ViewContent` event contains the `contents` param
1. Add the product to cart
1. Visit checkout
    - [ ] The `InitiateCheckout` event contains `content_category` with the product's first category name
1. Add a different product to cart & visit checkout
    - [ ] The `InitiateCheckout` event does not contain the `content_category` event

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -591,7 +591,9 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 					$categories = \WC_Facebookcommerce_Utils::get_product_categories( $item['data']->get_id() );
 
-					$event_data['custom_data']['content_category'] = $categories['name'];
+					if ( ! empty( $categories['name'] ) ) {
+						$event_data['custom_data']['content_category'] = $categories['name'];
+					}
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -559,6 +582,19 @@ public function inject_initiate_checkout_event() {
 				],
 			];
 
+			// if there is only one item in the cart, send its first category
+			if ( ( $cart = WC()->cart ) && count( $cart->get_cart() ) === 1 ) {
```
deivamagalhaes, line 586: What if there are more items, but only one category? And not set on the very first item?

ChaseWiseman: Handling multiple items with the same category will probably have to be out of scope for now given the timeframe. I don't think a product can have no category, but I added a check just in case f56841b185f682080ce551dc4756c6b0c7b52818

PR #1393, opened 2020-05-29
S2S: Handle the Purchase event

deivamagalhaes:
# Summary

This PRs updates the `WC_Facebookcommerce_EventsTracker::inject_purchase_event()` method to send the event server-side and add the event ID to the JS script.

### Setup

- Enable the Facebook Pixel Helper extension
- Click Troubleshoot Pixel and start listening to Pixel events
- Enable logging for the plugin

### Steps

1. Add the subscription product to the cart
1. Proceed to Checkout
1. Place the order
    - [x] An API call is made to create a `Purchase` event, with an event ID
    - [x] The corresponding JS event is triggered with the same event ID (check in the Events Manager or Pixel Helper)
1. Refresh the Thank You page
    - [x] No API call is made
    - [x] The corresponding JS event is not triggered

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -585,7 +585,9 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		 */
 		public function inject_purchase_event( $order_id ) {
 
-			if ( ! self::$isEnabled || $this->pixel->is_last_event( 'Purchase' ) ) {
+			$event_name = 'Purchase';
+
+			if ( ! self::$isEnabled || $this->pixel->is_last_event( $event_name ) ) {
 				return;
 			}
 
@@ -639,7 +641,7 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 				}
 			}
 
-			$this->pixel->inject_event( 'Purchase', [
+			$this->pixel->inject_event( $event_name, [
 				'num_items'    => $num_items,
 				'content_ids'  => wp_json_encode( array_merge( ... $product_ids ) ),
 				'contents'     => wp_json_encode( $contents ),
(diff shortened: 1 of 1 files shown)


deivamagalhaes:
@ChaseWiseman, I am probably missing something, but I do not think we need a different meta for the S2S call. Do you have any suggestions on how to trigger the duplicate event server-side?

ChaseWiseman:
@deivamagalhaes I think the idea of that was to trigger the S2S event on `woocommerce_checkout_update_order_meta` and store the event ID, then let `woocommerce_thankyou` find that meta and inject the JS event with it. It looks like in this PR both the S2S and JS events get fired on `woocommerce_thankyou` page output since the method returns early on `woocommerce_checkout_update_order_meta`. Did you find that's a solid enough solution? I did not see duplicates in my tests, FWIW.

deivamagalhaes:
@ChaseWiseman , I think so, but I am not sure where the original idea is coming from. Do we not show the Thank You page for some purchases?

wvega:
@ChaseWiseman @deivamagalhaes this is what I remember from the time Fulvio and I worked on a related PR:

The events were being injected in the Thank You page for all type of purchases (credit card and cash on delivery for example) because the other action was fired during Ajax, so injecting the JS script had no effect.

I think that approach has worked fine so far and it likely continues to work well in this case, as all kind of purchases normally end up in the Thank you page.

With server-side event, however, we can bypass the limitation of the other action firing during Ajax, since we can send the event using the API at the right time. That's the approach that uses Google Analytics Pro and seems to be more future-proof in case a plugin redirects to a different place (although I don't have specific examples).

So I think the end goal is to separate them, but perhaps is not a blocker at this point. If we keep them together or the action associated with the S2S event only fires once per order, then the other meta doesn't seem necessary after all.

PR #1396, opened 2020-05-29
Clean up integration tab and related code

deivamagalhaes:
# Summary

This PRs removes the Facebook tab from WC Settings > Integration and related code.

## QA

1. Navigate to the WC Settings > Integration
    - [x] There is no Facebook section

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -32,9 +32,6 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 		/** @var string the integration ID */
 		const INTEGRATION_ID = 'facebookcommerce';
 
-		/** @var string the integration class name (including namespaces) */
-		const INTEGRATION_CLASS = '\\WC_Facebookcommerce_Integration';
-
 
 		/** @var \WC_Facebookcommerce singleton instance */
 		protected static $instance;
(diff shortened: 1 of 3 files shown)


class-wc-facebookcommerce.php
```diff
@@ -124,9 +124,6 @@ public function init() {
 					$this->ajax = new \SkyVerge\WooCommerce\Facebook\AJAX();
 				}
 
-				// register the WooCommerce integration
-				add_filter( 'woocommerce_integrations', [ $this, 'add_woocommerce_integration' ] );
```
ChaseWiseman, line 128: I think we need to replace this with a call to `get_integration()` within `init` - `woocommerce_integrations` runs on init, and it was instantiated there. That means the integration and its hooks will only load the first time we call `->get_integration()` elsewhere in the plugin

deivamagalhaes: I get into a circular reference if I try that. Doesn't the check inside `get_integration()` already take care of only doing it once?

deivamagalhaes: Ok, I think I got it!

ChaseWiseman: It needs to be run inside `init`, like `add_action( 'init', function() { $this->get_integration() } );` or similar. That's to keep it loaded at the same time it would load with `woocommerce_integrations`.

The problem isn't doing it once, the problem is that previously the instance would always load on `init`, but if we don't do that then it won't be instantiated until something else calls `->get_instance()` in the plugin.

PR #1402, opened 2020-06-04
Draft variations should not be synced

deivamagalhaes:
# Summary

This PR fixes an issue that is causing variations of draft products to be synced and visible.

## Details

When a variable product is duplicated, the variation posts are created with "publish" status. So when checking the status to get the products, we need to check the parent status as well, for variations.

A pre-release was sent to the customer.

## UI Changes

N/A.

### Setup

- The plugin is connected and product sync is enabled

### Steps

1. Create a variable product with variations with prices
1. Duplicate this product and save the new one as a draft (do not publish it)
1. Sync all your products/regenerate your feed
    - [ ] The new draft product is not synced (not included in the feed)
1. Publish the product
1. Sync all your products/regenerate your feed
    - [ ] The product is synced (included in the feed)

## Before merge

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Facebook for WooCommerce Changelog ***
 
+2020.nn.nn - version 1.11.4-dev.1
+ * Fix - Do not sync variations for draft variable products created by duplicating products
+
 2020.05.20 - version 1.11.3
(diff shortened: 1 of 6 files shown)


wvega:
@deivamagalhaes I think we can include https://github.com/facebookincubator/facebook-for-woocommerce/pull/1362 in 1.11.4 too, but I need to rebase those changes first.

PR #1412, opened 2020-06-12
Modifications necessary to create a System User

wvega:
# Summary

This PR adds the necessary modifications to get a system user access token from the connect app and use that new token for API requests.

## Details

1. We now include in the external business ID in the redirect URL sent as the `state` parameter.
    
    We need to update the proxy app to create a System User before the user is redirected back to the website. The external business ID is used to retrieve the Business Manager ID connected to the FB Extension, which is a required parameter for the API requests that create the System User. It's also used to retrieve the IDs of the assets associated with the current installation.
2. The connect app now sends back the System User ID, the System User access token, and the merchant access token. This PR includes modifications to store the new values.
3. The System User access token is stored in the OPTION_ACCESS_TOKEN option and is used for most API requests.
4. The merchant access token is currently required to retrieve the IDs of the assets associated with the current installation.

## Open Questions

**Do we really need the merchant access token to retrieve the installation assets IDs?**

I [asked Connor](https://a8c.slack.com/archives/C010ND8MRPU/p1592418973316500?thread_ts=1591237950.248100&cid=C010ND8MRPU) whether is possible to use the System User access token to retrieve the installation assets IDs. If we can't, we would need to update `Connect::refresh_installation_data()` to use a different access token.

### Setup

1. Make sure the plugin is configured via snippets to connect through a local proxy running the changes from https://github.com/skyverge/woocommerce-connect/pull/4

#### Connection

1. Connect your store
    - [x] The connection flow completes successfully

#### Products sync

1. Edit a product that has Facebook sync enabled
    - [x] The product is successfully updated in the Catalog
1. Edit a variable product that has Facebook sync enabled
    - [x] The product variations are updated using the Catalog Batch API
    - [x] The product variations are successfully updated in the Catalog

#### S2S

1. Add a product to cart
    - [x] The logs include a successful S2S API request for the AddToCart event
1. Go to Checkout
    - [x] The logs include a successful S2S API request for the InitiateCheckout event
1. Place the order
    - [x] The logs include a successful S2S API request for the Purchase event

## Before merge

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -165,6 +165,7 @@ class Connection {
 			}
 
 			$this->update_access_token( $system_user_access_token );
+			$this->update_merchant_access_token( $merchant_access_token );
 			$this->update_system_user_id( $system_user_id );
 
 			$api = new \WC_Facebookcommerce_Graph_API( $merchant_access_token );
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -43,10 +43,10 @@ class Connection {
 	/** @var string the business manager ID option name */
 	const OPTION_BUSINESS_MANAGER_ID = 'wc_facebook_business_manager_id';
 
-	/** @var string the system user access token option name */
+	/** @var string the system user ID option name */
 	const OPTION_SYSTEM_USER_ID = 'wc_facebook_system_user_id';
 
-	/** @var string the access token option name */
+	/** @var string the system user access token option name */
 	const OPTION_ACCESS_TOKEN = 'wc_facebook_access_token';
 
 	/** @var string the merchant access token option name */
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -168,6 +168,7 @@ class Connection {
 			$this->update_merchant_access_token( $merchant_access_token );
 			$this->update_system_user_id( $system_user_id );
 
+			// TODO: use the system user access token once the fix for the fbe_installs endpoint is approved and deployed {WV 2020-06-17}
 			$api = new \WC_Facebookcommerce_Graph_API( $merchant_access_token );
 
 			$asset_ids = $api->get_asset_ids( $this->get_external_business_id() );
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -158,9 +158,14 @@ class Connection {
 				throw new SV_WC_API_Exception( 'System User access token is missing' );
 			}
 
-			$access_token = $this->create_system_user_token( $access_token );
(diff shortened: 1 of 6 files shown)


wvega:
https://app.clubhouse.io/skyverge/story/55664/implement-system-user-access-token-exchange

includes/Handlers/Connection.php
```diff
@@ -43,9 +43,14 @@ class Connection {
 	/** @var string the business manager ID option name */
 	const OPTION_BUSINESS_MANAGER_ID = 'wc_facebook_business_manager_id';
 
+	/** @var string the system user access token option name */
```
deivamagalhaes, line 46:
```suggestion
	/** @var string the system user ID option name */
```

includes/Handlers/Connection.php
```diff
@@ -43,9 +43,14 @@ class Connection {
...
+	const OPTION_SYSTEM_USER_ID = 'wc_facebook_system_user_id';
+
 	/** @var string the access token option name */
```
deivamagalhaes, line 49:
```suggestion
	/** @var string the system user access token option name */
```

includes/Handlers/Connection.php
```diff
@@ -141,17 +146,29 @@ public function handle_connect() {
...
-			$api = new \WC_Facebookcommerce_Graph_API( $access_token );
+			if ( ! $system_user_id ) {
+				throw new SV_WC_API_Exception( 'System User ID is missing' );
+			}
+
+			$this->update_access_token( $system_user_access_token );
+			$this->update_merchant_access_token( $merchant_access_token );
+			$this->update_system_user_id( $system_user_id );
+
+			$api = new \WC_Facebookcommerce_Graph_API( $merchant_access_token );
```
deivamagalhaes, line 171: I am probably confused, but I thought we would use the system user access token for everything. 🤔

deivamagalhaes: May be worth adding a comment above this line explaining why we use this token here, for visibility.

wvega: You're right @deivamagalhaes , the goal is to use the system user access token, however, the API request to get assets IDs currently does not accept that token. I asked Connor and he was able to find a problem with the way permissions were being checked on that endpoint.

A fix is currently [being reviewed](https://a8c.slack.com/archives/C010ND8MRPU/p1592438962330500?thread_ts=1591237950.248100&cid=C010ND8MRPU), so we should be able to switch that line to use the system user access token soon.

I added a TODO note for that: 3cf2082

wvega:
Thank you for the review @deivamagalhaes. I committed the suggestions and added a note about why we are currently not using the system user access token for that request.

Issue #33, opened 2017-09-10
Japanese characters in variations not working


jlev00:
That should read "200錠" and "50錠"

![image](https://user-images.githubusercontent.com/28719041/30245371-0ed1f21a-961b-11e7-8824-bb91cbe6ecf3.png)

tahoff:
Thanks for reporting this issue. I have a few follow-up questions because I'm unable to reproduce this issue.
- What version of the extension are you using?
- Do you have any internationalization/translation extension which you've used to create the above strings?

This is what I get when I copy & paste the strings you listed above:
![screen shot 2017-09-11 at 1 21 58 pm](https://user-images.githubusercontent.com/6645131/30295238-c1724c92-96f4-11e7-9c9f-8165ddf33e9a.png)

jlev00:
Hi Adam - thanks for taking a look at this.

v1.4.5

I'm pulling the strings into Woocommerce using another module (All WP Import Pro/Woocommerce) so that's definitely a possibility, although they appear fine in backend/frontend in Woocommerce, and in the database. 

Just to make sure, you created the variations in the Woocommerce backend? 

If you wanted to test whether the attribute name was affecting things you can copy-paste this: 
 
サイズ

Happy to help any way I can, and am open to the possibility that it's something odd with my setup. Let me know if you need links to specific products/etc.

tahoff:
@jlev00 Thanks for reporting this issue! We've fixed the issue in [v1.6.0](https://github.com/facebookincubator/facebook-for-woocommerce/releases/tag/v1.6.0). Please download the new version and re-open the issue if the problem re-occurs. The new version should be available on the WooCommerce site next week.

PR #1414, opened 2020-06-17
Messenger: pull the latest remote settings

ChaseWiseman:
# Summary

Ensure the latest Messenger settings are updated on the site.

## Details

In FBE 2 we cannot update the locale, greeting, or color settings from the plugin. This PR removes those settings, and adds the ability to _pull_ the messenger configuration from Facebook. This pull occurs in two ways:

1. Once an hour, sitewide
1. Whenever the Messenger settings page is visited

We _can_ enable or disable the feature from the plugin, so that setting is preserved. Whenever that changes, the FBE configuration is updated.

The messenger JS takes care of applying the color and greeting settings that are configured in FBE for us, so we stop passing those values as attributes when initializing the JS.

The new FBE flow also has a setting for allowing specific domains. If the user accidentally removes the site's domain, the plugin re-adds it whenever pulling the latest settings.

Note: a [later story](https://app.clubhouse.io/skyverge/story/57143/facebook-for-woocommerce-notice-upon-upgrade-about-messenger-customization) will add messaging and any extra handling for upgraders from FBE 1

## QA

- Disconnect the plugin & clear all DB options

1. Connect
1. Visit the frontend
    - [x] No messenger is displayed
1. From the Connection settings tab, click "Manage Connection"
1. In the FBE UI, click "Add" next to Messenger
1. Configure some custom options, like greeting and color, and complete the wizard
1. Visit the Messenger settings page
    - [x] The Enabled & Locale items are displayed correctly
1. Visit the frontend
    - [x] The messenger is loaded with the correct customizations
1. Visit the Messenger settings again
1. Disable the messenger & save
    - [x] After reload, the setting is still disabled
1. Visit the frontend
    - [x] The messenger is not loaded
1. Enable the messenger and save again
1. Visit the frontend
    - [x] The messenger is loaded

- [x] Integration tests pass
- [x] `run acceptance Admin/Settings` tests pass

## Before merge


deivamagalhaes:
I was trying to QA this one, but I keep getting this on Manage Connection: https://cloud.skyver.ge/KouBwn61

deivamagalhaes:
@ChaseWiseman, would it be possible to add me to the FBE app? I believe this may fix this error and unblock me from QAing your PRs.

deivamagalhaes:
@ChaseWiseman , just to double-check, when I re-enable Messenger using our settings, is it supposed to have lost all the customizations (greeting and color)?

includes/API/FBE/Configuration/Update/Request.php
```diff
@@ -0,0 +1,59 @@
...
+	 *
+	 * Only the enabled and domains values are able to accept updates right now.
+	 *
+	 * @since 2.0.0-dev.1
+	 *
+	 * @param Configuration\Messenger $configuration messenger configuration object
+	 */
+	public function set_messenger_configuration( Configuration\Messenger $configuration ) {
+
+		$this->data['messenger_chat'] = [
```
deivamagalhaes, line 52: Maybe we need to merge our values with the original values from the API here, as not to reset the color/greeting.

deivamagalhaes: Or maybe only send the `enabled` and `domains` keys, not the full array.

ChaseWiseman:
@deivamagalhaes yeah once the messenger is disabled, it'll reset to the defaults. Unfortunately we can't send any color or greeting values to Facebook, so we don't have a way around that.

includes/Admin/Settings_Screens/Messenger.php
```diff
@@ -151,100 +135,179 @@ public function sanitize_messenger_greeting( $value ) {
...
-				'title'    => __( 'Enable Messenger', 'facebook-for-woocommerce' ),
-				'type'     => 'checkbox',
-				'desc'     => __( 'Enable and customize Facebook Messenger on your store', 'facebook-for-woocommerce' ),
-				'default'  => 'no',
+				'id'      => \WC_Facebookcommerce_Integration::SETTING_ENABLE_MESSENGER,
+				'title'   => __( 'Enable Messenger', 'facebook-for-woocommerce' ),
+				'type'    => 'checkbox',
+				'desc'    => __( 'Enable and customize Facebook Messenger on your store', 'facebook-for-woocommerce' ),
+				'default' => 'no',
+				'value'   => $this->remote_configuration && $this->remote_configuration->is_enabled() ? 'yes' : 'no',
```
deivamagalhaes, line 153:
```suggestion
				'value'   => $is_enabled ? 'yes' : 'no',
```

includes/API/FBE/Configuration/Read/Response.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+namespace SkyVerge\WooCommerce\Facebook\API\FBE\Configuration\Read;
+
+defined( 'ABSPATH' ) or exit;
+
+use SkyVerge\WooCommerce\Facebook\API;
+
+/**
+ * FBE Installation API read response object.
```
deivamagalhaes, line 18:
```suggestion
 * FBE Configuration API read response object.
```

includes/API/FBE/Configuration/Request.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace SkyVerge\WooCommerce\Facebook\API\FBE\Configuration;
+
+defined( 'ABSPATH' ) or exit;
+
+use SkyVerge\WooCommerce\Facebook\API;
+
+/**
+ * FBE API request object.
```
deivamagalhaes, line 18:
```suggestion
 * FBE Configuration API request object.
```

includes/Admin/Settings_Screens/Messenger.php
```diff
@@ -50,95 +55,74 @@ public function __construct() {
...
 
-		$tab = SV_WC_Helper::get_requested_value( 'tab' );
-
-		// only enqueue assets on this specific screen
-		if ( Admin\Settings::PAGE_ID !== SV_WC_Helper::get_requested_value( 'page' ) || ( $tab && self::ID !== $tab ) ) {
-			return;
-		}
-
-		wp_enqueue_script( 'wc-facebook-admin-settings-messenger', facebook_for_woocommerce()->get_plugin_url() . '/assets/js/admin/facebook-for-woocommerce-settings-messenger.min.js', [ 'jquery', 'iris', 'wc-enhanced-select' ], \WC_Facebookcommerce::VERSION );
+		// TODO: empty for now, until we add more robust Messenger settings {CW 2020-06-17}
```
deivamagalhaes, line 58: Should we remove this JS file from the repo as well?

tests/integration/API/FBE/Configuration/MessengerTest.php
```diff
@@ -0,0 +1,164 @@
+<?php
+
+namespace SkyVerge\WooCommerce\Facebook\Tests\API\FBE\Configuration;
+
+use SkyVerge\WooCommerce\Facebook\API\FBE\Configuration\Messenger;
+
+/**
+ * Tests the API\FBE\Installation\Read\Response class.
```
deivamagalhaes, line 8:
```suggestion
 * Tests the API\FBE\Configuration\Read\Response class.
```

includes/Admin/Settings_Screens/Messenger.php
ChaseWiseman: I opted to keep it, as we'll soon be revamping it.

includes/API/FBE/Configuration/Update/Request.php
wvega: I noticed this too. If you follow the steps below, the selected greeting and color are removed from the settings on Facebook.

1. Uncheck Enable Messenger
1. Save changes (at this point you can click Add again for Messenger plugin on Facebook and the settings are still there)
1. Check Enable Messenger
1. Save changes (the Messenger plugin is added again, but the color and greeting settings are reset)

ChaseWiseman:
@wvega thanks for the testing!

No a public domain isn't required for this. As odd as it sounds, I think this might be related to the system user access token connection method. I was confused because I definitely didn't have a problem loading Messenger on the frontend while testing last week, but I tried again and now I see the same - UI not loaded, only console errors.

However, the plugin was connected via the new system user access token changes in the proxy app because I was testing that PR earlier. So I disconnected, and pointed my snippet back to the hosted proxy app and Messenger started working again! Very strange because the Messenger code doesn't use the access token.

Would you be able to confirm you see the same? To recap:
1. Uninstall/disconnect the plugin
1. Change your snippet to point to the hosted connect app, not your local app
1. Reconnect and enable messenger

If you can confirm the same, it may be a bug to report with the FBE manager, maybe something like the allowed domains setting not applying for a system user access token.

Thanks!

wvega:
@ChaseWiseman I was just writing that I don't think the Content Security Policy is related with the changes in this branch because I was able to see the Messenger when I upgraded from 1.11.4 to 2.0 and connected using the new flow in #1415. That seems to support what you describe and indicate that is a FBE bug.

However, I'll give the steps you mentioned a try too and report back.

wvega:
@ChaseWiseman I can confirm that if I uninstall and remove the plugin settings, and use the connect app without the system user changes (or the hosted app), I can connect and enable the Messenger plugin without problems.

The issue seems related to having an existing connection, even if is not using a system user. One way I was able to reproduce the issue with the Content Security Policy was:

The first three steps simulate installing the plugin and having a previously stablished connection using the FBE.

1. Uninstall using the link and remove all plugin settings from the DB
1. Connect the store using the hosted connect app
1. Remove all plugin settings from the DB (but don't uninstall)
1. Connect the store again using the hosted connect app. This fails with something like `(#33) An FBE installation does not exist for the business with ID skyverge-plugins-5ef130c2293ed or you do not have permission to see it`
1. Now uninstall using the link
1. Connect the store a third time using the hosted connect app. The connection should work this time.
1. Manage the connection and enable the Messenger plugin
1. Go to the Messenger settings
1. Visit the frontend. Now the messenger UI is not shown and one of the Content Security Policy headers in the iframe response is set to `frame-ancestors https://www.facebook.com;`

I don't think the above is connected to this PR, so perhaps we should merge this one, review and merge the rest, and make sure to test several upgrade and installation paths on the base branch afterwards. What do you think?

ChaseWiseman:
@wvega thanks for trying, and good finds!

I agree - the scenario you tried seems far outside of this PR's changes, and a more general issue with how FBE is handling existing installations in some of these edge cases. The messenger part of it seems to only be a symptom of the larger issue. If otherwise looks good, I say we merge them. Your suggestion for testing these scenarios on a final base branch sounds good to me, and we can both combine and report our findings to the FB team for their investigation (for items we can't fix ourselves) when we hand it off for their testing.

PR #1420, opened 2020-06-24
Truncate product brand values to 100 characters

wvega:
# Summary

This PR truncates the brand property included in product data to be 100 characters or less.

## Details

If the brand name is more than 100 characters long, the API fails with an error like the one below:

```
{"error":{"message":"Invalid parameter","type":"OAuthException","code":100,"error_subcode":1803059,"is_transient":false,"error_user_title":"El campo es demasiado largo","error_user_msg":"El valor de brand supera los 100 caracteres.","fbtrace_id":"A-j56AbnqBPho8EqrNeitBC"}}
```

### Setup

- Set the site title to something > 100 characters, like `Supposedly there are over one million words in the English Language - We trimmed some fat to take away really odd words and determiners`
- Enable debug log for Facebook for WooCommerce

### Sync products

1. Update a simple product with sync enabled
	- [x] The product is synced successfully
	- [x] No error notices are shown
	- [x] There are no errors in the response logs

## Before merge

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -10,6 +10,7 @@
 
 use SkyVerge\WooCommerce\Facebook\Admin;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Helper;
 use SkyVerge\WooCommerce\Facebook\Products;
 use SkyVerge\WooCommerce\Facebook\Products\Feed;
 
@@ -1791,7 +1792,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 				'description'  => strip_tags( $product_data['description'] ),
 				'id'           => $product_data['retailer_id'],
 				'image_link'   => $product_data['image_url'],
-				'brand'        => strip_tags( WC_Facebookcommerce_Utils::get_store_name() ),
+				'brand'        => SV_WC_Helper::str_truncate( strip_tags( WC_Facebookcommerce_Utils::get_store_name() ), 100 ),
 				'link'         => $product_data['url'],
 				'price'        => $product_data['price'] . ' ' . get_woocommerce_currency(),
 			];
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -10,7 +10,6 @@
 
 use SkyVerge\WooCommerce\Facebook\Admin;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
-use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Helper;
 use SkyVerge\WooCommerce\Facebook\Products;
 use SkyVerge\WooCommerce\Facebook\Products\Feed;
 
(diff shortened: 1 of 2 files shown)


facebook-commerce.php
```diff
@@ -1791,7 +1792,7 @@ function get_sample_product_feed() {
 				'description'  => strip_tags( $product_data['description'] ),
 				'id'           => $product_data['retailer_id'],
 				'image_link'   => $product_data['image_url'],
-				'brand'        => strip_tags( WC_Facebookcommerce_Utils::get_store_name() ),
+				'brand'        => SV_WC_Helper::str_truncate( strip_tags( WC_Facebookcommerce_Utils::get_store_name() ), 100 ),
```
unfulvio, line 1795:
```suggestion
				'brand'        => Framework\SV_WC_Helper::str_truncate( strip_tags( WC_Facebookcommerce_Utils::get_store_name() ), 100 ),
```

includes/fbproduct.php
```diff
@@ -9,6 +9,7 @@
  */
 
 use SkyVerge\WooCommerce\Facebook\Products;
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Helper;
```
unfulvio, line 12:
```suggestion
use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
```

includes/fbproduct.php
```diff
@@ -555,7 +556,7 @@ public function prepare_product( $retailer_id = null, $prepare_for_product_feed
 				'additional_image_urls' => array_slice( $image_urls, 1 ),
 				'url'                   => $product_url,
 				'category'              => $categories['categories'],
-				'brand'                 => $brand,
+				'brand'                 => SV_WC_Helper::str_truncate( $brand, 100 ),
```
unfulvio, line 559:
```suggestion
				'brand'                 => Framework\SV_WC_Helper::str_truncate( $brand, 100 ),
```

no need to strip tags here?

facebook-commerce.php
```diff
@@ -10,6 +10,7 @@
 
 use SkyVerge\WooCommerce\Facebook\Admin;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Helper;
```
unfulvio, line 13:
```suggestion
```

includes/fbproduct.php
wvega: Good catch! I tested and WordPress allows tags in the Site Title setting so it makes sense to strip the tags here as well.

wvega:
Thank you for the review @unfulvio. I committed the suggestions and added code to strip tags from the store name.

unfulvio:
thanks @wvega I see in one part we use `strip_tags()` in another `wp_strip_all_tags()` the main difference should be that the former removes text elements in `<script>` and `<style>` while `strip_tags()` preserves it. If I'm not mistaken, `wp_strip_all_tags()` also removes shortcodes?

wvega:
@unfulvio I let this one fall between the cracks. I'm sorry for the long delay in responding. 

I tried no to touch `get_sample_product_feed()` (where `strip_tags()` is used) too much because I think we are going to end up removing that method in the short future. Is not really used in 2.0.0.

However, to keep this PR consistent, I updated that method to use `wp_strip_all_tags()` too. The difference is that `wp_strip_all_tags()` transforms `<script>something</script>` into an empty string, while `strip_tags()` returns `something`. In this case we are using the function to clean up the store name so it's unlikely that someone enters any tags in there, but if they do, this should do the trick.

PR #1422, opened 2020-06-28
Add tests for existing sync behavior

wvega:
# Summary

This PR adds tests for existing behavior of some of the methods that sync products, either with direct requests or using sync background handler.

## Details

The goal of [CH 56109](https://app.clubhouse.io/skyverge/story/56109) is to update the different methods that sync products to use the same logic to decide whether a product or product variation should be synced or not. The tests here should reduce the risk of changing several of those methods at once, by ensuring those behaviors are preserved after the changes in the next PR.

## QA

- [x] Code review


tests/integration/WC_Facebookcommerce_Integration_Test.php
```diff
@@ -771,8 +773,396 @@ public function test_is_product_sync_enabled_filter() {
...
+
+	/** @see \WC_Facebookcommerce_Integration::on_product_delete() */
+	public function test_on_product_delete_with_simple_product() {
+
+		$product = $this->tester->get_product( [ 'status' => 'publish' ] );
+
+		$integration = $this->make( \WC_Facebookcommerce_Integration::class, [
+			'delete_product_item'  => Expected::once(),
+			// called from delete_product_item() and from delete_product_group()
+			'get_product_fbid'     => null,
```
deivamagalhaes, line 785: Should we maybe explicitly expect `delete_product_group` to be called?

tests/integration/WC_Facebookcommerce_Integration_Test.php
```diff
@@ -771,8 +773,396 @@ public function test_is_product_sync_enabled_filter() {
...
+	public function test_on_product_delete_with_variable_product() {
+
+		$product = $this->tester->get_variable_product( [
+			'children' => 3,
+			'status'   => 'publish',
+		] );
+
+		$integration = $this->make( \WC_Facebookcommerce_Integration::class, [
+			'delete_product_item' => Expected::never(),
+			// called from delete_product_group()
```
deivamagalhaes, line 802: Should we maybe explicitly expect `delete_product_group` to be called?

tests/integration/WC_Facebookcommerce_Integration_Test.php
```diff
@@ -771,8 +773,396 @@ public function test_is_product_sync_enabled_filter() {
...
+
+
+	/**
+	 * Tests that on_product_publish() delegates to the appropriate handler.
+	 *
+	 * @see \WC_Facebookcommerce_Integration::on_product_publish()
+	 *
+	 * @param \WC_Product $product product object
+	 */
+	private function check_on_product_publish( \WC_Product $product ) {
```
deivamagalhaes, line 900: This one feels a bit weird to me. I expected this method to be a test method and `test_on_product_publish_with_simple_product` and `test_on_product_publish_with_variable_product ` to be set in a provider.

tests/integration/WC_Facebookcommerce_Integration_Test.php
```diff
@@ -771,8 +773,396 @@ public function test_is_product_sync_enabled_filter() {
...
+
+
+	/** @see \WC_Facebookcommerce_Integration::on_product_publish() */
+	public function test_on_product_publish_with_invalid_product_id() {
+
+		$this->check_on_product_publish_returns_early( 0 );
+	}
+
+
+	private function check_on_product_publish_returns_early( $product_id, \WC_Product $product = null ) {
```
deivamagalhaes, line 931: Same feedback as the other one, it would be a lot clearer if this used a provider.

tests/integration/WC_Facebookcommerce_Integration_Test.php
wvega: @deivamagalhaes I started adding these helper methods because I was unable to access `$this->tester` in the provider. Although I'm not sure why and maybe there is a way to make it available.

tests/integration/WC_Facebookcommerce_Integration_Test.php
wvega: Codeception wouldn't let me because `delete_product_group()` is private. It works with `delete_product_item()` because that one doesn't have the accessibility defined.

Do you know if there is a way to bypass that restriction? If not, I think I can add a note to explain.

tests/integration/WC_Facebookcommerce_Integration_Test.php
wvega: @deivamagalhaes I took another look at this one and found why the tester is not available in the data provider.

The tester is injected in the [`_setUp()` method](https://github.com/Codeception/Codeception/blob/4691ecac31992aba50a7c1c58935a76be29ae8c2/src/Codeception/Test/Unit.php#L50-L53), which is called from PHPUnit's [`TestCase::setup()`](https://github.com/Codeception/phpunit-wrapper/blob/e5edaad5583e74165c59d710a8e49a462b1e4a7a/src/TestCase.php#L12). The problem is that those methods are executed [after the data providers are called](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers):

>All data providers are executed before both the call to the setUpBeforeClass static method and the first call to the setUp method. Because of that you can't access any variables you create there from within a data provider. This is required in order for PHPUnit to be able to compute the total number of tests.

I tried passing just the product properties, but since I use `get_product()` in one case and `get_variable_product()` in the other, I'm forced to add a conditional in `check/test_on_product_publish()` that doesn't seem to help with making the implementation more clear.

What do you think?

deivamagalhaes: I'm fine with that. Thanks for clarifying.

tests/integration/WC_Facebookcommerce_Integration_Test.php
wvega: I added a note about why we are expecting that `get_product_fbid()` is called instead of using `delete_product_group()`: 92a9356.

wvega:
Thank you for reviewing @deivamagalhaes. I'm going to merge this one since tests pass and the two PRs that were merged here were approved already.

PR #1423, opened 2020-06-28
Update on_variable_product_publish() to verify individual variations should be synced

wvega:
# Summary

This PR updates `WC_Facebookcommerce_Integration::on_variable_product_publish()` to also check whether each variation should be synced or not.

## Details

The current implementation accidentally allows variations excluded from Facebook sync or that should be deleted given their stock status to be synced.

The story also mentions that `update_fb_visibility()` doesn't perform any verifications for variations, but I realized we are only calling that method from places where those verifications are being made ([here](https://github.com/facebookincubator/facebook-for-woocommerce/blob/f686e897198ab51df5cafe8cf19923a0a291e361/facebook-commerce.php#L1038) and [here](https://github.com/facebookincubator/facebook-for-woocommerce/blob/f686e897198ab51df5cafe8cf19923a0a291e361/facebook-commerce.php#L3355)), so I no longer believe we need to make changes there.

### Setup

- Check Out of stock visibility in WooCommerce > Settings > Products > Inventory to hide out of stock items from the catalog

### Steps

1. Add a variable product
1. Add a variation with price, in stock, and configured to Sync and show
1. Add a variation with price and out of stock
1. Add a variation with price and configured to Do not sync
    - [x] The variation with price, in stock, and configured to Sync and show is synced to Facebook
    - [x] No other variations are synced

#### Tests

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -1127,9 +1127,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		// scheduled update for each variation that should be synced
 		foreach ( $woo_product->get_children() as $variation_id ) {
 
-			if ( ! $variation = wc_get_product( $variation_id ) ) {
-				continue;
-			}
+			$variation = wc_get_product( $variation_id )
 
 			if ( $variation instanceof \WC_Product && $this->product_should_be_synced( $variation ) && ! $this->delete_on_out_of_stock( $variation_id, $variation ) ) {
 				$variation_ids[] = $variation_id;
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -1122,12 +1122,10 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 			$this->create_product_group( $woo_product, $retailer_id, true );
 		}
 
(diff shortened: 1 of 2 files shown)


facebook-commerce.php
```diff
@@ -1122,13 +1122,21 @@ function on_variable_product_publish( $wp_id, $woo_product = null ) {
...
 
-		// scheduled update for each variation
-		foreach ( $child_products as $item_id ) {
+		// scheduled update for each variation that should be synced
+		foreach ( $woo_product->get_children() as $variation_id ) {
 
-			facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_products( [ $item_id ] );
+			if ( ! $variation = wc_get_product( $variation_id ) ) {
+				continue;
+			}
```
unfulvio, line 1132:
```suggestion
			$variation = wc_get_product( $variation_id )
```

later you are checking if `$variation` is an instance of `\WC_Product` so the condition check with the variable assignment is redundant

wvega:
Thank you for reviewing @unfulvio. I committed your suggestion for the redundant verification.

PR #1277, opened 2020-05-12
Release 2.0.0

wvega:
## Summary

Base branch for v2.0.0

### Stories

- [CH 54003](https://app.clubhouse.io/skyverge/story/54003)

## Before merge

- [x] Test suites pass
- [x] Confirm the changes introduced in this PR work in each supported minor WooCommerce version

## Deploy

- Use `npx sake deploy --deployAssets=0` to skip the `shell:svn_commit_assets` action

PR #1439, opened 2020-07-22
Fix compatibility conflict with YITH Bookings

wvega:
# Summary

This PR fixes a conflict with YITH Booking and Appointment for WooCommerce plugin that triggers a Fatal error when the feed is being generated.

### Setup

1. Install YITH Bookings
1. Create a Booking product (define at least the name and the base cost, and remove the virtual check from the product)

### Steps

Checkout tag `1.11.4`

1. Try and generate a feed file
    - Go to WooCommerce > Settings > Integration > Facebook for WooCommerce
    - Open the browser's console and type `copy(facebookAdsToolboxConfig.feedPrepared.feedUrl)`
    - Paste the feed URL from the clipboard and append `&regenerate=true` to the URL
    - Access the resulting URL
    - [x] `call to undefined function is_wc_booking_product()`

Checkout this branch

1. Try and generate a feed file
    - [x] The feed file is generated correctly
    - [x] There are not Fatal errors or PHP warnings

## Before merge

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@
  * Tweak - Allow to bulk enable sync for virtual products, but automatically set them to Sync and hide
  * Fix - Use the plugin version instead of a timestamp as the version number for enqueued scripts and stylesheets
  * Fix - Use the short description of the parent product for product variations that don't have a description or Facebook description
+ * Fix - Prevent an error when YITH Booking and Appointment for WooCommerce plugin is active
 
 2020.06.04 - version 1.11.4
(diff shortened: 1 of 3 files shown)


wvega:
I just rebased the commits on this branch on top #1277 so that we can include this fix in 2.0.0 instead.

adrianostanley:
@wvega I was able to see the error on 1.11.4 and now it's working for this branch. It looks like I can't check the QA steps - probably I don't have permissions for that yet -  but consider they're ✅ from my point of view.

wvega:
Thank you @adrianostanley for reviewing. LGTM!

PR #1440, opened 2020-07-28
Fire server-side Search event when results include a single product 

wvega:
# Summary

This PR updates the event tracker to send server-side Search event on `template_redirect` before WooCommerce gets a chance to redirect users to a product page.

## Details

If the search results include a single product, WooCommerce automatically [redirects users](https://github.com/woocommerce/woocommerce/blob/4b1a8cd96c0db479d7a227bed75e52e433ec4cae/includes/wc-template-functions.php#L57-L65) to that product's page. As a result the search page results are never rendered and the plugin is unable to inject a snippet to trigger a browser event.

We were previously injecting the browser event and sending the server-side event at the same time, but we don't need to wait until the shop loop starts to send a server-side event.

## UI Changes

N/A

### Setup

- Install and configure Facebook for WooCommere
- Enable debug log
- Install the [Facebook Pixel Helper](https://chrome.google.com/webstore/detail/facebook-pixel-helper/fdgfkebogiimcoedlicjlajpkdmockpc?hl=en) extension

### Steps

1. Search for a product using a search term that returns two or more results
    - [x] The pixel helper registers a Search event for the products in the results
    - [x] A server-side Search event is successfully sent (check the logs)
1. Search for a product using a search term that returns a single result
    - [x] You're redirected to the product page
    - [x] A server-side Search event is successfully sent (check the logs)

### Tests

- [ ] `codecept run integration` pass

## Before merge

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -26,6 +26,10 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		const FB_PRIORITY_HIGH    = 2;
 		const FB_PRIORITY_LOW     = 11;
 
+		/** @var Event search event instance */
+		private $search_event;
+
+
 		public function __construct( $user_info ) {
 			$this->pixel = new WC_Facebookcommerce_Pixel( $user_info );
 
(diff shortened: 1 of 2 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -178,84 +182,116 @@ function( $item ) {
...
 				}
+
+				$event_data = [
+					'event_name'  => 'Search',
+					'custom_data' => [
+						'content_type'  => $content_type,
+						'content_ids'   => json_encode( array_slice( $product_ids, 0, 10 ) ),
+						'contents'      => $contents,
+						'search_string' => get_search_query(),
+						'value'         => \SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Helper::number_format( $total_value ),
```
unfulvio, line 269: can we add a `use` at the beginning of the file like it was done for `use SkyVerge\WooCommerce\Facebook\Events\Event;` so were the helper method can be written as `Framework\SV_WC_Helper::...`

wvega: Yes, fixed in 9db9c5b.

PR #1446, opened 2020-08-03
Make sure timezone identifiers don't have escaped slashes in requests

unfulvio:
# Summary

Ensures that timezone identifiers don't have slashes escaped.

### Steps

- [x] Code review
- [x] Integration tests pass

## Before merge

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -606,7 +606,7 @@ class Connection {
 			'display'       => 'page',
 			'response_type' => 'code',
 			'scope'         => implode( ',', $this->get_scopes() ),
-			'extras'        => json_encode( $this->get_connect_parameters_extras(), JSON_UNESCAPED_SLASHES ),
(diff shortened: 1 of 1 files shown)


unfulvio:
:open_mouth: sorry @wvega I checked again the message and yes I really got it the way around because it looked like an escaping issue to me... I believe we can close this then if the problem is not on our end

wvega:
@unfulvio no worries. I thought the same when were reviewing that conversation yesterday. I asked on the Slack channel to confirm what's the expected result and whether is there anything we can do on the plugin: https://a8c.slack.com/archives/C010ND8MRPU/p1596563456491900?thread_ts=1596452086.478600&cid=C010ND8MRPU

unfulvio:
@wvega should we close this?

wvega:
@unfulvio yes, thank you for checking! We don't need to change anything on our side. Last update confirmed this was mostly a logging bug on Facebook's side.

PR #1447, opened 2020-08-03
Use the site URL as business name fallback if blog name is empty

unfulvio:
# Summary

Ensures that the business name is never empty string.

### Steps

- [ ] Code review
- [ ] Integration tests pass

## Before merge

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -482,18 +482,25 @@ class Connection {
 	 */
 	public function get_business_name() {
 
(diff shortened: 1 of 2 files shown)


unfulvio:
@wvega sounds like a good idea, how about https://github.com/facebookincubator/facebook-for-woocommerce/pull/1447/commits/9470f7f5dbadcb19cf807d1e3f0ddda448168de2 ? this would exclude the dot which comes with `uniquid( '', true )` - so you'd have a string like `skyverge-plugins-development-5f29c5f622045` from `'SkyVerge Plugins Development' === get_bloginfo( 'name' )`

PR #1468, opened 2020-08-18
Allow simple and variable products with zero/empty price to sync

wvega:
# Summary

This PR consolidates the logic to determine whether a product should be synced into the `Products::product_should_be_synced()` and updates the rules to allow simple and variable products (and their variations) that have zero or empty price to be synced.

## Details

The tests from #1422 and #1423 should ensure that existing behavior doesn't change as we update other methods to use the rules now defined in `Products::product_should_be_synced()`.

### Setup

1. Install Composite Products: https://cloud.skyver.ge/mXuANdoD
1. Install and Configure Facebook for WooCommerce 
1. Set Default customer location to Shop base address 
1. Set Prices entered with tax to `No`
1. Set Display prices in the shop to `Excluding tax`
1. Add a 10% tax that matches the shop base address

#### Display prices excluding tax

1. Create a product with empty price
    - [x] The product is synced to Facebook with price $0.00
1. Create a product with $10 as price
    - [x] The product is synced to Facebook with price $10.00
1. Create a product with $10 as price and set Facebook price to 15
    - [x] The product is synced to Facebook with price $15.00
1. Create a variable product with:
    - Variation 1 with empty price
    - Variation 2 with $10 as price
    - Variation 3 with $10 as price and $15 as Facebook price
1. The variations are synced to Facebook
    - [x] Variation 1 is synced to Facebook with price $0.00
    - [x] Variation 2 is synced to Facebook with price $10.00
    - [x] Variation 3 is synced to Facebook with price $15.00
1. Create a Composite Product
    - Do not define a regular price or sale price
    - Add two components configured to be priced individually
1. The Composite Product is synced to Facebook
    - [x] The price of the composite product on Facebook matches the price of the product in the shop
1. Add a Bookable Product with a base cost of $5 and block cost of $2
    - [x] The price of the bookable product on Facebook is $7 

#### Display prices including tax

1. Set Display prices in the shop to `Including tax`
1. Sync products from WooCommerce > Facebook > Product sync
    - [x] The empty price product is synced to Facebook with price $0.00
    - [x] The $10 product is synced to Facebook with price $11.00
    - [x] The $15 product is synced to Facebook with price $15.00 (taxes are not applied to overwritten Facebook prices)
    - [x] Variation 1 is synced to Facebook with price $0.00
    - [x] Variation 2 is synced to Facebook with price $11.00
    - [x] Variation 3 is synced to Facebook with price $15.00 (taxes are not applied to overwritten Facebook prices)
    - [x] The price of the composite product on Facebook includes taxes and matches the price of the product in the shop
    - [x] The price of the bookable product on Facebook is $7.70

## Before merge

- [x] I have confirmed that the changes from https://github.com/facebookincubator/facebook-for-woocommerce/pull/1439 were integrated into this branch

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -1299,10 +1299,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 	function create_product_item( $woo_product, $retailer_id, $product_group_id ) {
 
-		$product_data               = $woo_product->prepare_product( $retailer_id );
-		if ( ! $product_data['price'] ) {
-			return 0;
-		}
+		$product_data = $woo_product->prepare_product( $retailer_id );
 
 		$product_result = $this->check_api_result(
 			$this->fbgraph->create_product_item(
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -896,6 +896,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 				case 'simple':
 				case 'booking':
 				case 'external':
+				case 'composite':
 					$this->on_simple_product_publish( $wp_id );
 				break;
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -572,11 +572,11 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		global $post;
 
 		$woo_product         = new WC_Facebook_Product( $post->ID );
-		$fb_product_group_id = $this->get_product_fbid(
-			self::FB_PRODUCT_GROUP_ID,
-			$post->ID,
-			$woo_product
-		);
+		$fb_product_group_id = null;
+
+		if ( $woo_product->woo_product instanceof \WC_Product && Products::product_should_be_synced( $woo_product->woo_product ) ) {
+			$fb_product_group_id = $this->get_product_fbid( self::FB_PRODUCT_GROUP_ID, $post->ID, $woo_product );
+		}
 
 		?>
 			<span id="fb_metadata">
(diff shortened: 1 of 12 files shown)


wvega:
@ChaseWiseman thank you for reviewing and for chatting those errors! I was able to reproduce all of them 🙏

>The `visibility` property seems to be acting strange

I broke thise one on an previous PR ([commit](https://github.com/facebookincubator/facebook-for-woocommerce/commit/d7d2c85ef27b9a8dc9a00538186608ea4057384a)). I think a proper fix is to rely on `Products::is_product_visible()` to set the value of that property, instead of using `::is_hidden()`.

I added integration tests (dfa6a9c2623dc17f5176ec6c3414e5cc07a7d4a4) and a fix (c57b6dc). Then removed the last usage of `is_hidden()` (0b360b3) and deprecated the method (8171747).

>Create a product with $10 as price and Create a product with $10 as price and set Facebook price to 15

These two were caused by the price cache I added to `Products::get_product_price()`. I removed the cache for now as I can't guarantee that the correct price is set the first time the method is called for a product (b0b0fed).

In this case, the `fb_change_product_published_status()` method was triggering a call to `get_product_price()` before WooCommerce had updated the product properties, so the initial zero value was stored in the cache. This commit fixes the current case by changing the order of the verifications (345e47b), but I don't think that prevents cache from causing issues in other scenarios.

>Create a Composite Product

I'm sorry I didn't test this properly. There was no way for those steps to work, because the plugin is currently ignoring `composite` products.

I added `composite` to the list of types that we sync to start handling them as simple products: 0af04d1. Previous versions of the plugin would handle non-variable or non-variable-subscription product as simple products: 

https://github.com/facebookincubator/facebook-for-woocommerce/blob/bbadcb536c8bcdaa4ae6ebd881d861f5ebd1c282/facebook-commerce.php#L950-L954

PR #1471, opened 2020-08-21
Move expensive sync queries into the background job

ChaseWiseman:
# Summary

Moves the checks for determining a product's sync availability to the background processor.

## Details

When doing a full catalog sync, the plugin first gathers all potential product IDs and filters out any that _shouldn't_ sync. After that, it creates the background job to process all of those IDs and send them to Facebook via the batch API.

If the merchant has a massive catalog, then that first step of checking the products can run out of memory because we need to instantiate the product, check meta, potentially its parent, and its terms for term exclusions.

Since we already have a background process that loops through the products and instantiates them before sending to Facebook, we can move that expensive stuff to the background so it benefits from the out of memory recovery that provides.

## UI Changes

_N/A_

## QA

- Connect the plugin
- Have a few "Sync and show" products
- Have a few "Do not sync" products

1. Trigger a catalog sync via the "Sync products" button
    - [x] The products that should sync are synced
    - [x] The products that should not sync are not synced

- [x] Integration tests pass

## Before merge

### DIFF
--- a/includes/Products/Sync.php
+++ b/includes/Products/Sync.php
@@ -103,27 +103,8 @@ class Sync {
 		// remove parent products because those can't be represented as Product Items
 		$product_ids = array_diff( $product_ids, $parent_product_ids );
 
-		// make sure the product should be synced and add it to the sync queue
-		foreach ( $product_ids as $product_id ) {
-
-			$woo_product = new \WC_Facebook_Product( $product_id );
-
-			// skip if we don't have a valid product object
-			if ( ! $woo_product->woo_product instanceof \WC_Product ) {
-				continue;
-			}
-
-			if ( Products::product_should_be_deleted( $woo_product->woo_product ) ) {
-				continue;
-			}
-
-			// skip if not enabled for sync
-			if ( ! Products::product_should_be_synced( $woo_product->woo_product ) ) {
-				continue;
-			}
-
-			$this->create_or_update_products( [ $product_id ] );
-		}
+		// queue up these IDs for sync. they will only be included in the final requests if they should be synced
+		$this->create_or_update_products( $product_ids );
 	}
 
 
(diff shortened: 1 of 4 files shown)


simonporter007:
@wvega how many total products to be synced was that in your largest catalog that still failed?

wvega:
@simonporter007 the `get_posts()` call returns `44466` IDs to check. The loop doesn't end so I can't tell how many end up in the `$product_ids` variable, but I have `43454` product variations on the database, so it's at least 43k product items. 

Those are all created with Smooth Generator and there are no excluded terms configured, so I think they all sync.

simonporter007:
@wvega awesome thanks! I haven't seen many if any off the top of my head, that were as large as that. I was just trying to gauge what our limits were, as if we still hit issues at around 20k then it might not make us any better off even with the changes added here. Doesn't seem like that's the case though, so that's awesome, thanks!

ChaseWiseman:
@wvega thanks for testing with that big catalog! Yeah I think the changes here are a good improvement as well. I definitely didn't test with anything that large. I _think_ mine was ~10k. You mentioned trying other catalogs - what was the largest that worked for you?

wvega:
The other catalogs I had tested were much much smaller with around 1K, so those don't tells us a lot.

Last week I created two other large catalogs and got better results. The largest that worked for me has `21491` product records in the database, `512` products, and `20979` product variations.

|  Products | Variations | Records | Result |
|:-:	|:-:	|:-:	|:-:	|
| 262 	| 10137 |10339 | No errors |
| 512 	|  20979 | 21491 | No errors |
| 1012	| 43454 | 44466 | 512 MB Memory exhausted |

I think we should merge this one. 21K products is a big improvement and should give a better experience for a good portion of the users.

wvega (approved):

wvega:
@simonporter007 please note that these improvements cover the background sync only. The feed generator still tries to determine whether a product is eligible and also prepare its data for all products in a single request, so there is no change there.

PR #1480, opened 2020-08-21
Implement the Product_Categories methods

deivamagalhaes:
# Summary

This PR implements the `Product_Categories::get_google_product_category_id()` and `Product_Categories::update_google_product_category_id()` methods.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Product_Categories.php
+++ b/includes/Product_Categories.php
@@ -46,7 +46,7 @@ class Product_Categories {
 	 */
 	public static function update_google_product_category_id( $id, $category_id ) {
 
-		add_term_meta( $id, Products::GOOGLE_PRODUCT_CATEGORY_META_KEY, $category_id, true );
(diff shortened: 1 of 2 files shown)


includes/Product_Categories.php
```diff
@@ -47,7 +46,7 @@ public static function get_google_product_category_id( $category_id ) {
 	 */
 	public static function update_google_product_category_id( $id, $category_id ) {
 
-		// TODO: implement
+		add_term_meta( $id, Products::GOOGLE_PRODUCT_CATEGORY_META_KEY, $category_id, true );
```
ChaseWiseman, line 49: we need to use `update_term_meta()`, no? With `$unique = true` `add_term_meta()` bails if the meta key already exists so subsequent updates won't persist.

PR #1481, opened 2020-08-22
Implement Commerce's availability methods

wvega:
# Summary

This PR implements the `is_available()` and `is_connected()` methods in the `Commerce` handler.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Commerce.php
+++ b/includes/Commerce.php
@@ -78,7 +78,7 @@ class Commerce {
 		list( $country ) = explode( ':', get_option( 'woocommerce_default_country' ) );
 
 		/**
-		 * Filters whether Commerce features should be available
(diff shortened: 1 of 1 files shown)


ChaseWiseman:
@deivamagalhaes @wvega yeah we don't yet have any sort of decision tree on when to add return filters. The ones here seem fine.

Generally I think we've gone with asking ourselves "is this something developers may want to customize? How would they use it?" types of questions when deciding when to add a hook and leaving it up to the story owner's discretion. Or if you as a reviewer see an opportunity for a filter, you can call it out there as well! imo it's probably not something we want to detail in every doc for every method description unless there are specific requirements where something _must_ be filtered, or _must not_ have a filter.

PR #1482, opened 2020-08-22
Implement the Products methods used to check/set if Commerce is enabled and if the product is ready

deivamagalhaes:
# Summary

This PR implements the `Products::is_product_ready_for_commerce()`, `Products::is_commerce_enabled_for_product()` and  `Products::update_commerce_enabled_for_product()` methods.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -480,7 +480,7 @@ class Products {
 	 */
 	public static function update_commerce_enabled_for_product( \WC_Product $product, $is_enabled ) {
 
-		// TODO: implement
+		$product->update_meta_data( Products::COMMERCE_ENABLED_META_KEY, wc_bool_to_string( $is_enabled ) );
 	}
 
 
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -449,7 +449,7 @@ class Products {
 	 */
 	public static function is_product_ready_for_commerce( \WC_Product $product ) {
 
-		return $product->get_manage_stock()
+		return $product->managing_stock()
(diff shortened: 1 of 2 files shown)


includes/Products.php
```diff
@@ -479,7 +480,7 @@ public static function is_commerce_enabled_for_product( \WC_Product $product )
 	 */
 	public static function update_commerce_enabled_for_product( \WC_Product $product, $is_enabled ) {
 
-		// TODO: implement
+		$product->update_meta_data( Products::COMMERCE_ENABLED_META_KEY, wc_bool_to_string( $is_enabled ) );
```
ChaseWiseman, line 483: How about we also `save_meta_data()` here like the other update methods, as right now it won't be persisted

tests/integration/Products_Test.php
```diff
@@ -321,6 +322,116 @@ public function test_get_product_price_filter() {
...
+	 *
+	 * @dataProvider provider_update_commerce_enabled_for_product
+	 */
+	public function test_update_commerce_enabled_for_product( $param_value, $expected_meta_value ) {
+
+		$product = $this->get_product();
+
+		Products::update_commerce_enabled_for_product( $product, $param_value );
+
+		$this->assertEquals( $expected_meta_value, $product->get_meta( Products::COMMERCE_ENABLED_META_KEY ) );
```
ChaseWiseman, line 418: Mind updating this test to get a fresh instance of the product before asserting? Right now it doesn't test whether the meta is persisted.

PR #1483, opened 2020-08-22
Implement the Products::update_google_product_category_id() method

deivamagalhaes:
# Summary

This PR implements the `Products::update_google_product_category_id()` method.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -510,7 +510,7 @@ class Products {
 	 */
 	public static function update_google_product_category_id( \WC_Product $product, $category_id ) {
 
-		// TODO: implement
+		$product->update_meta_data( Products::GOOGLE_PRODUCT_CATEGORY_META_KEY, $category_id );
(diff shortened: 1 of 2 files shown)


includes/Products.php
```diff
@@ -510,7 +510,7 @@ public static function get_google_product_category_id( \WC_Product $product ) {
 	 */
 	public static function update_google_product_category_id( \WC_Product $product, $category_id ) {
 
-		// TODO: implement
+		$product->update_meta_data( Products::GOOGLE_PRODUCT_CATEGORY_META_KEY, $category_id );
```
ChaseWiseman, line 513: Similar to https://github.com/facebookincubator/facebook-for-woocommerce/pull/1482#pullrequestreview-473784482 we'll need to save the meta and the test needs to be updated to test for the persisted data

PR #1484, opened 2020-08-22
Implement the Connection page access token methods

wvega:
# Summary

This PR implements the `get_page_access_token()` and `update_page_access_token()` methods to the `Connection` handler and deprecates the methods with the same names in the integration class.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -173,6 +173,24 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 
 
 		/**
+		 * Gets deprecated and removed hooks.
+		 *
+		 * @since 2.1.0-dev.1
+		 *
+		 * @return array
+		 */
+		protected function get_deprecated_hooks() {
+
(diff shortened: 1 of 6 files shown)


ChaseWiseman:
>I'm not so sure I'm convinced we gain much by moving these methods.

@deivamagalhaes the only reason is to move that method into a more appropriate place, in the connection handler. Is there a change you wanted to suggest here, or can this be merged?

deivamagalhaes:
> > I'm not so sure I'm convinced we gain much by moving these methods.
> 
> @deivamagalhaes the only reason is to move that method into a more appropriate place, in the connection handler. Is there a change you wanted to suggest here, or can this be merged?

But is it worth it, having to deprecate hooks and all?

ChaseWiseman:
@deivamagalhaes yep I think so - were this a patch release I wouldn't do it but since this is going in a feature release and the page access token is being re-implemented as part of those new features I think it makes sense

PR #1485, opened 2020-08-22
Implement the Products::get_google_product_category_id() method

deivamagalhaes:
# Summary

This PR implements the `Products::get_google_product_category_id()` method.

## Details

Attempts to get the Google product category ID from the product or parent product (if the product is a variation) metadata. If not set, gets it from the highest level (the highest leaf level) product category. If not set or if there are multiple categories on the same level with conflicting Google product category IDs, falls back to the plugin-level option.

## QA

- [x] Integrations test pass 

**Note**: some tests are failing because some required code was not merged yet:
- `test_get_google_product_category_id_product_single_category` and `test_get_google_product_category_id_product_multiple_categories`: require code from #1480
- `test_get_google_product_category_id_default`: requires code from #1479 

We can wait for these to be merged to QA this one.

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -495,8 +495,107 @@ class Products {
 	 */
 	public static function get_google_product_category_id( \WC_Product $product ) {
 
-		// TODO: implement
-		return '';
+		// attempt to get from product or parent product metadata
+		if ( $product->is_type( 'variation' ) ) {
+			$google_product_category_id = get_post_meta( $product->get_parent_id(), self::GOOGLE_PRODUCT_CATEGORY_META_KEY, true );
+		} else {
+			$google_product_category_id = $product->get_meta( self::GOOGLE_PRODUCT_CATEGORY_META_KEY );
+		}
+
+		// fallback to the highest category's Google product category ID
+		if ( empty( $google_product_category_id ) ) {
+
+			$google_product_category_id = self::get_google_product_category_id_from_highest_category( $product );
+		}
+
+		// fallback to plugin-level default Google product category ID
+		if ( empty( $google_product_category_id ) ) {
+
+			$google_product_category_id = facebook_for_woocommerce()->get_commerce_handler()->get_default_google_product_category_id();
+		}
+
+		return $google_product_category_id;
+	}
+
+
+	/**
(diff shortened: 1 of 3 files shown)


includes/Products.php
```diff
@@ -497,8 +497,107 @@ public static function update_commerce_enabled_for_product( \WC_Product $product
 	 */
 	public static function get_google_product_category_id( \WC_Product $product ) {
 
-		// TODO: implement
-		return '';
+		// attempt to get from product or parent product metadata
+		if ( $product->is_type( 'variation' ) ) {
+			$google_product_category_id = get_post_meta( $product->get_parent_id(), self::GOOGLE_PRODUCT_CATEGORY_META_KEY, true );
```
ChaseWiseman, line 502: we need to implement the same update here to use `\WC_Product::get_meta()` instead, after checking for an instance

includes/Products.php
```diff
@@ -497,8 +497,107 @@ public static function update_commerce_enabled_for_product( \WC_Product $product
...
+	 *
+	 * @param \WC_Product $product the product object
+	 * @return string
+	 */
+	private static function get_google_product_category_id_from_highest_category( \WC_Product $product ) {
+
+		$google_product_category_id = '';
+
+		// get all categories for the product
+		$categories = get_the_terms( $product->get_id(), 'product_cat' );
```
ChaseWiseman, line 536: this won't get the terms if it's a variable product will it? I think you'll need to either get the parent in here or before calling this method.

includes/Products.php
```diff
@@ -497,8 +497,107 @@ public static function update_commerce_enabled_for_product( \WC_Product $product
...
+			foreach ( $categories as $key => $category ) {
+
+				$google_product_category_id = Product_Categories::get_google_product_category_id( $category->term_id );
+				if ( empty( $google_product_category_id ) ) {
+					unset( $categories_per_level[ $level ][ $key ] );
+				}
+			}
+
+			if ( empty( $categories_per_level[ $level ] ) ) {
+				unset ( $categories_per_level[ $level ] );
```
ChaseWiseman, line 573:
```suggestion
				unset( $categories_per_level[ $level ] );
```

PR #1486, opened 2020-08-22
Implement the gender methods

deivamagalhaes:
# Summary

This PR implements the `Products::get_product_gender()` and `Products:update_product_gender()` methods.

## Details

Validation is done on the getter. The setter allows you to save whatever value you want.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -639,8 +639,17 @@ class Products {
 	 */
 	public static function get_product_gender( \WC_Product $product ) {
 
-		// TODO: implement
-		return '';
+		if ( $product->is_type( 'variation' ) ) {
+			$gender = get_post_meta( $product->get_parent_id(), self::GENDER_META_KEY, true );
+		} else {
+			$gender = $product->get_meta( self::GENDER_META_KEY );
+		}
+
+		if ( ! in_array( $gender, [ 'female', 'male', 'unisex' ] ) ) {
+			$gender = 'unisex';
+		}
+
+		return $gender;
 	}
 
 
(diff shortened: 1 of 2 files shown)


includes/Products.php
```diff
@@ -528,8 +528,17 @@ public static function update_google_product_category_id( \WC_Product $product,
 	 */
 	public static function get_product_gender( \WC_Product $product ) {
 
-		// TODO: implement
-		return '';
+		if ( $product->is_type( 'variation' ) ) {
+			$gender = get_post_meta( $product->get_parent_id(), self::GENDER_META_KEY, true );
```
ChaseWiseman, line 532: Need to get the product object and use its `get_meta()` method here, as direct post meta functions are essentially deprecated for products and orders.

includes/Products.php
```diff
@@ -543,7 +552,7 @@ public static function get_product_gender( \WC_Product $product ) {
 	 */
 	public static function update_product_gender( \WC_Product $product, $gender ) {
 
-		// TODO: implement
+		$product->update_meta_data( Products::GENDER_META_KEY, $gender );
```
ChaseWiseman, line 555: Same here as https://github.com/facebookincubator/facebook-for-woocommerce/pull/1482#pullrequestreview-473784482 - needs to save the meta 👍

includes/Products.php
```diff
@@ -530,8 +530,18 @@ public static function update_google_product_category_id( \WC_Product $product,
 	 */
 	public static function get_product_gender( \WC_Product $product ) {
 
-		// TODO: implement
-		return '';
+		if ( $product->is_type( 'variation' ) ) {
+			$parent_product = wc_get_product( $product->get_parent_id() );
+			$gender         = $parent_product->get_meta( self::GENDER_META_KEY );
```
ChaseWiseman, line 535: `$parent_product` might not be an instance of `\WC_product` so we need to avoid that fatal in case an orphaned variation is passed in

PR #1487, opened 2020-08-24
Implement the Products methods to get available/distinct product attributes

deivamagalhaes:
# Summary

This PR implements the `Products::get_available_product_attributes()` and `Products::get_distinct_product_attributes()` methods.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -709,8 +709,11 @@ class Products {
 	 */
 	public static function get_distinct_product_attributes( \WC_Product $product ) {
 
-		// TODO: implement
-		return [];
+		return [
+			'color'   => self::get_product_color_attribute( $product ),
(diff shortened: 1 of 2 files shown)


includes/Products.php
```diff
@@ -710,8 +709,11 @@ public static function get_available_product_attributes( \WC_Product $product )
 	 */
 	public static function get_distinct_product_attributes( \WC_Product $product ) {
 
-		// TODO: implement
-		return [];
+		return [
+			self::get_product_color_attribute( $product ),
```
ChaseWiseman, line 713: we should probably filter this array or something, right? If one of these isn't set, it'll be an empty string in the array if I'm reading this correctly.

PR #1491, opened 2020-08-26
Add commerce data to the prepared product data

ChaseWiseman:
# Summary

Adds gender, inventory, and google product category values for commerce-enabled products

## Details

Implements `gender` here instead of `prepare_variants_for_item()` because it's a base product value and doesn't have the potential to change per-variation.

Ensures the `inventory` value is never negative - I've confirmed that the API rejects requests with negative inventory.

## UI Changes

_N/A_

## QA

- [x] `integration WC_Facebook_Product_Test` tests pass

## Before merge

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -534,6 +534,11 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 				'visibility'            => Products::is_product_visible( $this->woo_product ) ? \WC_Facebookcommerce_Integration::FB_SHOP_PRODUCT_VISIBLE : \WC_Facebookcommerce_Integration::FB_SHOP_PRODUCT_HIDDEN,
 			);
 
+			// add the Commerce values
+			if ( Products::is_product_ready_for_commerce( $this->woo_product ) ) {
+				$product_data['gender']    = Products::get_product_gender( $this->woo_product );
+			}
+
 			// Only use checkout URLs if they exist.
 			if ( $checkout_url ) {
 				  $product_data['checkout_url'] = $checkout_url;
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -536,8 +536,13 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 
 			// add the Commerce values
 			if ( Products::is_product_ready_for_commerce( $this->woo_product ) ) {
+
 				$product_data['gender']    = Products::get_product_gender( $this->woo_product );
 				$product_data['inventory'] = (string) $this->woo_product->get_stock_quantity();
+
+				if ( $google_product_category = Products::get_google_product_category_id( $this->woo_product ) ) {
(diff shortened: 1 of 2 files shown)


tests/integration/WC_Facebook_Product_Test.php
```diff
@@ -31,6 +31,122 @@ protected function _after() {
 	/** Test methods **************************************************************************************************/
 
 
+	/** @see \WC_Facebook_Product::prepare_product() */
+	public function test_prepare_product_not_ready_for_commerce_gender() {
```
deivamagalhaes, line 35: Can't these 3 tests be one?

includes/fbproduct.php
```diff
@@ -534,6 +534,17 @@ public function prepare_product( $retailer_id = null, $prepare_for_product_feed
 				'visibility'            => Products::is_product_visible( $this->woo_product ) ? \WC_Facebookcommerce_Integration::FB_SHOP_PRODUCT_VISIBLE : \WC_Facebookcommerce_Integration::FB_SHOP_PRODUCT_HIDDEN,
 			);
 
+			// add the Commerce values
+			if ( Products::is_product_ready_for_commerce( $this->woo_product ) ) {
```
deivamagalhaes, line 538: From the discovery doc, I thought we were going to send these values regardless of Commerce being enabled (except for the Google product category ID). Have we decided to change that?

tests/integration/WC_Facebook_Product_Test.php
ChaseWiseman: They could, but we should avoid multiple asserts in a single integration test method where possible

includes/fbproduct.php
ChaseWiseman: While adding this I realized that it may not be a good idea to do that, after going over the frontend implementation. If the product is not ready or disabled for Commerce, the settings field for gender can't be changed. Folks could end up with products that send along values that they no longer want to use for gender, and inventory isn't relevant outside of the Commerce context.

deivamagalhaes: Sounds good, we should probably update the doc to reflect that decision. 😬

tests/integration/WC_Facebook_Product_Test.php
ChaseWiseman: Whoops, looks like I forgot to link: https://github.com/skyverge/lumiere/wiki/Best-practices#single-asserts--logic

That’s from back when we first started defining this stuff though, ages ago, so we can certainly amend. Thoughts?

deivamagalhaes: I think it's a tradeoff. Sometimes I believe it's worth having multiple asserts, if it would make the tests a bit faster and would not add a lot of complexity. 🤷‍♀️

includes/fbproduct.php
ChaseWiseman: Updated!

PR #1496, opened 2020-08-26
Add the configured attribute values to prepared product data

ChaseWiseman:
# Summary

Adds the color, size, and pattern attribute values to the product data.

## Details

Added to both the API updates and feed.

Note: I had to update the `Products::is_commerce_enabled_for_product()` method to handle variations, so the parent is checked.

## UI Changes

_N/A_

## QA

- [x] Integration tests pass

## Before merge

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -622,6 +622,11 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 			// For each product field type, pull the single variant
 			foreach ( $variant_names as $original_variant_name ) {
 
+				// don't handle any attributes that are designated as Commerce attributes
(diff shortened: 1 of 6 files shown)


includes/fbproduct.php
```diff
@@ -617,6 +622,11 @@ public function prepare_variants_for_item( &$product_data ) {
 			// For each product field type, pull the single variant
 			foreach ( $variant_names as $original_variant_name ) {
 
+				// don't handle any attributes that are designated as Commerce attributes
+				if ( in_array( str_replace( array( 'attribute_' ), '', strtolower( $original_variant_name ) ), Products::get_distinct_product_attributes( $this->woo_product ), true ) ) {
```
deivamagalhaes, line 626: Did you intend to have multiple needles on the `str_replace`?

ChaseWiseman: Nah that's just left over from an earlier approach - removed in 0c4713234b32c857ec05d77f767b05ad285eae67

PR #1501, opened 2020-08-27
Use attribute slugs for comparison & return values

ChaseWiseman:
# Summary

Ensures the various attribute methods can be used consistently with both taxonomy & non-taxonomy attributes, regardless of their naming.

## Details

Name was targeted as the value to use initially, but it turns out only the name is sanitized for taxonomy attributes. Custom attributes retain their nice name, e.g. `Custom attribute`, and can cause hidden issues when building product data.

WooCommerce stores variation attribute data using only the slug, and the existing Facebook data generation relies on the slugs as well, so we should probably stick to that.

## UI Changes

_N/A_

## QA

- [x] Integration tests pass

## Before merge

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -703,10 +703,10 @@ class Products {
 
 			if ( empty( $attribute_name ) ) {
 				// try to find a matching attribute
-				foreach ( self::get_available_product_attributes( $product ) as $attribute ) {
+				foreach ( self::get_available_product_attributes( $product ) as $slug => $attribute ) {
 
 					if ( stripos( $attribute->get_name(), 'color' ) !== false || stripos( $attribute->get_name(), 'colour' ) !== false ) {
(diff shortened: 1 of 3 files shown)


includes/Products.php
```diff
@@ -989,9 +989,10 @@ public static function product_has_attribute( \WC_Product $product, $attribute_n
 
 		$found = false;
 
-		foreach ( self::get_available_product_attributes( $product ) as $attribute ) {
+		foreach ( self::get_available_product_attributes( $product ) as $slug => $attribute ) {
 
-			if ( $attribute_name === $attribute->get_name() ) {
+			// taxonomy attributes have a slugged name, but custom attributes do not so we check the attribute key first
```
deivamagalhaes, line 994:
```suggestion
			// taxonomy attributes have a slugged name, but custom attributes do not so we check the attribute key
```

PR #1513, opened 2020-08-27
Mark out-of-stock products has hidden in the feed file

ChaseWiseman:
# Summary

Instead of omitting these products from the feed file, this PR changes their `visibility` value to `staging` when the Hide out of stock setting is enabled.

## Details

This prevents an edge case specifically for our users who upgraded to v1.11 (FBE 1.5) and then to v2. In their case, the feed was switched to a full replace schedule, so omitting the products resulted in deleting the products, which preserved the behavior from v1.10.

However, the feed replacement does not delete omitted products from _other sources_, in this case the other source being the new v2 API. So any product that was updated via API again in v2 would not get omitted and thus not get the "out of stock" update when their feed schedule triggers.

## UI Changes

_N/A_

## QA

- Connect the plugin
- Create 2 products, set to sync with Facebook

1. Mark one of the products as Out of stock
1. Manually trigger the feed file generation via action scheduler
    - [x] The feed contains both products
    - [x] The out of stock product's `visibility` is `staging`
    - [x] The in-stock product's `visibility` is `published`

## Before merge

### DIFF
--- a/includes/fbproductfeed.php
+++ b/includes/fbproductfeed.php
@@ -571,10 +571,6 @@ if ( ! class_exists( 'WC_Facebook_Product_Feed' ) ) :
 						continue;
 					}
 
-					if ( Products::product_should_be_deleted( $woo_product->woo_product ) ) {
-						continue;
-					}
-
 					// skip if not enabled for sync
 					if ( ! Products::product_should_be_synced( $woo_product->woo_product ) ) {
 						continue;
@@ -737,6 +733,11 @@ if ( ! class_exists( 'WC_Facebook_Product_Feed' ) ) :
 				$this->no_default_product_count++;
 
 				$product_data['default_product'] = '';
+			}
+
+			// when dealing with the feed file, only set out-of-stock products as hidden
+			if ( Products::product_should_be_deleted( $woo_product->woo_product ) ) {
+				$product_data['visibility'] = \WC_Facebookcommerce_Integration::FB_SHOP_PRODUCT_HIDDEN;
 			}
 
 			return $product_data['retailer_id'] . ',' .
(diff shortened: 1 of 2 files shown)


ChaseWiseman:
@deivamagalhaes oh you know what, looks like my QA steps are missing a step - can you try and enable "Hide out of stock items from the catalog" from WooCommerce -> Products -> Inventory first?

>Also, if the API sync will mark the product as Sold on Facebook https://cloud.skyver.ge/12uJLG2m, should the feed hide it?

I'm not really sure what that is - is that what you see when a product is set to `out of stock` from the API request?

deivamagalhaes:
> @deivamagalhaes oh you know what, looks like my QA steps are missing a step - can you try and enable "Hide out of stock items from the catalog" from WooCommerce -> Products -> Inventory first?

Sure, I'll check it again in a bit. Should the code check for this option as well?

> > Also, if the API sync will mark the product as Sold on Facebook https://cloud.skyver.ge/12uJLG2m, should the feed hide it?
> 
> I'm not really sure what that is - is that what you see when a product is set to `out of stock` from the API request?

Yes, it is what I see when I set the product to "Out of Stock" and save it (I assume triggering an API update).

deivamagalhaes:
> Should the code check for this option as well?

Dumb question, just noticed this is exactly what the `product_should_be_deleted()` method checks.

deivamagalhaes:
It may not be a bad idea to rename the method to `product_should_be_hidden()` though.

deivamagalhaes:
Well, in other places the product is actually deleted. So maybe it is a bad idea to rename the method. 🤔

ChaseWiseman:
@deivamagalhaes yeah the most up-to-date behavior, in v2+, is to delete the product (via API request) when out of stock and that setting is enabled, and we're not changing that here. The feed file is a deprecated sync method, and this PR is specifically for users of v1.11.x who updated to v2. We can open a different story for any suggested changes in behavior beyond that.

PR #1514, opened 2020-08-27
Implement the Commerce\Orders::find_local_order() method

deivamagalhaes:
# Summary

This PR implements the `Commerce\Orders::find_local_order()` method.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Commerce/Orders.php
+++ b/includes/Commerce/Orders.php
@@ -40,8 +40,21 @@ class Orders {
 	 */
 	public function find_local_order( $remote_id ) {
 
-		// TODO: implement
-		return null;
+		$order_id_array = get_posts( [
+			'post_type'   => 'shop_order',
+			'nopaging'    => true,
+			'numberposts' => 1,
+			'fields'      => 'ids',
+			'post_status' => 'any',
+			'meta_key'    => self::REMOTE_ID_META_KEY,
+			'meta_value'  => $remote_id,
+		] );
+
+		if ( ! empty( $order_id_array ) ) {
+			$order_id = current( $order_id_array );
+		}
+
+		return ! empty( $order_id ) ? wc_get_order( $order_id ) : null;
 	}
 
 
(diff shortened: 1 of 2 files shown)


includes/Commerce/Orders.php
```diff
@@ -40,8 +40,21 @@ class Orders {
 	 */
 	public function find_local_order( $remote_id ) {
 
-		// TODO: implement
-		return null;
+		$order_id_array = get_posts( [
```
ChaseWiseman, line 43: How about we use [`wc_get_orders()`](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query)? That way we'll have less to update once WC switches away from custom post types. It seems to still support the same meta query args.

includes/Commerce/Orders.php
```diff
@@ -69,10 +82,84 @@ public function create_local_order( Order $remote_order ) {
 	 * @param Order $remote_order Orders API order object
 	 * @param \WC_Order $local_order local order object
 	 * @return \WC_Order
+	 * @throws SV_WC_Plugin_Exception|\WC_Data_Exception
 	 */
 	public function update_local_order( Order $remote_order, \WC_Order $local_order ) {
 
-		// TODO: implement
+		// add/update items
+		foreach ( $remote_order->get_items() as $item ) {
```
ChaseWiseman, line 90: It looks like this PR and #1515 contain the same changes, both to `find_local_order()` and `update_local_order()`. Can we get the two isolated to their respective methods?

deivamagalhaes: Weird. I force-pushed this one to the correct commit.

includes/Commerce/Orders.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1514/commits/b6e5820a5b270a5cea1b187a24eca56984c80bea

PR #1515, opened 2020-08-28
Implement the Commerce\Orders::update_local_order() method

deivamagalhaes:
# Summary

This PR implements the `Commerce\Orders::update_local_order()` method.

## Details

So many questions! I've added several TODOs to the code to the points I am not sure how to map the remote/local data. 

## QA

- [x] Integrations test pass

### Setup

- Have the plugin configured and connected
- Have at least one product with price and inventory set, sync enabled, Commerce enabled (set the post `_wc_facebook_commerce_enabled` meta data to `yes`), and a Google product category set (set the post `_wc_facebook_google_product_category`)

### Steps

1. Place a Commerce order purchasing that product
1. Wait for the scheduled `wc_facebook_commerce_fetch_orders` action to run (or run it via the Scheduled Actions page)
    - [x] A new order is created in WC
1. Open the new order
    - [x] The order information (item, item tax, shipping item, shipping taxes, totals, billing information, and shipping address) matches what is returned by the API

## Before merge


includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
 	 * @return \WC_Order
+	 * @throws SV_WC_Plugin_Exception|\WC_Data_Exception
 	 */
 	public function update_local_order( Order $remote_order, \WC_Order $local_order ) {
 
-		// TODO: implement
+		// add/update items
+		foreach ( $remote_order->get_items() as $item ) {
+
+			$wc_product_id = $item['retailer_id'];
```
ChaseWiseman, line 92: The FB retailer ID isn't 1:1 with the Woo product ID. See `\WC_Facebookcommerce_Utils::get_fb_retailer_id()`.

Without a SKU, the retailer ID will be `wc_post_id_{product-id}`
With a SKU, it'll be `{SKU}_{product-id}`

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+				}
+
+				$matching_wc_order_item = new \WC_Order_Item_Product();
+				$local_order->add_item( $matching_wc_order_item );
+			}
+
+			$matching_wc_order_item->set_product_id( $wc_product_id );
+			$matching_wc_order_item->set_quantity( $item['quantity'] );
+			$matching_wc_order_item->set_subtotal( $item['quantity'] * $item['price_per_unit']['amount'] );
+			// TODO: should we use the estimated_tax or the captured_tax on the line below?
```
ChaseWiseman, line 120: I think we'll be using `estimated_tax` - captured tax represents the tax after the order/item has been shipped, but we don't fulfill orders at the line-item level, only on a full order status change of Completed

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+			$matching_wc_order_item->set_subtotal_tax( $item['tax_details']['estimated_tax']['amount'] );
+			$matching_wc_order_item->save();
+		}
+
+		// update information from selected_shipping_option
+		$selected_shipping_option = $remote_order->get_selected_shipping_option();
```
ChaseWiseman, line 126: we should probably also create a shipping line item for the shipping. It looks like they pass the `name` we can use for that in addition to the amount values

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		// update information from shipping_address
+		$shipping_address = $remote_order->get_shipping_address();
+
+		$local_order->set_shipping_last_name( $shipping_address['name'] );
+		$local_order->set_shipping_address_1( $shipping_address['street1'] );
+		$local_order->set_shipping_address_2( $shipping_address['street2'] );
+		$local_order->set_shipping_city( $shipping_address['city'] );
+		$local_order->set_shipping_state( $shipping_address['state'] );
+		$local_order->set_shipping_postcode( $shipping_address['postal_code'] );
+		$local_order->set_shipping_country( $shipping_address['country'] );
```
ChaseWiseman, line 140: We'll need to confirm whether the country value is an alpha-2 or alpha-3 code, as their docs don't list which. If alpha-3, we'll need to convert to Woo's format using the framework's `Country_Helper::convert_alpha_country_code()`

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		// update information from selected_shipping_option
+		$selected_shipping_option = $remote_order->get_selected_shipping_option();
+
+		$local_order->set_shipping_total( $selected_shipping_option['price']['amount'] );
+		$local_order->set_shipping_tax( $selected_shipping_option['calculated_tax']['amount'] );
+
+		// update information from shipping_address
+		$shipping_address = $remote_order->get_shipping_address();
+
+		$local_order->set_shipping_last_name( $shipping_address['name'] );
```
ChaseWiseman, line 134: Why just the order's last name here?

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+					break;
+				}
+			}
+
+			if ( empty( $matching_wc_order_item ) ) {
+
+				$wc_product = wc_get_product( $wc_product_id );
+
+				if ( ! $wc_product instanceof \WC_Product ) {
+					throw new SV_WC_Plugin_Exception( "Product with WC ID $wc_product_id not found" );
```
ChaseWiseman, line 110: I'm not sure we should throw an exception that stops the entire order update here. It's not really detailed in documentation, but we could have a situation where an order has one item that does belong to this store and another item that does not. Each catalog can have multiple data feed sources outside of WooCommerce, so it's a possibility.

How about we add an order note that a product with the FB item's retailer ID wasn't found?

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		$local_order->set_shipping_postcode( $shipping_address['postal_code'] );
+		$local_order->set_shipping_country( $shipping_address['country'] );
+
+		// update information from estimated_payment_details
+		$estimated_payment_details = $remote_order->get_estimated_payment_details();
+
+		// TODO: confirm if we should use $estimated_payment_details['subtotal']['items'] for something
+		// TODO: confirm if we should use $estimated_payment_details['subtotal']['shipping'] for something
+		// TODO: confirm if we should use $estimated_payment_details['tax'] for something
+		$local_order->set_total( $estimated_payment_details['total_amount']['amount'] );
```
ChaseWiseman, line 148: We'll also need to set the order's currency along with its amount.

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		// TODO: confirm if we should use $estimated_payment_details['tax_remitted'] for something
+
+		// update information from buyer_details
+		$buyer_details = $remote_order->get_buyer_details();
+
+		$local_order->set_billing_last_name( $buyer_details['name'] );
+		$local_order->set_billing_email( $buyer_details['email'] );
+		// TODO: confirm if we should use $buyer_details['email_remarketing_option'] for something
```
ChaseWiseman, line 156: yeah let's save an order meta item for this value, in the standard Woo `yes|no` value format. Maybe `_wc_facebook_commerce_email_remarketing`?

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		$local_order->set_shipping_address_2( $shipping_address['street2'] );
+		$local_order->set_shipping_city( $shipping_address['city'] );
+		$local_order->set_shipping_state( $shipping_address['state'] );
+		$local_order->set_shipping_postcode( $shipping_address['postal_code'] );
+		$local_order->set_shipping_country( $shipping_address['country'] );
+
+		// update information from estimated_payment_details
+		$estimated_payment_details = $remote_order->get_estimated_payment_details();
+
+		// TODO: confirm if we should use $estimated_payment_details['subtotal']['items'] for something
```
ChaseWiseman, line 145: Would be good to confirm but I don't think we can use these order-level subtotal values, as Woo calculates them on the fly from the order items.

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+		}
+
+		// update information from selected_shipping_option
+		$selected_shipping_option = $remote_order->get_selected_shipping_option();
+
+		$local_order->set_shipping_total( $selected_shipping_option['price']['amount'] );
+		$local_order->set_shipping_tax( $selected_shipping_option['calculated_tax']['amount'] );
+
+		// update information from shipping_address
+		$shipping_address = $remote_order->get_shipping_address();
```
ChaseWiseman, line 132: Let's code defensively here and only set this customer data if it's returned in the API order. iirc I thought I remember seeing some case where pending orders don't send along customer data until it's finalized.

includes/Commerce/Orders.php
ChaseWiseman: For products that are synced via API, we also have stored the concrete Facebook product ID to post meta, so that could be potentially used for quick retrieval as well, then fall back to matching the retailer ID.

includes/Commerce/Orders.php
```diff
@@ -82,10 +82,86 @@ public function create_local_order( Order $remote_order ) {
...
+
+		$local_order->set_billing_last_name( $buyer_details['name'] );
+		$local_order->set_billing_email( $buyer_details['email'] );
+		// TODO: confirm if we should use $buyer_details['email_remarketing_option'] for something
+
+		// update order status
+		if ( Order::STATUS_CREATED === $remote_order->get_status() ) {
+			$local_order->set_status( 'processing' );
+		}
+
```
ChaseWiseman, line 162: One last thing we need to store is the remote order ID, which will be needed by the `find_local_order()` method

includes/Commerce/Orders.php
deivamagalhaes: Because we only have one `name` field, so we have nothing to set in the first name.

includes/Commerce/Orders.php
deivamagalhaes:
```suggestion
			// we use the estimated_tax because the captured_tax represents the tax after the order/item has been shipped and we don't fulfill order at the line-item level
```

includes/Commerce/Orders.php
deivamagalhaes: Does it return all the fields as empty strings or an empty array?

includes/Commerce/Orders.php
deivamagalhaes: Added in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/f86144a44cb465ab1c91e6f901da42c78a48eefc

includes/Commerce/Orders.php
deivamagalhaes: Added in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/c1cc330e9c3a4d82a1d355e9997a27a8bc4d22e5

includes/Commerce/Orders.php
ChaseWiseman: Right but in the [doc examples](https://developers.facebook.com/docs/commerce-platform/order-management/order-api) for buyer name (and very likely shipping name), they show a full name `John Smith`. I think we could at least split the string after the first space to try and separate out first and last. That won't cover all cases, but it should for a lot, and then next steps would be:
1. Test against a real API response and confirm that assumption, that the full name is returned
1. Bring the issue to Facebook to see if they can return distinct first and last

includes/Commerce/Orders.php
ChaseWiseman: I don't know - the docs aren't clear on that so it'll need to be tested against a real API response as well

deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/abe1ac3dbb93d742492913328bd58b8e4cfceac0

includes/Commerce/Orders.php
deivamagalhaes: Added in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/4de35ff0126f97eeb92ee6b519430a996136fda4

includes/Commerce/Orders.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/d91f21223c4063004478ee53582a873eb3397dac

includes/Commerce/Orders.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/a721ed6f18b03e1364bf27b91c974d31609baf8a

includes/Commerce/Orders.php
deivamagalhaes: Added in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/78ba9805cdc160505dd77a361b0350bd863ff256

deivamagalhaes:
@ChaseWiseman, I've tweaked most of the details you mentioned. 

For the ones we should confirm with them, should we do it on Slack or is there a better channel?

Questions we currently have:
- Is this `shipping_address` > `country` an alpha-2 or alpha-3 code?
- Is the `shipping_address` > `name` always a full name? Would it be possible for them to provide separate values for first and last name?
- Should we use the `estimated_payment_details` > `subtotal` for anything when creating the WC order?
- Should we use the `estimated_payment_details` > `tax` for anything when creating the WC order?
- Should we use the `estimated_payment_details` > `tax_remitted` for anything when creating the WC order?

ChaseWiseman:
@deivamagalhaes 

>Is this shipping_address > country an alpha-2 or alpha-3 code?

Looks like it's alpha-2 based on a real API response.

>Should we use the estimated_payment_details > subtotal for anything when creating the WC order?
>Should we use the estimated_payment_details > tax for anything when creating the WC order?

This is more a [question for our implementation](https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515#discussion_r479520057) - do you see anywhere they can be used, or is Woo always calculating these values based on the line items? I _think_ it's the latter, but we'll need to confirm that and probably add a doc note in there somewhere about the reasoning.

>Is the shipping_address > name always a full name? Would it be possible for them to provide separate values for first and last name?
>Should we use the estimated_payment_details > tax_remitted for anything when creating the WC order?

These have been passed along to FB in Slack.

includes/Commerce/Orders.php
deivamagalhaes: I've confirmed they send the country as alpha-2: https://cloud.skyver.ge/kpubePyG

includes/Commerce/Orders.php
deivamagalhaes: From their reply on Slack:

> 2. I don't believe there is a guarantee it will be the customer's full name. IIRC for IG it can sometimes by the IG user's handle. I also just confirmed on FB I am able to enter any arbitrary text in the Name field for the Delivery address.

I believe we don't have a lot of options here. Do you think we should split after the first space if there is a space?

ChaseWiseman: yeah I think that's the only thing we can do really, just try and find the first space

deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/a495851b6b6e52c5bf03e612c207c8981d595907.

deivamagalhaes:
@ChaseWiseman, I think I _almost_ got this, but I'm getting duplicate taxes (https://cloud.skyver.ge/X6uNXoy9) and I don't know why, as the data in the database seems to be the same as one of my regular orders, which looks ok. Help?

deivamagalhaes:
I'll try to add a single tax item!

deivamagalhaes:
Yay, that was it! 🎉

includes/Commerce/Orders.php
```diff
@@ -96,10 +99,190 @@ public function create_local_order( Order $remote_order ) {
...
+		$estimated_payment_details = $remote_order->get_estimated_payment_details();
+
+		// we do not use subtotal values from the API because WC calculates them on the fly based on the items
+		$local_order->set_total( $estimated_payment_details['total_amount']['amount'] );
+		$local_order->set_currency( $estimated_payment_details['total_amount']['currency'] );
+
+		// update information from buyer_details
+		$buyer_details = $remote_order->get_buyer_details();
+
+		if ( strpos( $buyer_details['name'], ' ' ) !== false ) {
```
ChaseWiseman, line 262: It looks like the buyer details are missing if the order is still `FB_PROCESSING`, so this throws a notice in that case.

ChaseWiseman (changes requested):
@deivamagalhaes works well! See one comment about notices being thrown for `FB_PROCESSING` orders - just need to check for keys there.

The only other thing I see that's missing is stock reduction. Per the doc, we want to reduce product stock even when the order is pending, so the stock is held immediately. Right now the stock does get reduced, but only once the order is `CREATED` and we mark it as Processing. We should be able to make use of `wc_reduce_stock_levels( $order )` to handle that cleanly.

As for tests, most are passing but I see 12 failures, I think all related to the `items{}` fields changes. One example from `tests/integration/API/Orders/Read/RequestTest.php:test_constructor`:

```
Failed asserting that two arrays are equal.
- Expected | + Actual
@@ @@
Array (
-    'fields' => 'id,order_status,created,last_updated,items,ship_by_date,merchant_order_id,channel,selected_shipping_option,shipping_address,estimated_payment_details,buyer_details'
+    'fields' => 'id,order_status,created,last_updated,items{id,retailer_id,product_id,quantity,price_per_unit,tax_details},ship_by_date,merchant_order_id,channel,selected_shipping_option,shipping_address,estimated_payment_details,buyer_details'
)
```

includes/Commerce/Orders.php
deivamagalhaes: Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515/commits/7b725748e1aac89a8cdb1347f0d604f28f7b0489.

ChaseWiseman: I'm afraid I still see notices for this case - maybe you'll need to check the individual parameters?

```
PHP Notice:  Undefined index: name in includes/Commerce/Orders.php on line 264
PHP Notice:  Undefined index: name in includes/Commerce/Orders.php on line 272
PHP Notice:  Undefined index: email in includes/Commerce/Orders.php on line 275
```

ChaseWiseman (commented):
@deivamagalhaes other issues are fixed. I'm still seeing notices as per https://github.com/facebookincubator/facebook-for-woocommerce/pull/1515#discussion_r483791059

includes/Commerce/Orders.php
deivamagalhaes: This is weird, can you send me the API response from the logs for the `FB_PROCESSING` order?

ChaseWiseman: sure! Looks like `email_remarketing_option` still returns regardless

```json
{
  "data": [
    {
      "id": "*********",
      "order_status": {
        "state": "FB_PROCESSING"
      },
      "created": "2020-09-04T18:34:57+00:00",
      "last_updated": "2020-09-04T18:35:08+00:00",
      "items": {
        "data": [
          {
            "id": "*********",
            "retailer_id": "*********",
            "product_id": "*********",
            "quantity": 3,
            "price_per_unit": {
              "amount": "1.00",
              "currency": "USD"
            },
            "tax_details": {
              "estimated_tax": {
                "amount": "0.29",
                "currency": "USD"
              },
              "captured_tax": {
                "total_tax": {
                  "amount": "0.00",
                  "currency": "USD"
                }
              }
            }
          }
        ],
        "paging": {
          "cursors": {
            "before": "*********",
            "after": "*********"
          }
        }
      },
      "ship_by_date": "2020-09-10",
      "channel": "facebook",
      "selected_shipping_option": {
        "name": "STANDARD (3 - 5 Business days)",
        "price": {
          "amount": "1.00",
          "currency": "USD"
        },
        "calculated_tax": {
          "amount": "0.09",
          "currency": "USD"
        },
        "estimated_shipping_time": {
          "min_days": 3,
          "max_days": 5
        }
      },
      "estimated_payment_details": {
        "subtotal": {
          "items": {
            "amount": "3.00",
            "currency": "USD"
          },
          "shipping": {
            "amount": "1.00",
            "currency": "USD"
          }
        },
        "tax": {
          "amount": "0.38",
          "currency": "USD"
        },
        "total_amount": {
          "amount": "4.38",
          "currency": "USD"
        },
        "tax_remitted": true
      },
      "buyer_details": {
        "email_remarketing_option": false
      }
    }
  ],
  "paging": {
    "cursors": {
      "before": "*********",
      "after": "*********"
    }
  }
}
```

deivamagalhaes: Yikes, could they be weirder? 🤷‍♀️ 
I'll tweak the ifs...

ChaseWiseman (approved):

PR #1520, opened 2020-08-28
Implement the update local orders callback method

deivamagalhaes:
# Summary

This PR implements the `Commerce\Orders::update_local_orders()` method.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Commerce/Orders.php
+++ b/includes/Commerce/Orders.php
@@ -124,7 +124,7 @@ class Orders {
 					$local_order = $this->update_local_order( $remote_order, $local_order );
 				}
 
-			} catch ( SV_WC_Plugin_Exception $exception ) {
+			} catch ( \Exception $exception ) {
 
 				if ( ! empty( $local_order ) ) {
 					// add note to order
--- a/includes/Commerce/Orders.php
+++ b/includes/Commerce/Orders.php
@@ -101,7 +101,7 @@ class Orders {
 
 		try {
 
-			$response = facebook_for_woocommerce()->get_api()->get_new_orders( $page_id );
+			$response = facebook_for_woocommerce()->get_api( facebook_for_woocommerce()->get_connection_handler()->get_page_access_token() )->get_new_orders( $page_id );
 
 		} catch ( SV_WC_API_Exception $exception ) {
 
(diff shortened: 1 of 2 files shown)


includes/Commerce/Orders.php
```diff
@@ -96,7 +97,55 @@ public function update_local_order( Order $remote_order, \WC_Order $local_order
 	 */
 	public function update_local_orders() {
 
-		// TODO: implement
+		$page_id = facebook_for_woocommerce()->get_integration()->get_facebook_page_id();
+
+		try {
+
+			$response = facebook_for_woocommerce()->get_api()->get_new_orders( $page_id );
```
ChaseWiseman, line 104: We'll need to pass in the page access token for this API request, which we can now after implementing [this story](https://app.clubhouse.io/skyverge/story/62622/update-the-base-api-getter-to-accept-an-access-token)

includes/Commerce/Orders.php
```diff
@@ -96,7 +97,55 @@ public function update_local_order( Order $remote_order, \WC_Order $local_order
...
+			} catch ( \Exception $exception ) {
+
+				if ( ! empty( $local_order ) ) {
+					// add note to order
+					$local_order->add_order_note( 'Error updating local order from Commerce order from the Orders API: ' . $exception->getMessage() );
+				} else {
+					facebook_for_woocommerce()->log( 'Error creating local order from Commerce order from the Orders API: ' . $exception->getMessage() );
+				}
+
+				return;
```
ChaseWiseman, line 136: I think this should be `continue` to keep processing remaining orders, right?

includes/Commerce/Orders.php
```diff
@@ -96,7 +97,55 @@ public function update_local_order( Order $remote_order, \WC_Order $local_order
...
+			}
+
+			if ( ! empty( $local_order ) && Order::STATUS_CREATED === $remote_order->get_status() ) {
+
+				// acknowledge the order
+				try {
+					facebook_for_woocommerce()->get_api()->acknowledge_order( $remote_order->get_id(), $local_order->get_id() );
```
ChaseWiseman, line 143: this will also need the page access token

includes/Commerce/Orders.php
deivamagalhaes: Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1520/commits/5c72c5e8b4d3b62a9761eddd85f85824051f82e9

includes/Commerce/Orders.php
deivamagalhaes: Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1520/commits/5c72c5e8b4d3b62a9761eddd85f85824051f82e9

PR #1525, opened 2020-09-02
Implement the Commerce\Orders::add_order_refund() method

deivamagalhaes:
# Summary

This PR implements the `Commerce\Orders::add_order_refund()` method.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/Commerce/Orders.php
+++ b/includes/Commerce/Orders.php
@@ -12,6 +12,7 @@ namespace SkyVerge\WooCommerce\Facebook\Commerce;
 
 use SkyVerge\WooCommerce\Facebook\API\Orders\Cancel\Request;
 use SkyVerge\WooCommerce\Facebook\API\Orders\Order;
+use SkyVerge\WooCommerce\Facebook\API\Orders\Refund\Request as Refund_Request;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_API_Exception;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WC_Plugin_Exception;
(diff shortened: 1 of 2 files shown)


includes/Commerce/Orders.php
```diff
@@ -298,12 +299,82 @@ public function fulfill_order( \WC_Order $order, $tracking_number, $carrier ) {
...
+			$remote_id = $parent_order->get_meta( self::REMOTE_ID_META_KEY );
+
+			if ( ! $remote_id ) {
+				throw new SV_WC_Plugin_Exception( __( 'Remote ID for parent order not found.', 'facebook-for-woocommerce' ) );
+			}
+
+			$items = [];
+
+			/** @var \WC_Order_Item_Product $item */
+			foreach ( $refund->get_items() as $item ) {
```
ChaseWiseman, line 341: Per the docs:

>For partial refund, specify the item level breakdown. Not required for full refund.

We should probably only include items if the refund is partial

includes/Commerce/Orders.php
```diff
@@ -298,12 +299,82 @@ public function fulfill_order( \WC_Order $order, $tracking_number, $carrier ) {
...
+
+				if ( $product = $item->get_product() ) {
+
+					$items[] = [
+						'retailer_id'          => \WC_Facebookcommerce_Utils::get_fb_retailer_id( $product ),
+						'item_refund_quantity' => $item->get_quantity(),
```
ChaseWiseman, line 347: Refund item quantities are negative, so this will need to be converted to positive. I _think_ that's consistent across our minimum supported WC versions, but that needs to be confirmed to handle both cases if not.

Also, there's no guarantee that every line item will have a quantity. See the amount inputs in the refund UI - you can bump the quantity but also input amounts free-form. Looks like `item_refund_amount` could be used in that case.

includes/Commerce/Orders.php
```diff
@@ -298,12 +299,82 @@ public function fulfill_order( \WC_Order $order, $tracking_number, $carrier ) {
...
+				throw new SV_WC_Plugin_Exception( __( 'No valid Facebook products were found.', 'facebook-for-woocommerce' ) );
+			}
+
+			$refund_data = [
+				'items'       => $items,
+				'reason_code' => $reason_code,
+				'reason_text' => $refund->get_reason(),
+				'shipping'    => [
+					'amount'   => $refund->get_shipping_total(),
+					'currency' => $refund->get_currency(),
```
ChaseWiseman, line 362: According to the doc examples, this needs to be wrapped in a `shipping_refund` key.

includes/Commerce/Orders.php
```diff
@@ -298,12 +299,82 @@ public function fulfill_order( \WC_Order $order, $tracking_number, $carrier ) {
...
+
+			if ( empty( $items ) ) {
+				throw new SV_WC_Plugin_Exception( __( 'No valid Facebook products were found.', 'facebook-for-woocommerce' ) );
+			}
+
+			$refund_data = [
+				'items'       => $items,
+				'reason_code' => $reason_code,
+				'reason_text' => $refund->get_reason(),
+				'shipping'    => [
```
ChaseWiseman, line 360: We should also only include this if there is a shipping total. Partial refunds can be made without including shipping in the refund.

includes/Commerce/Orders.php
deivamagalhaes: Ok, what is the easiest way to know if it is partial? Comparing the total with the parent order's total?

includes/Commerce/Orders.php
deivamagalhaes: Do you mean always send the `item_refund_amount`, instead of the quantity?

includes/Commerce/Orders.php
deivamagalhaes: Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1525/commits/a5d4690d9d39acc799125b3a46dd132914158219.

includes/Commerce/Orders.php
deivamagalhaes: Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1525/commits/af67f4a09bf2ac6e463e7519fdb834ce6252a6b8.

includes/Commerce/Orders.php
ChaseWiseman: I think so. There's prior art in our plugins, like this [AvaTax example](https://github.com/skyverge/wc-plugins/blob/master/woocommerce-avatax/includes/class-wc-avatax-ajax.php#L396)

includes/Commerce/Orders.php
ChaseWiseman: I was thinking, based on the doc descriptions, that if a quantity exists use that, then if not use the amount. What do you think?

includes/Commerce/Orders.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1525/commits/29661eb37e27f8f5f8c639e02a5650720d1415d4.

includes/Commerce/Orders.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1525/commits/da235caba6d375faeed5e0d9fd4fd7aa126dfb12.

ChaseWiseman (approved):

PR #1526, opened 2020-09-03
Implement rate limit and delay handling

ChaseWiseman:
# Summary

This PR implements the rate limit handling for API requests.

## Details

Sets the rate limit delay timestamp whenever the associated error codes are detected, and halts future API requests of that type until the delay has passed.

Uses transients to store the delay values so that they are cleared periodically in case the options get stuck in a delayed state, so requests are at least let through every 24 hours.

Adds a `get_throttle_end()` method to the rate limit exception so that potentially in the future we can display more detailed admin notices about when requests should be re-attempted.

## QA

- [x] Integrations test pass

## Before merge

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -23,7 +23,7 @@ use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
  *
  * @since 2.0.0
  *
- * @method Response perform_request( $request )
+ * @method API\Request get_request()
  */
 class API extends Framework\SV_WC_API_Base {
 
@@ -97,6 +97,7 @@ class API extends Framework\SV_WC_API_Base {
 
 		/** @var API\Response $response */
 		$response = $this->get_response();
+		$request  = $this->get_request();
 
 		if ( $response && $response->has_api_error() ) {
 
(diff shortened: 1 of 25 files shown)


includes/API.php
```diff
@@ -84,6 +84,34 @@ public function set_access_token( $access_token ) {
...
+	 * @since 2.1.0-dev.1
+	 *
+	 * @param API\Request $request request object
+	 * @return API\Response
+	 * @throws API\Exceptions\Request_Limit_Reached|Framework\SV_WC_API_Exception
+	 */
+	public function perform_request( $request ) {
+
+		$rate_limit_id   = $request::get_rate_limit_id();
+		$delay_timestamp = $this->get_rate_limit_delay( $request::get_rate_limit_id() );
```
deivamagalhaes, line 99:
```suggestion
		$delay_timestamp = $this->get_rate_limit_delay( $rate_limit_id );
```

deivamagalhaes (changes requested):
Just a tiny suggestion and a tweak to the tests (added in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1526/commits/41a70a7379939dbe2a5c30e6dfb01965255a66d0).

PR #1527, opened 2020-09-03
Hook into product stock change and trigger a product sync

deivamagalhaes:
# Summary

This PR implements the `Products\Sync::handle_stock_update()` method and hooks it into the stock update actions.

## QA

- [x] Integrations test pass

### Setup

- Have the plugin configured, connected, and with logging enabled
- Have two products with price and inventory set, sync enabled, Commerce enabled (set the post `_wc_facebook_commerce_enabled` meta data to `yes`), and a Google product category set (set the post `_wc_facebook_google_product_category`)

#### Order with a single product

1. Place an order with one of the products
    - [x] One API call is made to update the product and it sends the `availability` and `inventory` values

#### Order with two products

1. Place an order with the two products
    - [x] One batch API call is made to update both products and it sends the `availability` and `inventory` values

#### Admin editing a single product 

1. Update the inventory for one of the products via the admin UI
    - [x] One API call is made to update the product and it sends the `availability` and `inventory` values

#### Bulk admin editing 

1. Update the inventory for both products via a bulk action
    - [x] Two API calls are made (one to update each product) and send the `availability` and `inventory` values

## Before merge

### DIFF
--- a/includes/Products/Sync.php
+++ b/includes/Products/Sync.php
@@ -54,6 +54,10 @@ class Sync {
 	public function add_hooks() {
 
 		add_action( 'shutdown', [ $this, 'schedule_sync' ] );
+
+		// stock update actions
+		add_action( 'woocommerce_product_set_stock', [ $this, 'handle_stock_update' ] );
+		add_action( 'woocommerce_variation_set_stock', [ $this, 'handle_stock_update' ] );
 	}
 
 
--- a/includes/Products/Sync.php
+++ b/includes/Products/Sync.php
@@ -158,6 +158,18 @@ class Sync {
 
 
 	/**
+	 * Adds the products with stock changes to the requests array to be updated.
+	 *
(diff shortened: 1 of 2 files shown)


deivamagalhaes:
@ChaseWiseman , I am uncertain about one of the acceptance criteria:
> If multiple product's stocks are updated (like an order with multiple items), the plugin sends only one batch API request with all of the product data

Does this also apply to admin bulk updates? What I am seeing for admin bulk updates is one API for each product.

ChaseWiseman:
@deivamagalhaes any admin functionality can remain as-is, so as long as that's the case in `master` then that's fine. We only want to make sure that the addition of this callback handling doesn't trigger a _duplicate_ API request in those cases

deivamagalhaes:
It is not making duplicate API requests, just a separate request for each product. Thanks for clarifying.

ChaseWiseman (approved):

PR #1528, opened 2020-09-04
Implement retry in API requests

ChaseWiseman:
# Summary

This PR implements the retry handling for API requests.

Please do not merged until #1526 is merged.

## Details

This updates `API\Request` with some retry helper methods. Each request can now have:
- A list of response error codes that indicate the request should be retried
- A limit on the number of retries (default 5) to prevent a loop
- The number of times the request has been retried

The API will retry a request with known codes until that limit has been reached, then it will throw an exception as it normally would on any non-retry error.

So far this only applies to a single error code for Orders API requests, but this will allow us to add more in the future as we find them.

## QA

- [ ] Integrations test pass

## Before merge

### DIFF
--- a/includes/API/Request.php
+++ b/includes/API/Request.php
@@ -101,7 +101,7 @@ class Request extends Framework\SV_WC_API_JSON_Request {
 
 
 	/**
-	 * Gets the number of retries to attempt if told to do so by Facebook.
+	 * Gets the maximum number of retries to attempt if told to do so by Facebook.
 	 *
 	 * @since 2.1.0-dev.1
 	 *
--- a/includes/API/Request.php
+++ b/includes/API/Request.php
@@ -25,6 +25,16 @@ class Request extends Framework\SV_WC_API_JSON_Request {
 	use Traits\Rate_Limited_Request;
 
 
+	/** @var int number of retries to attempt if told to do so by Facebook */
+	protected $retry_limit = 5;
+
+	/** @var int number of times this request has been retried */
+	protected $retry_count = 0;
+
+	/** @var int[] the response codes that should trigger */
+	protected $retry_codes = [];
+
+
 	/**
 	 * API request constructor.
 	 *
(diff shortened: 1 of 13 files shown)


includes/API/Request.php
```diff
@@ -25,6 +25,16 @@ class Request extends Framework\SV_WC_API_JSON_Request {
 	use Traits\Rate_Limited_Request;
 
 
+	/** @var int number of retries to attempt if told to do so by Facebook */
+	protected $retry_limit = 5;
+
+	/** @var int number of times this request has been retried */
+	protected $retry_count = 0;
+
+	/** @var int[] the response codes that should trigger */
```
deivamagalhaes, line 34:
```suggestion
	/** @var int[] the response codes that should trigger a retry */
```

includes/API/Request.php
```diff
@@ -25,6 +25,16 @@ class Request extends Framework\SV_WC_API_JSON_Request {
 	use Traits\Rate_Limited_Request;
 
 
+	/** @var int number of retries to attempt if told to do so by Facebook */
```
deivamagalhaes, line 28:
```suggestion
	/** @var int maximum number of retries to attempt if told to do so by Facebook */
```

includes/API/Request.php
```diff
@@ -66,4 +76,54 @@ public function set_data( $data ) {
...
+	 * @since 2.1.0-dev.1
+	 */
+	public function mark_retry() {
+
+		$this->retry_count++;
+	}
+
+
+	/**
+	 * Gets the number of retries to attempt if told to do so by Facebook.
```
deivamagalhaes, line 104:
```suggestion
	 * Gets the maximum number of retries to attempt if told to do so by Facebook.
```

includes/API/Request.php
```diff
@@ -66,4 +76,54 @@ public function set_data( $data ) {
...
+	 *
+	 * @since 2.1.0-dev.1
+	 *
+	 * @return int
+	 */
+	public function get_retry_limit() {
+
+		return $this->retry_limit;
```
deivamagalhaes, line 112: Is this something we may want to filter?

tests/integration/APITest.php
```diff
@@ -64,6 +64,36 @@ public function test_set_access_token() {
...
+			'get_parsed_response'    => \Codeception\Stub\Expected::exactly( 6, new API\Response( json_encode( [
+				'error' => [
+					'message'          => 'Retry me!',
+					'type'             => 'OAuthException',
+					'code'             => 5678,
+				]
+			] ) ) ),
+		] );
+
+		$this->expectException( Framework\SV_WC_API_Exception::class );
```
deivamagalhaes, line 91: We may want to add a comment explaining it will be retried 5 times and then throw an exception, just to make things a bit clearer.

deivamagalhaes (commented):
Just a few suggestions/questions.

tests/integration/APITest.php
ChaseWiseman: Added with 35be89b24d28a7bc4dfdd3c4ab28941888353b41

includes/API/Request.php
ChaseWiseman: Sure! added with c0a2519d475e923039a91d082b8e9f3866f184df

deivamagalhaes (approved):
👍

PR #1540, opened 2020-09-11
Add the shipment utilities class

deivamagalhaes:
# Summary

This PR adds the `Utilities\Shipment` class.

## Open questions

- [ ] I am unsure about a few Shipment Tracking carriers. I've added a TODO comment next to each of them. Should we map these to "OTHER" instead? Or is there a reliable source where we can find this information?

## QA

- [x] `tests/integration/Utilities/ShipmentTest.php` tests pass

## Before merge


includes/Utilities/Shipment.php
```diff
@@ -0,0 +1,674 @@
...
+			'YUN_EXPRESS'                                       => 'Yun Express',
+			'ZEPTOEXPRESS'                                      => 'ZeptoExpress',
+			'ZINC'                                              => 'Zinc',
+			'ZJS_INTERNATIONAL'                                 => 'ZJS International',
+			'ZTO_EXPRESS'                                       => 'ZTO Express',
+			'ZYLLEM'                                            => 'Zyllem',
+		];
+
+		/**
+		 * @see https://docs.woocommerce.com/document/shipment-tracking/?_ga=2.193545197.1907176699.1599852448-1652361994.1593116723&_gac=1.227990895.1596051708.Cj0KCQjwvIT5BRCqARIsAAwwD-QunfzzbbyhGEEVmH4kdtPM1uID0g72ial0LcVu0eGdkLBcv_okGwcaAjdGEALw_wcB#section-5
```
wvega, line 552: I think we can drop the tracking parameters:

```suggestion
		 * @see https://docs.woocommerce.com/document/shipment-tracking/#section-5
```

wvega (approved):
This looks good to me @deivamagalhaes. Only have one suggestion. 

>Should we map these to "OTHER" instead? Or is there a reliable source where we can find this information?

I'm not able to suggest better defaults for the Shipment Tracking carriers were you left notes. I think the ones you selected are more appropriate than using OTHER. 

As for a reliable source, perhaps it we can bring up the question to the developers of Shipment Tracking asking them how would they map those values to Facebook codes.

includes/Utilities/Shipment.php
deivamagalhaes: Yeah, that's a lot better, thanks!

deivamagalhaes:
@ChaseWiseman, can we ask Facebook or Shipment Tracking about the codes that we don't know?

ChaseWiseman:
@deivamagalhaes yes and I concur with @wvega and think in this case, the best source of truth would be the a8c team (they manage Shipment tracking) as most of the TODOs relate to whether to use a specific version of the Facebook code, so they can likely clarify the intention behind the Shipment Tracking version

deivamagalhaes:
Question posted on Slack: https://a8c.slack.com/archives/C07414X4N/p1600368624044100

deivamagalhaes:
The answer I got on Slack wasn't very useful. I'd say let's leave as is for now and we can review later if needed. I'll merge this PR to unblock other stories.

PR #1542, opened 2020-09-11
Add the Google Product Category field JS class

deivamagalhaes:
# Summary

This PR adds the `WC_Facebook_Google_Product_Category_Fields` JS class with its stub methods.

## QA

- [x] Code review only

## Before merge


includes/Admin.php
```diff
@@ -142,6 +142,8 @@ public function enqueue_scripts() {
 				wp_enqueue_script( 'wc-enhanced-select' );
 			}
 		}
+
+		wp_enqueue_script( 'wc-facebook-google-product-category-fields', plugins_url( '/facebook-for-woocommerce/assets/js/admin/google-product-category-fields.min.js' ), [ 'jquery' ], \WC_Facebookcommerce::PLUGIN_VERSION );
```
ChaseWiseman, line 146: Can use `facebook_for_woocommerce()->get_plugin_url()` for this URL now that the plugin is frameworked

ChaseWiseman (changes requested):
Looks good! Just a note about using a framework method

includes/Admin.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1542/commits/9b2148faf24012cb6ca1da4b845731db88dbfc26.

ChaseWiseman (approved):

PR #1545, opened 2020-09-15
Implement the Google Product Category field getter

deivamagalhaes:
# Summary

This PR implements the `Google_Product_Category_Field::get_categories()` method.

## Details

I've added a couple of auxiliar methods, just so this one wasn't larger than life.

I've also refactored the categories' structure as [discussed on the implementation doc](https://docs.google.com/document/d/11AG_ZVIYGKn96BvEn_Qwy91FOALeE7GmB5cpfN1D6LU/edit?disco=AAAAJ4F5PxA), to be a single level array of categories, indexed by their IDs, with their:
- label
- parent
- options (key-value array with the options' labels, to avoid the JS having to lookup the label for each option)

The JS will be able to retrieve the options (IDs and labels) for any category by looking for the category ID on the array keys and grabbing its `options` value.

## QA

- [x] `tests/integration/Admin/GoogleProductCategoryFieldTest.php` tests pass

## Before merge


ChaseWiseman (approved):
works for me!

PR #1547, opened 2020-09-15
Implement the constructor for the WC Google Product Category Field JS class

wvega:
# Summary

This PR inserts a wrapper element to hold the select fields used to set the Google Product Category.

### Steps

1. Go to WooCommerce > Facebook > Instagram Checkout
    - [x] A `div` element with ID `#wc-facebook-google-product-category-fields` was inserted above the input element that holds the ID of the category


ChaseWiseman (approved):
@wvega looks good! approved, though it looks like we'll wait to merge until its base PR is merged.

PR #1549, opened 2020-09-15
Admin\Product_Categories - Implement the field rendering methods

deivamagalhaes:
# Summary

This PR implements the `Admin\Product_Categories::get_google_product_category_field_title()`, `Admin\Product_Categories::render_google_product_category_tooltip()`, `Admin\Product_Categories::render_add_google_product_category_field()`, and `Admin\Product_Categories::render_edit_google_product_category_field()` methods.

## QA

- [x] `tests/integration/Admin/ProductCategoriesTest.php` pass

### Setup

- The plugin is active and connected

### Steps

1. Navigate to Products > Categories
    - [x] The add form displays the field label "Default Google product category"
       - [x] The field label displays the tooltip icon
          - [x] The tooltip displays the text "Choose a default Google product category for products in this category. Products need at least two category levels defined to be sold on Instagram."
1. Click to edit a category
    - [x] The edit form displays the field label "Default Google product category"
       - [x] The field label displays the tooltip icon
          - [x] The tooltip displays the text "Choose a default Google product category for products in this category. Products need at least two category levels defined to be sold on Instagram."

## Before merge

- [ ] I have confirmed these changes in each supported minor WooCommerce version

PR #1551, opened 2020-09-16
Admin\Product_Categories - Implement the field saving method

deivamagalhaes:
# Summary

This PR implements the `Admin\Product_Categories:: save_google_product_category()` method.

## Details

This PR was branched off `ch63657/products-categories-field-render`, please merge #1549 first.

## QA

- [x] `tests/integration/Admin/ProductCategoriesTest.php` pass

Manual QA is not possible yet because the JS to create the selects is not implemented.

## Before merge

### DIFF
--- a/includes/Admin/Product_Categories.php
+++ b/includes/Admin/Product_Categories.php
@@ -165,7 +165,6 @@ class Product_Categories {
 			// get the products in the category being saved
 			$products = wc_get_products(
 				[
-					'type'     => ['simple', 'variable'],
 					'category' => [ $term->slug ],
 				]
 			);
--- a/includes/Admin/Product_Categories.php
+++ b/includes/Admin/Product_Categories.php
@@ -158,19 +158,22 @@ class Product_Categories {
 
 		\SkyVerge\WooCommerce\Facebook\Product_Categories::update_google_product_category_id( $term_id, $google_product_category_id );
 
-		$term = get_term( $term_id, $taxonomy, ARRAY_A );
+		$term = get_term( $term_id, $taxonomy );
 
-		// get the products in the category being saved
-		$product_ids = wc_get_products(
-			[
-				'return'   => 'ids',
-				'category' => [ $term['slug'] ],
-			]
-		);
+		if ( $term instanceof \WP_Term ) {
 
-		if ( ! empty( $product_ids ) ) {
+			// get the products in the category being saved
+			$product_ids = wc_get_products(
+				[
+					'return'   => 'ids',
+					'category' => [ $term->slug ],
+				]
+			);
 
-			facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_products( $product_ids );
+			if ( ! empty( $product_ids ) ) {
+
+				facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_products( $product_ids );
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


includes/Admin/Product_Categories.php
```diff
@@ -143,10 +148,30 @@ public function get_google_product_category_field_title() {
...
 	 *
-	 * @return string
+	 * @param int $term_id term ID
+	 * @param int $tt_id term taxonomy ID
+	 * @param string $taxonomy Taxonomy slug
 	 */
-	public function save_google_product_category() {
+	public function save_google_product_category( $term_id, $tt_id, $taxonomy ) {
+
+		$google_product_category_id = Framework\SV_WC_Helper::get_posted_value( self::FIELD_GOOGLE_PRODUCT_CATEGORY_ID );
```
wvega, line 157: I think we should clean the value and perhaps ensure it is a numeric string.

```suggestion
		$google_product_category_id = wc_clean( Framework\SV_WC_Helper::get_posted_value( self::FIELD_GOOGLE_PRODUCT_CATEGORY_ID ) );
```

includes/Admin/Product_Categories.php
```diff
@@ -143,10 +148,30 @@ public function get_google_product_category_field_title() {
...
+
+		\SkyVerge\WooCommerce\Facebook\Product_Categories::update_google_product_category_id( $term_id, $google_product_category_id );
+
+		$term = get_term( $term_id, $taxonomy, ARRAY_A );
```
wvega, line 161: We should check that `get_term()` returned an array instead of `null` or `WP_Error`.

includes/Admin/Product_Categories.php
```diff
@@ -143,10 +148,30 @@ public function get_google_product_category_field_title() {
...
+		$product_ids = wc_get_products(
+			[
+				'return'   => 'ids',
+				'category' => [ $term['slug'] ],
+			]
+		);
+
+		if ( ! empty( $product_ids ) ) {
 
+			facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_products( $product_ids );
```
wvega, line 173: We should only pass the ID of simple products and product variations to `create_or_update_products()`. That method doesn't check the type of the product and we may end up using the Batch API to try to sync info for a variable product instead of the individual variations.

Simon and I ran into that accidentally on a support conversation and the plugin started creating a variation with zero price in addition to the normal variations of variable products.

wvega (changes requested):
@deivamagalhaes changes are looking good, but I think we should add a sanity check and ensure we only attempt to sync individual product variations and simple products (instead of the variable products).

includes/Admin/Product_Categories.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1551/commits/c6bf98999488a4f950272fd83249940ab5b32c74.

includes/Admin/Product_Categories.php
deivamagalhaes: Updated in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1551/commits/7ee343bad8659462d54dcb4f1096cca6e68ac08a.

wvega (approved):
@deivamagalhaes thank you for the updates. Changes look good. LGTM!

PR #1553, opened 2020-09-16
[WIP] Add Enhanced Catalog support

maxdupenois:
This pull request is just to make it easier to see what I'm doing, a couple of things to note as currently missing:

1. Unit Tests/Integration Tests
2. Minimising the js file I've edited
3. Some sort of auto format support (O_o? VScode letting me down with the tab alignment in here)

*NB:* I will probably break this code into smaller commits, when I have a clearer idea of direction consider this a proof of concept or spike.

I'll put some comments on things I think are worth noting.

Screenshot 2020-09-16 at 13 07 47
Screenshot 2020-09-16 at 13 08 02


includes/Products/FBCategories.php
```diff
@@ -0,0 +1,165 @@
...
+namespace SkyVerge\WooCommerce\Facebook\Products;
+
+defined( 'ABSPATH' ) or exit;
+
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
+
+/**
+ * The main product feed handler.
+ *
+ * This will eventually replace \WC_Facebook_Product_Feed as we refactor and move its functionality here.
```
maxdupenois, line 20: hah oops, that comment shows which file I copied this header out of :p

includes/Products/FBCategories.php
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * @since 1.11.0
+ */
+class FBCategories {
```
maxdupenois, line 24: This is the class that does most of the work loading and maintaining the attributes. It also sets up the hooks for the ajax calls.

includes/Products/FBCategories.php
```diff
@@ -0,0 +1,165 @@
...
+    );
+
+		add_action(
+      'wp_ajax_wc_facebook_category_attributes',
+      [$this, 'handle_category_attribute_selection'],
+      self::ACTION_PRIORITY,
+    );
+	}
+
+	private function ensure_data_is_loaded(){
```
maxdupenois, line 62: No idea why this tabbing is all over the place, will need to find an autoformat that works properly

includes/fbproduct.php
```diff
@@ -631,6 +681,61 @@ public function prepare_product( $retailer_id = null, $prepare_for_product_feed
...
+
+		/**
+		* Adds enhanced catalog fields to product data array. Separated from
+		* the main function to make it easier to develop and debug, potentially
+		* worth refactoring into main prepare_product function when complete.
+		*
+		* @param array $product_data map
+		* @return array
+		*/
+		private function apply_enhanced_catalog_fields_from_attributes($product_data) {
```
maxdupenois, line 715: Function that actually applies any enhanced catalog fields we have access to

includes/fbproduct.php
```diff
@@ -631,6 +681,61 @@ public function prepare_product( $retailer_id = null, $prepare_for_product_feed
...
+			// image_url is used on the batch api, the same field is image_link on the feed
+			// url is used on the batch api, the same field is link on the feed
+
+			$fb_category_id = $this->get_fb_category();
+			$category_handler = facebook_for_woocommerce()->get_facebook_category_handler();
+			if(!$fb_category_id || !$category_handler->is_category($fb_category_id)){
+				return $product_data;
+			}
+			$enhanced_data = array();
+			$enhanced_data['fb_product_category'] = $fb_category_id;
```
maxdupenois, line 727: So this will be google product category in a bit, making those changes now but the attribute mapping is more awkward as there are so many more categories.

maxdupenois:
Changing the way this is going to be merged based on skyverge's pitch document.

PR #1565, opened 2020-09-18
Check for the commerce manager id when determining connection status

adrianostanley:
# Summary

This PR will add another condition to determine whether the site is connected by also checking for a commerce manager ID.

## Details

This PR comes from a fork for Facebook for WooCommerce: Instagram checkout - frontend by SkyVerge.

## QA

- [x] `tests/integration/CommerceTest.php` tests pass
- [x] `tests/integration/Handlers/ConnectionTest.php` tests pass

## Before merge

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -983,6 +983,19 @@ class Connection {
 
 
 	/**
+	 * Determines whether the Commerce has a manager ID.
+	 *
+	 * @since 2.1.0-dev.1
+	 *
+	 * @return bool
+	 */
+	public function has_commerce_manager_id() {
+
+		return ! empty( $this->get_commerce_manager_id() );
+	}
+
+
+	/**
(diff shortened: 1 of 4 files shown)


includes/Handlers/Connection.php
```diff
@@ -982,6 +982,19 @@ public function is_connected() {
...
 
 
+	/**
+	 * Determines whether the Commerce has a manager ID.
+	 *
+	 * @since 2.1.0-dev.1
+	 *
+	 * @return bool
+	 */
+	public function has_commerce_manager_id() {
```
deivamagalhaes, line 992: Are you sure we need this helper method? Is it going to be used anywhere else? In this case I think having the whole if inside the `is_connected()` may be preferable...

adrianostanley: I added that to improve readability, but yeah, I will do that in the `is_connected()` method.

deivamagalhaes (approved):
LGTM!

PR #1567, opened 2020-09-18
Implement the email blocking for Admin Orders

adrianostanley:
# Summary

This PR will provide an implementation to the `Orders:maybe_stop_order_email` method.

## Details

This PR comes from a fork for Facebook for WooCommerce: Instagram checkout - frontend by SkyVerge.

## QA

- [x] `tests/integration/Admin/OrdersTest.php` tests pass

## Before merge

### DIFF
--- a/includes/Admin/Orders.php
+++ b/includes/Admin/Orders.php
@@ -174,11 +174,12 @@ class Orders {
 			 * Filters the flag used to determine whether the email is enabled.
 			 *
 			 * @param bool $is_enabled whether the email is enabled
+			 * @param \WC_Order $order order object
 			 * @param Orders $this admin orders instance
 			 * @since 2.1.0-dev.1
 			 *
 			 */
-			$is_enabled = (bool) apply_filters( 'wc_facebook_commerce_send_woocommerce_emails', $is_enabled, $this );
+			$is_enabled = (bool) apply_filters( 'wc_facebook_commerce_send_woocommerce_emails', $is_enabled, $order, $this );
 		}
 
 		return $is_enabled;
--- a/includes/Admin/Orders.php
+++ b/includes/Admin/Orders.php
@@ -159,6 +159,8 @@ class Orders {
 	 */
 	public function maybe_stop_order_email( $is_enabled, $order ) {
 
+		$is_enabled = $is_enabled && ! ( $order instanceof \WC_Order && \SkyVerge\WooCommerce\Facebook\Commerce\Orders::is_commerce_order( $order ) );
+
 		return $is_enabled;
 	}
 
--- a/includes/Admin/Orders.php
+++ b/includes/Admin/Orders.php
@@ -161,6 +161,16 @@ class Orders {
 
 		$is_enabled = $is_enabled && ! ( $order instanceof \WC_Order && \SkyVerge\WooCommerce\Facebook\Commerce\Orders::is_commerce_order( $order ) );
 
+		/**
+		 * Filters the flag used to determine whether the email is enabled.
+		 *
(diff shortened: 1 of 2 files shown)


tests/integration/Admin/OrdersTest.php
```diff
@@ -46,7 +47,44 @@ protected function _after() {
...
+	/**
+	 * @see Admin\Orders::maybe_stop_order_email()
+	 *
+	 * @throws WC_Data_Exception
+	 */
+	public function test_maybe_stop_order_email() {
+
+		$orders_handler = $this->get_orders_handler();
+
+		$this->assertFalse( $orders_handler->maybe_stop_order_email( false, null ) );
```
ChaseWiseman, line 60: How about using a provider to consolidate this test method down to a single assert? See [this example](https://github.com/facebookincubator/facebook-for-woocommerce/blob/epic/instagram-checkout/tests/integration/Commerce/OrdersTest.php#L45-L71) which is similar to what could be added here.

includes/Admin/Orders.php
```diff
@@ -159,6 +159,18 @@ public function handle_bulk_update() {
...
+
+		/**
+		 * Filters the flag used to determine whether the email is enabled.
+		 *
+		 * @since 2.1.0-dev.1
+		 *
+		 * @param bool $is_enabled whether the email is enabled
+		 * @param Orders $this admin orders instance
+		 */
+		$is_enabled = (bool) apply_filters( 'wc_facebook_commerce_send_woocommerce_emails', $is_enabled, $this );
```
ChaseWiseman, line 172: I think we'll want to scope this filter to only apply if the order is a commerce order and `$is_enabled` was previously true. The name implies that it would only affect commerce orders, but filtering this to `true` would also enable these emails for regular orders, even if they are disabled in settings.

Also, perhaps we can pass the order as an additional filter param?

ChaseWiseman (changes requested):
@adrianostanley looking good! I added some comments around the filter logic, as well as the tests approach.

tests/integration/Admin/OrdersTest.php
adrianostanley: Yeap, sure thing. My bad I didn't refactor that after testing. Done in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1567/commits/2946c43c5f40f341694ad95e38acf953c522b70c.

includes/Admin/Orders.php
adrianostanley: > I think we'll want to scope this filter to only apply if the order is a commerce order and $is_enabled was previously true. The name implies that it would only affect commerce orders, but filtering this to true would also enable these emails for regular orders, even if they are disabled in settings.

Yeah, that makes sense and it's fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1567/commits/3e69335932ad8ee7991ed25415d8360123b12d32 and https://github.com/facebookincubator/facebook-for-woocommerce/pull/1567/commits/7902bd26959a0eb01a7919b6e2d99f2259b78358.

In `maybe_stop_order_email` a more elegant solution IMO would be starting with

```
if ( ! $is_enabled ) {
    return false;
} 
```

However, considering that early return is not a good practice, I decided to make use of a `$is_previously_enabled` var.

> Also, perhaps we can pass the order as an additional filter param?

Done in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1567/commits/6f4284d20de1a5a0408d68a9ed55e49e73387c6a.

ChaseWiseman (approved):

PR #1568, opened 2020-09-18
Add product-categories.js

deivamagalhaes:
# Summary

This PR adds the `product-categories.js` script and enqueues it and it's dependencies.

## QA

1. Open the page to list the product categories
1. Check the Sources
    - [x] The `product-categories.min.js` is loaded
1. Open the page to edit a product category
1. Check the Sources
    - [x] The `product-categories.min.js` is loaded
1. Open another admin page (for instance the Orders page)
1. Check the Sources
    - [x] The `product-categories.min.js` is **not** loaded

## Before merge

### DIFF
--- a/includes/Admin/Product_Categories.php
+++ b/includes/Admin/Product_Categories.php
@@ -52,7 +52,16 @@ class Product_Categories {
 	 */
 	public function enqueue_assets() {
 
-		wp_enqueue_script( 'wc-facebook-product-categories', facebook_for_woocommerce()->get_plugin_url() . '/assets/js/admin/product-categories.min.js', [ 'jquery', 'wc-backbone-modal', 'jquery-blockui', 'jquery-tiptip', 'facebook-for-woocommerce-modal' ], \WC_Facebookcommerce::PLUGIN_VERSION );
+		if ( $this->is_categories_screen() ) {
+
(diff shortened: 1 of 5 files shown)


includes/Admin/Product_Categories.php
```diff
@@ -52,6 +52,16 @@ public function __construct() {
...
 	public function enqueue_assets() {
 
+		if ( $this->is_categories_screen() ) {
+
+			wp_enqueue_script( 'wc-facebook-product-categories', facebook_for_woocommerce()->get_plugin_url() . '/assets/js/admin/product-categories.min.js', [
+					'jquery',
+					'wc-backbone-modal',
+					'jquery-blockui',
+					'jquery-tiptip',
+					'facebook-for-woocommerce-modal',
```
adrianostanley, line 62:
```suggestion
				'jquery',
				'wc-backbone-modal',
				'jquery-blockui',
				'jquery-tiptip',
				'facebook-for-woocommerce-modal',
```
It looks like there are two tabs instead of one for indentation here.

adrianostanley (changes requested):
The code looks good and the QA passes! 🎉

I left only a requested change regarding tabs, but nothing big. Besides that, it's good to be merged.

PR #1569, opened 2020-09-18
Add orders.js

ChaseWiseman:
# Summary

This PR adds the `orders.js` script and enqueues it and it's dependencies.

## Details

Originally meant to load on the orders list page as well, but I realized the modals and other JS handling won't be needed there after all. I scoped this to just the edit screen, and we can expand that scope if needed later.

## QA

1. Go to the edit order page
    - [x] The `orders.min.js` is loaded
    - [x] The `wc_facebook_commerce_orders` object is printed with the expected data
1. Go to any other admin page
    - [x] The `orders.min.js` is **not** loaded

## Before merge


wvega (approved):
Looks good. LGTM!

PR #1571, opened 2020-09-18
Implement the admin product google product category render method

ChaseWiseman:
# Summary

Renders the Google Product Category fields for a product.

## Details

Some extra styling was needed to ensure the select widths would match the existing select fields, and to ensure the new tooltips align.

## UI Changes

- The new fields: https://cloud.skyver.ge/geuznqW2

### Setup

1. Add `add_filter( 'wc_facebook_commerce_is_available', '__return_true' )`
1. Add `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' )`

### Steps

1. Go to edit a product
1. Select the Facebook tab
    - [x] The Google fields are present
1. Select some categories
    - [x] The selects are added as expected
1. Add `_wc_facebook_google_product_category` meta to the product, value `5710`
1. Reload the edit product page
    - [x] The Google fields load pre-selected with `Arts & Entertainment -> Hobbies & Creative Arts` selected

## Before merge

### DIFF
--- a/includes/Admin/Products.php
+++ b/includes/Admin/Products.php
@@ -12,6 +12,8 @@ namespace SkyVerge\WooCommerce\Facebook\Admin;
 
 defined( 'ABSPATH' ) or exit;
 
+use SkyVerge\WooCommerce\Facebook\Products as Products_Handler;
+
 /**
  * General handler for product admin functionality.
  *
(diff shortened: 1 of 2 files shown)


includes/Admin/Products.php
```diff
@@ -88,7 +108,7 @@ public static function render_commerce_fields( \WC_Product $product ) {
 			<input type="checkbox" class="enable-if-sync-enabled"
 			       name="<?php echo esc_attr( self::FIELD_COMMERCE_ENABLED ); ?>"
 			       id="<?php echo esc_attr( self::FIELD_COMMERCE_ENABLED ); ?>" value="yes"
-			       checked="<?php echo \SkyVerge\WooCommerce\Facebook\Products::is_commerce_enabled_for_product( $product ) ? 'checked' : ''; ?>">
+			       checked="<?php echo Products_Handler::is_commerce_enabled_for_product( $product ) ? 'checked' : ''; ?>">
```
unfulvio, line 111: can also use `<?php checked( Products_Handler::is_commerce_enabled_for_product( $product ) ); ?>` here

unfulvio (approved):

includes/Admin/Products.php
wvega: I think we should use `checked()`. The `checked` HTML attribute [is boolean](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked), so if present (regardless of the value) the field will be checked by default.

unfulvio: good catch if echoing should echo `checked="checked"` or nothing at all

PR #1572, opened 2020-09-18
Implement the admin product attributes render method

wvega:
# Summary

This PR implements the method used to render the attribute fields in the Facebook tab for Products.

## Details

I added a couple of helpers to turn the return value of `Products::get_available_product_attributes()` into an array of attribute names (like `pa_color`) and labels (like `Color`), and use that as the default options of the different select fields.

However, I noticed that select2 doesn't show any options when it's configured to load results using Ajax. It seems that all we can do is provide a default selected value.

I kept the helpers because perhaps we can move them and use them in the handler for the Ajax action, but it seems we should populate the options of the select with the already selected value, if any, and leave them empty to show the placeholder text in other cases.

Finally, I added a simple test to check that the name of the field appears in the output. I think it would be better to cover this section with an acceptance test once we implement the changes to save and hide the fields.

## UI Changes

None yet.

## QA

- [x] Code review
- [x] `run integration Admin/ProductsTest.php` pass

## Before merge

### DIFF
--- a/includes/Admin/Products.php
+++ b/includes/Admin/Products.php
@@ -77,6 +77,7 @@ class Products {
 				'female' => __( 'Female', 'facebook-for-woocommerce' ),
 				'male'   => __( 'Male', 'facebook-for-woocommerce' ),
 			],
+			'value'       => FacebookProducts::get_product_gender( $product ),
 		] );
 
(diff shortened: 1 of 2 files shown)


ChaseWiseman (approved):
LGTM!

PR #1574, opened 2020-09-19
Implement the admin product attribute AJAX search

ChaseWiseman:
# Summary

Enables AJAX search for the product attribute selection fields.

## UI Changes

_N/A_

## QA

Visibility, storage, and overall acceptance will be handled in later stories.

### Setup

- A product with more than one attribute

### Steps

1. Try searching for one of the attributes in each of the attribute settings fields
    - [x] Your expected attribute is returned based on the search term

## Before merge

### DIFF
--- a/includes/AJAX.php
+++ b/includes/AJAX.php
@@ -11,6 +11,7 @@
 namespace SkyVerge\WooCommerce\Facebook;
 
 use SkyVerge\WooCommerce\Facebook\Admin\Settings_Screens\Product_Sync;
+use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
 
 defined( 'ABSPATH' ) or exit;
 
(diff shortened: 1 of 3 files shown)


includes/AJAX.php
```diff
@@ -41,6 +46,55 @@ public function __construct() {
...
+
+			if ( ! $product instanceof \WC_Product ) {
+				throw new Framework\SV_WC_Plugin_Exception( 'A valid product ID is required' );
+			}
+
+			/** @var array $attributes */
+			$attributes = Admin\Products::get_available_product_attribute_names( $product );
+
+			// filter out any attributes whose slug or proper name don't at least partially match the search term
+			$results = array_filter( $attributes, function( $name, $slug ) use ( $term ) {
```
unfulvio, line 86:
```suggestion
			$results = array_filter( $attributes, static function( $name, $slug ) use ( $term ) {
```

micro optimization

includes/AJAX.php
```diff
@@ -41,6 +46,55 @@ public function __construct() {
...
+				throw new Framework\SV_WC_Plugin_Exception( 'A search term is required' );
+			}
+
+			$product = wc_get_product( (int) Framework\SV_WC_Helper::get_requested_value( 'request_data' ) );
+
+			if ( ! $product instanceof \WC_Product ) {
+				throw new Framework\SV_WC_Plugin_Exception( 'A valid product ID is required' );
+			}
+
+			/** @var array $attributes */
```
unfulvio, line 82:
```suggestion
```
isn't this redundant? the phptag of the method used here already has `@return array` in the phpdoc

unfulvio (approved):

PR #1576, opened 2020-09-21
Update the admin orders add notices method

adrianostanley:
# Summary

This PR will add an implementation to the `SkyVerge\WooCommerce\Facebook\Admin\Orders:add_notices`.

## UI Changes

An admin notice was added.

image

### Setup

1. Make sure to have at least one Order with a **pending payment** status
2. Take the Order post id and run the following query
```
UPDATE wp_postmeta SET meta_value = 'instagram' WHERE meta_key = '_created_via' AND post_id = {post_id};
```

### Steps

1. Go to WooCommerce > Orders
2. Click on the Order you had setup for this QA
    - [x] A dismissible notice with the following content is added: _This order is currently being held by Instagram and cannot be edited. Once released by Instagram, it will move to **Processing** or **Cancelled** status._
3. Dismiss the notice
4. Refresh the page
    - [x] No notices are shown
5. Repeat the setup for another order and follow steps 1. and 2.
    - [x] The notice for _This order is currently being held by Instagram and cannot be edited. Once released by Instagram, it will move to **Processing** or **Cancelled** status._ is shown

## Before merge


deivamagalhaes:
@adrianostanley , when I refresh the page for the pending order, I do see the same notice:
`This order is currently being held by Instagram and cannot be edited. Once released by Instagram, it will move to Processing or Cancelled status.`

includes/Admin/Orders.php
```diff
@@ -103,7 +103,45 @@ public function enqueue_assets() {
...
+
+		$order  = wc_get_order( $post );
+		$plugin = facebook_for_woocommerce();
+
+		if ( Commerce\Orders::is_order_pending( $order ) ) {
 
+			/* translators: Placeholders: %1$s - <strong> tag, %2$s - </strong>, %3$s - <strong> tag, %4$s - </strong> tag */
+			$message = sprintf(
+				__( 'This order is currently being held by Instagram and cannot be edited. Once released by Instagram, it will move to %1$sProcessing%2$s or %3$sCancelled%4$s status.', 'facebook-for-woocommerce' ),
+				'<strong>', '</strong>',
```
deivamagalhaes, line 120: No need to have this twice, you can use %1$s and %2$s multiple times on the text.

includes/Admin/Orders.php
```diff
@@ -103,7 +103,45 @@ public function enqueue_assets() {
...
+
+		if ( get_transient( 'wc_facebook_bulk_order_update' ) ) {
+
+			/* translators: Placeholders: %1$s - <strong> tag, %2$s - </strong> */
+			$message = sprintf(
+				__( '%1$sHeads up!%2$s Instagram order statuses can’t be updated in bulk. Please update Instagram orders individually so you can provide order details required by Instagram.', 'facebook-for-woocommerce' ),
+				'<strong>', '</strong>'
+			);
+
+			$plugin->get_admin_notice_handler()->add_admin_notice( $message, $plugin::PLUGIN_ID . '_facebook_bulk_order_update', [
```
deivamagalhaes, line 138: I'd probably remove the second "facebook" or replace it with "commerce".

includes/Admin/Orders.php
adrianostanley: Done in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1576/commits/8a78e4d5de7f094320bd47353a7d82e4699d8d70.

includes/Admin/Orders.php
adrianostanley: _Related to my most recent comment._

This code is now part of another PR that was already merged. Since this is a minor change, maybe we can change this ID while working on another `Admin\Order` story.

adrianostanley:
@deivamagalhaes thanks for reviewing. The placeholder requested change is addressed in https://github.com/facebookincubator/facebook-for-woocommerce/commit/8a78e4d5de7f094320bd47353a7d82e4699d8d70.

So, it looks like @unfulvio had implemented the bulk update notice in #1583.

I had to merge that in this PR so the scope of this story as well as its QA steps were reduced.

deivamagalhaes (approved):
I noticed I need to refresh the page twice for the notice to disappear, but I don't think that is a big issue.

adrianostanley:
> I noticed I need to refresh the page twice for the notice to disappear, but I don't think that is a big issue.

@deivamagalhaes maybe we can create a bug fix story as I'm not able to reproduce this. 🤔

PR #1579, opened 2020-09-22
Implement the admin product save fields method

deivamagalhaes:
# Summary

This PRs implements the `Admin\Products::save_commerce_fields()` method.

### Setup

1. Add `add_filter( 'wc_facebook_commerce_is_available', '__return_true' )`
1. Add `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' )`

### Steps

1. Edit a Product, enabling Commerce and configuring the Google product category, the gender and the color/size/pattern attributes
1. Click "Update"
    - [x] The changes are persisted after the page is reloaded
1. Edit the Product again, trying to use the same attribute for two different fields (for instance, size and pattern)
1. Click "Update"
    - [x] The change for the second field to use the already used attribute is not persisted after the page is reloaded
    - [x] The other changes are persisted after the page is reloaded
    - [x] A log entry similar to `There was an error trying to save the product attributes for product` is added to the plugin logs

### Automated tests

- [x] `vendor/bin/codecept run tests/integration/Admin/ProductsTest.php:test_save_commerce_fields` pass

## Before merge


unfulvio (changes requested):
Hi @deivamagalhaes while the integration test passes, I have encountered an issue where the product color, size, pattern attributes are empty or the corresponding update methods throw an unhandled exception for any reason, the page would just turn into a WSOD. Upon page reload I can see that the product has been partially updated. 

It also appears that the `FacebookProducts` class may have been renamed into `\SkyVerge\WooCommerce\Facebook\Products` in the base branch?

unfulvio (approved):

PR #1581, opened 2020-09-22
Implement Admin/Orders::render_refund_reason_field()

unfulvio:
# Summary

This PR implements the `render_refund_reason_field()`

## UI Changes

A dropdown HTML is added to the order page, but it's invisible by default. 

## Open questions

~~I have encountered an issue while implementing this.~~

~~It makes sense that the dropdown options are populated via the `SkyVerge\WooCommerce\Facebook\API\Orders\Refund\Request` constants, however this class is not loaded in `Admin/Orders` class context.~~ **UPDATE** as per @deivamagalhaes [suggestion](https://github.com/facebookincubator/facebook-for-woocommerce/pull/1581#issuecomment-697040009) I moved those constants in `Commerce/Orders`.

~~Furthermore, it is not clear from the engineering doc if the refund reason _label_ will be utilized in the WooCommerce admin UI once it has been saved (as if when the merchant enters a refund reason in the regular text input field for this purpose). So, should we have a method that returns an array of key-values with constants and labels, so we can set consistently a refund reason on the order?~~ Update: clarified by @ChaseWiseman [here](https://github.com/facebookincubator/facebook-for-woocommerce/pull/1581#issuecomment-697055884)

~~For this reason too I'm thinking that maybe the refund constants should be moved out of that class? I'm not really sure of the intention because there's not enough detail in the engineering doc or the story, sorry.~~

### Steps

1. Go to WooCommerce > Orders
2. Click on a Order 
    - [x] A `#wc_facebook_refund_reason` select element with `display:none` is present at the bottom of the page
    - [x] The options are the ones listed in the [pitch doc](https://docs.google.com/document/d/1FgojSVozyO7AnLgoS-qIe1XT0pE78tYEHqnPJFhO_tY/edit#bookmark=id.d94qyloky6yi)
3. [x] The dropdown is not present when visiting other admin pages than a single order edit screen page

## Before merge


includes/Admin/Orders.php
```diff
@@ -140,6 +141,22 @@ public function render_modal_templates() {
 	 */
 	public function render_refund_reason_field() {
 
+		$current_screen = Framework\SV_WC_Helper::get_current_screen();
+
+		if ( ! $current_screen || 'shop_order' !== $current_screen->id ) {
```
deivamagalhaes, line 146: You can use `is_edit_order_screen()` instead.

deivamagalhaes (changes requested):
You're right about this class not being available. I am getting a FATAL on the order edit page:
```
Fatal error: Uncaught Error: Class 'SkyVerge\WooCommerce\Facebook\API\Orders\Refund\Request' not found in /user/Code/facebook-for-woocommerce/includes/Admin/Orders.php on line 152
```

I agree with adding a new method to return these options with their labels, if we think we may use these in other places. Maybe it could live in `Commerce\Orders`?

deivamagalhaes:
@unfulvio , I've added a similar method for the cancellation reasons on https://github.com/facebookincubator/facebook-for-woocommerce/commit/ddae8f0eefb7c7b85f6e7cf169139fad3a1d7aa2.

ChaseWiseman:
@unfulvio 

>Furthermore, it is not clear from the engineering doc if the refund reason label will be utilized in the WooCommerce admin UI once it has been saved (as if when the merchant enters a refund reason in the regular text input field for this purpose)

Nope it won't be used in the UI or saved - the select is only to choose which code to send in the API request.

N-Molham (approved):
Works as expected

PR #1584, opened 2020-09-22
Block the UI elements for Commerce orders

N-Molham:
# Summary

As outlined [here](https://docs.google.com/document/d/11AG_ZVIYGKn96BvEn_Qwy91FOALeE7GmB5cpfN1D6LU/edit#heading=h.3g2gjt89qyl1)

## UI Changes

- If order is Commerce order:
    - remove all statuses from the “Status” select except for Pending, Processing, Completed, Refunded, and Cancelled
    - and the order's status is pending, disable the "Date created" field
    - and the order's status is pending, disable the "Status" field
    - and the order's status is pending, hide the Billing & Shipping “Edit” icons
    - and the order's status is "Cancelled", disable the "Status" field

#### User testing

1. Create a couple of orders
   - [x] An order by WC normal checkout to test if all of the above conditions don't apply
   - [x] Create an order via instagram or facebook stores or manually and change the `_created_via` meta to `instagram` or `facebook`
2. Test each of the conditions mentioned above
    - [x] Only allowed statuses
    - [x] order's status is pending, disable the "Date created" field
    - [x] order's status is pending, disable the "Status" field
    - [x] order's status is pending, hide the Billing & Shipping “Edit” icons
    - [x] order's status is "Cancelled", disable the "Status" field

## Before merge


deivamagalhaes:
Hey @N-Molham , just a tip, we usually put the story title as the PR title. 🙂

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
 
 	'use strict';
 
+	const isCommerceOrder = Boolean( wc_facebook_commerce_orders.is_commerce_order );
+
+	window.WCFacebookCommerceOrderOperations = {
```
deivamagalhaes, line 16: I don't think we need to create a window object for these methods. We usually don't unless we plan on instantiating it, and when we do it's usually described on the discovery doc.

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+
+			$orderStatus.find( 'option' ).each( function ( index, option ) {
+
+				// check if option value in the allowed list or not
+				if ( wc_facebook_commerce_orders.allowed_commerce_statuses.indexOf( option.value ) === -1 ) {
+					// delete/remove option if not allowed
+					option.remove();
+				}
+			} );
+		},
```
deivamagalhaes, line 32:
```suggestion
		},
		
		
```

We usually use the same spacing rules as we use for PHP, so double lines between methods.

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		/**
+		 * Enable or Disable order created fields.
+		 *
+		 * @param {Boolean} enable wither to enable date fields (true) or not (false)
+		 */
+		toggle_created_date_fields_status   : enable => {
+			$( '#order_data' ).find( 'input[name*=order_date]' ).prop( 'disabled', !enable ).toggleClass( 'disabled', !enable );
+		},
```
deivamagalhaes, line 40:
```suggestion
		},
		
		
```

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+				if ( wc_facebook_commerce_orders.allowed_commerce_statuses.indexOf( option.value ) === -1 ) {
+					// delete/remove option if not allowed
+					option.remove();
+				}
+			} );
+		},
+		/**
+		 * Enable or Disable order created fields.
+		 *
+		 * @param {Boolean} enable wither to enable date fields (true) or not (false)
```
deivamagalhaes, line 36:
```suggestion
		 * @param {Boolean} enable whether to enable date fields (true) or not (false)
```

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+			$( '#order_data' ).find( 'input[name*=order_date]' ).prop( 'disabled', !enable ).toggleClass( 'disabled', !enable );
+		},
+		/**
+		 * Disable order status field
+		 *
+		 * @param {Object} $orderStatus Order select jQuery DOM object
+		 */
+		disable_order_status_field          : ( $orderStatus ) => {
+			$orderStatus.prop( 'disabled', true ).addClass( 'disabled' );
+		},
```
deivamagalhaes, line 48:
```suggestion
		},
		
		
```

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		/**
+		 * Toggle customer field
+		 *
+		 * @param {Boolean} hide
+		 */
+		toggle_order_customer_field         : ( hide ) => {
+			$( '#order_data' ).find( '.form-field.wc-customer-user' ).addClass( 'hidden', hide );
```
deivamagalhaes, line 55: I am not familiar with this second param to `addClass`. Will it only add the class if `hide` is true?

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		},
```
deivamagalhaes, line 56:
```suggestion
		},
		
		
```

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		/**
+		 * Toggle customer field
+		 *
+		 * @param {Boolean} hide
+		 */
+		toggle_billing_and_shipping_fields  : ( hide ) => {
+			$( '#order_data' ).find( 'a.edit_address' ).addClass( 'hidden', hide );
```
deivamagalhaes, line 63: Same question about the `addClass` call.

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		},
```
deivamagalhaes, line 64:
```suggestion
		},
		
		
```

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+		/**
+		 * Disable order status field
```
deivamagalhaes, line 66: This description probably needs an update.

includes/Admin/Orders.php
```diff
@@ -88,9 +88,11 @@ public function enqueue_assets() {
 		], \WC_Facebookcommerce::VERSION );
 
 		wp_localize_script( 'wc-facebook-commerce-orders', 'wc_facebook_commerce_orders', [
-			'order_id'          => $order->get_id(),
-			'is_commerce_order' => Commerce\Orders::is_commerce_order( $order ),
-			'shipment_tracking' => $order->get_meta( '_wc_shipment_tracking_items', true ),
+			'order_id'                  => $order->get_id(),
+			'order_status'              => $order->get_status(),
```
deivamagalhaes, line 92: Out of curiosity, why you decided not to use the value of the select on document ready, instead of adding this param?

includes/Admin/Orders.php
```diff
@@ -88,9 +88,11 @@ public function enqueue_assets() {
...
+			'is_commerce_order'         => Commerce\Orders::is_commerce_order( $order ),
+			'shipment_tracking'         => $order->get_meta( '_wc_shipment_tracking_items', true ),
+			'allowed_commerce_statuses' => [ 'wc-pending', 'wc-processing', 'wc-completed', 'wc-refunded', 'wc-cancelled' ],
```
deivamagalhaes, line 95: Also curious as to why you added this param, instead of having this list on the JS.

deivamagalhaes (changes requested):
A few suggestions and a few questions.

Also, we usually use the story title as the PR title.

assets/js/admin/orders.js
```diff
@@ -11,4 +11,91 @@ jQuery( document ).ready( ( $ ) => {
...
+			WCFacebookCommerceOrderOperations.disable_pending_order_related_fields( $orderStatus );
+		}
+
+		if ( 'cancelled' === wc_facebook_commerce_orders.order_status ) {
+			WCFacebookCommerceOrderOperations.disable_order_status_field( $orderStatus );
+		}
+
+		$orderStatus.on( 'change', () => {
+
+			if ( 'wc-pending' === $orderStatus.val() ) {
```
deivamagalhaes, line 95: If we are evaluating the change to `wc-pending` in real time (on the change event), should we not do the same for the `wc-cancelled`?

assets/js/admin/orders.js
N-Molham: Understood, so you recommend removing it from the window scope?

assets/js/admin/orders.js
N-Molham: @deivamagalhaes Updated to what exactly? I see it's pretty descriptive? does it miss something I don't see?

includes/Admin/Orders.php
N-Molham: Well, to execute any related code as soon as the page load without depending on the select input as that might change on page load for any reason by another code from different plugin

includes/Admin/Orders.php
N-Molham: To pipe that into WP Enqueue localization workflow as well as be able to easily change it down the road by `apply_filters` and/or loading the list based on some condition in the backend

assets/js/admin/orders.js
N-Molham: I guess you are right, we don't need the on-change event handler as that would prevent the data saving upon order update ... I will remove it

assets/js/admin/orders.js
deivamagalhaes: Well, the method does more than just update the status field, so the description is not accurate as is.

includes/Admin/Orders.php
deivamagalhaes: Ok, fair enough.

includes/Admin/Orders.php
deivamagalhaes: I do not believe it to be necessary...

assets/js/admin/orders.js
deivamagalhaes: I believe it would be good to double check (with Chase or perhaps Lindsey) if we expect the fields to be blocked on change as well or not.

assets/js/admin/orders.js
```diff
@@ -11,4 +11,94 @@ jQuery( document ).ready( ( $ ) => {
 
 	'use strict';
 
+	const isCommerceOrder = Boolean( wc_facebook_commerce_orders.is_commerce_order );
+
+	const WCFacebookCommerceOrderOperations = {
```
deivamagalhaes, line 16: Why do we need an object at all?

includes/Admin/Orders.php
deivamagalhaes: For this param.

deivamagalhaes:
Can you please fix the conflict as well? And remove the changes to the other minified JS from this PR?

assets/js/admin/orders.js
N-Molham: Ah, I thought you were talking about a different method ... going to update the description now

includes/Admin/Orders.php
N-Molham: I don't have strong feeling about it ... I just thought it would be more efficient and easier to change in the future

assets/js/admin/orders.js
N-Molham: Based on another story which hides the update button meta box if order status is pending, I believe that is the case

assets/js/admin/orders.js
N-Molham: At first I though about exposing it to the `window` scope for later use, but in the same time it's nice to have them grouped for easy to follow and understand ... I don't have strong feeling about it

ChaseWiseman (approved):
conflicts fixed, QA checks out. Note: I noticed the the refund button/status needed to be hidden as well, so I added that: 68c439d38128d0eb5634af6107fe586d0e2669f6 (this wasn't planned in the doc)

PR #1585, opened 2020-09-22
Ch63778/implement cancel order ajax callback

N-Molham:
# Summary

This PR implements and register `AJAX::admin_cancel_order()` as AJAX callback for `AJAX::ACTION_CANCEL_ORDER` action as detailed [here](https://docs.google.com/document/d/11AG_ZVIYGKn96BvEn_Qwy91FOALeE7GmB5cpfN1D6LU/edit#heading=h.1dj1smskr86c)

## UI Changes

None

## QA

TODO write `AJAX_Test::test_admin_cancel_order` Integration test as I am unable to test it without API calls

## Before merge


includes/AJAX.php
```diff
@@ -49,6 +51,58 @@ public function __construct() {
...
+
+			wp_send_json_success( 'cancelled' );
+
+		} catch ( Framework\SV_WC_Plugin_Exception $exception ) {
+
+			if ( $order instanceof \WC_Abstract_Order ) {
+				$order->add_order_note( sprintf( __( 'Could not cancel order. Code: "%s", message: "%s"', 'facebook-for-woocommerce' ), $exception->getCode(), $exception->getMessage() ) );
+			}
+
+			wp_send_json_error( $exception->getMessage() );
```
unfulvio, line 104: the exceptions don't seem having a code, though, and the message is not translatable - if  exceptions messages are meant to be displayed to a user other than a developer they could probably be translated via `__()`

also the `sprintf()` here is missing the `/* translators: Placeholders: */` comment to help translators with the placholders content

N-Molham: @unfulvio understood ... the implementation doc suggested to include the error code as well ... but I agree it doesn't look like any of the errors have codes

N-Molham: @unfulvio see new commits below, They should address your comments

unfulvio (approved):
looks better, I made a couple of edits (check commits below)

PR #1586, opened 2020-09-22
Implement the admin orders is order editable

adrianostanley:
# Summary

This PR will add an implementation to the callback for order editability.

## QA

- [x] `tests/integration/Admin/OrdersTest.php` tests pass
 * _Although this is a very simple PR, my test environment is not working properly, so it's important to run the test above to QA._ 

## Before merge

### DIFF
--- a/includes/Admin/Orders.php
+++ b/includes/Admin/Orders.php
@@ -308,9 +308,13 @@ class Orders {
 	 */
 	public function is_order_editable( $maybe_editable, $order ) {
 
-		$maybe_editable = $maybe_editable && ! Commerce\Orders::is_order_pending( $order );
+		// checks if the order is an instance of WC_Order
+		$is_wc_order = isset( $order ) && $order instanceof \WC_Order;
 
-		return $maybe_editable;
+		// if the order is a WC_Order, determines whether it is pending or not
+		$is_order_pending = $is_wc_order && Commerce\Orders::is_order_pending( $order );
+
+		return $maybe_editable && ! $is_order_pending;
 	}
 
 
--- a/includes/Admin/Orders.php
+++ b/includes/Admin/Orders.php
@@ -306,7 +306,7 @@ class Orders {
 	 * @param \WC_Order $order order object
 	 * @return bool
 	 */
-	public function is_order_editable( $maybe_editable, \WC_Order $order ) {
+	public function is_order_editable( $maybe_editable, $order ) {
 
 		$maybe_editable = $maybe_editable && ! Commerce\Orders::is_order_pending( $order );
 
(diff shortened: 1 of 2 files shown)


includes/Admin/Orders.php
```diff
@@ -308,6 +308,8 @@ public function maybe_stop_order_email( $is_enabled, $order ) {
 	 */
 	public function is_order_editable( $maybe_editable, \WC_Order $order ) {
```
ChaseWiseman, line 309:
```suggestion
	public function is_order_editable( $maybe_editable, $order ) {
```

I realized that this is a direct hook callback, so we can't enforce a type here. I think we'll need to check that `$order instanced \WC_Order` within the method.

ChaseWiseman (changes requested):
looks good! I think we just need to add an instance check in the callback method

includes/Admin/Orders.php
adrianostanley: Good point, @ChaseWiseman.

https://github.com/facebookincubator/facebook-for-woocommerce/pull/1586/commits/d9cd5c36fbbe84143c34c174525836733fb4c507 and https://github.com/facebookincubator/facebook-for-woocommerce/pull/1586/commits/f5b70b88dcb7bafd7b897f317d5ca69fde3d8a26 will fix that.

Note: https://github.com/facebookincubator/facebook-for-woocommerce/pull/1586/commits/f5b70b88dcb7bafd7b897f317d5ca69fde3d8a26 determines whether `$order` is an instance of `\WC_Order` and if it's pending. This method body could be reduced to a single `return` conditional line, however, to make it readable, I've introduced two vars. If a better practice for this project suggests that it should be reduced as much as possible, let me know and I'll submit another commit with that.

includes/Admin/Orders.php
```diff
@@ -306,9 +306,15 @@ public function maybe_stop_order_email( $is_enabled, $order ) {
 	 * @param \WC_Order $order order object
 	 * @return bool
 	 */
-	public function is_order_editable( $maybe_editable, \WC_Order $order ) {
+	public function is_order_editable( $maybe_editable, $order ) {
 
-		return $maybe_editable;
+		// checks if the order is an instance of WC_Order
+		$is_wc_order = isset( $order ) && $order instanceof \WC_Order;
```
ChaseWiseman, line 312: @adrianostanley my only question is the purpose of the `isset()` here - `$order` will always be set (even if null) by virtue of being a known param, no?

adrianostanley: @ChaseWiseman correct. I used an `isset` before calling `Commerce\Orders::is_order_pending( $order )` and then probably didn't pay a good attention when refactoring to check if it's an instance of `\WC_Order`. That should be fixed in 5f823fce1980c900b4d1632736cac3db04476cae.

ChaseWiseman (approved):

PR #1587, opened 2020-09-22
Fix undefined index PHP Notices thrown by WC_Facebook_Product_Feed::prepare_product_for_feed()

adrianostanley:
# Summary

PHP Notices were being printed to the product catalog CSV file when `$product_data` was missing indexes like `gender`, `color`, `size` and `pattern`.

Besides the solution for those fields, this PR also covers the prevention for the other fields as well.

## QA

In order to reproduce the error, checkout `epic/instagram-checkout` and run the same QA steps.

### Setup

1. Make sure you have a few products in your store
2. Make sure your PHP is set to display errors and notices
3. Make sure your WP instance is saving entries to the `debug.log` file

### Steps

1. Checkout branch `epic/instagram-checkout`
2. Go to WooCommerce > Facebook
3. Open your browser's console and run `copy(facebookAdsToolboxConfig.feedPrepared.feedUrl)`
4. Append `&regenerate=1` to the Feed URL
5. Go to the resulting URL
 - [x] The .csv file contains PHP Notices
 - [x] The `debug.log` has a few PHP Notices
6. Checkout branch `ch64484/undefined-index-php-notices-thrown-by-wc`
7. Run the same steps
 - [x] The .csv file is generated
 - [x] The `debug.log` has no PHP Notices

## Before merge

### DIFF
--- a/includes/fbproductfeed.php
+++ b/includes/fbproductfeed.php
@@ -766,10 +766,10 @@ if ( ! class_exists( 'WC_Facebook_Product_Feed' ) ) :
 			) . ',' .
 			'new' . ',' .
 			$product_data['visibility'] . ',' .
-			$product_data['gender'] . ',' .
-			$product_data['color'] . ',' .
-			$product_data['size'] . ',' .
-			$product_data['pattern'] . ',' .
+			static::get_value_from_product_data( $product_data, 'gender' ) . ',' .
+			static::get_value_from_product_data( $product_data, 'color' ) . ',' .
+			static::get_value_from_product_data( $product_data, 'size' ) . ',' .
+			static::get_value_from_product_data( $product_data, 'pattern' ) . ',' .
 			$google_product_category . ',' .
 			$product_data['default_product'] . ',' .
 			$product_data['variant'] . PHP_EOL;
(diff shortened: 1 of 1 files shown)


unfulvio (approved):
@adrianostanley this seems to work correctly

Probably this could be a FW helper in the future? (like we do for posted, requested data...)

On a sidenote, there's a difference between `self::` and `static::` (which I had to refresh my memory on as well...). https://www.leaseweb.com/labs/2014/04/static-versus-self-php/ here (see links at the top) some info. Key difference seems to be when you're in child/parent context. Which doesn't really affect our scope here, but it was a good refresher as we normally default to `self` and should probably be more wary of the intentional usage.

adrianostanley:
> Probably this could be a FW helper in the future? (like we do for posted, requested data...)

Sounds good @unfulvio! I could do that in a helper class or something but for the sake of keeping this PR as a simple bugfix I placed it side by side with a few other inner helpers.

> On a sidenote, there's a difference between `self::` and `static::` (which I had to refresh my memory on as well...). https://www.leaseweb.com/labs/2014/04/static-versus-self-php/ here (see links at the top) some info. Key difference seems to be when you're in child/parent context. Which doesn't really affect our scope here, but it was a good refresher as we normally default to `self` and should probably be more wary of the intentional usage.

To be very honest this was the first time I used `static::` and I just wanted to follow the class standards. By testing, I realized there was no difference, but since the class uses `static` already, I thought it'd be wise to keep it standardize.

PR #1589, opened 2020-09-22
Fix background job to hide virtual products

wvega:
# Summary

This PRs updates the background job handler used to hide virtual products to prevent it from creating duplicate meta data entries.

## Details

The background job doesn't replace existing `fb_visibility` meta entries set to `yes` before or when adding the corresponding meta entries set to `no`. As a result, the number of items to processes never decreases and the job can continue to insert new records in the `postmeta` table indefinitely.

This PR updates the job to insert new rows for products that have no `fb_visibility` set and to update existing rows for products that have `fb_visibility` set to `yes`.

A new background job is also included to remove any duplicates created by the previous version. Whenever it finds a post with duplicate `fb_visibility` meta data, it removes all but the last entry. The last entry should have the most recent value for that meta because:

1. If the plugin hasn't updated visibility for that post, the last entry is the one added by the background job
1. If the plugin updated visibility for that post using `update_post_meta()` all entries would have been updated to have the most recent value
1. If the plugin updated visibility for that post using `$product->update_meta_data()` all duplicate rows would have been removed and the remaining entry would have the most recent value (and the post won't be picked up by the new background job anyway)

## UI Changes

N/A

### Setup

1. Install Facebook for WooCommerce 1.10.2

#### Simple Product

1. Create a Simple product
1. Define a price
1. Make the product virtual
1. Enable Facebook sync

#### Variable Product

1. Create a Variable Product
1. Add an attribute with two terms
1. Create a variation for each attribute
1. Define a price for each variation
1. Make one variation virtual
1. Enable Facebook sync for both variations

#### Set visibility

The next steps will ensure that `fb_visibility` is set to `yes` instead of `1`.

1. Go to the Products page
    1. Use the Hide button to hide the new products from the Catalog
    1. Use the Show button to show the new products again 

#### Reproduce the bug

Create a snapshot of the database so that you can go back to this point to test the solution after you reproduce the bug.

1. Upgrade to Facebook for WooCommerce 2.0.1
1. Go to the Dashboard to trigger the lifecycle events
1. Run `SELECT COUNT(meta_id) FROM wp_postmeta WHERE meta_key = 'fb_visibility'` a couple of times against the website's database: 
    - [ ] The number of `fb_visibility` meta entries is constantly increasing

#### Solution

Once you upgrade to this branch, it may take several seconds or attempts to get the expected results if an old job is still running in the background. First the background job that handles virtual products needs to complete execution and then the background job that removes duplicates will start removing the additional rows.

1. Switch to this branch
1. Go to the Dashboard to trigger the lifecycle events
1. Run the SQL query again
    - [x] The number of `fb_visibility` meta entries stops increasing or suddenly decreases to a normal number
1. Run `SELECT post_id, COUNT(meta_key) entries, MAX(meta_id) last_meta_id FROM wp_postmeta WHERE meta_key = 'fb_visibility' GROUP BY post_id HAVING entries > 1` against the website's database:
    - [x] No results are returned - there are no posts with duplicate `fb_visibility` meta

## Before merge

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Facebook for WooCommerce Changelog ***
 
 2020.nn.nn - version 2.0.3-dev.1
+ * Fix - Remove duplicate visibility meta entries from postmeta table
 
 2020.09.25 - version 2.0.2
  * Tweak - Allow simple and variable products with zero/empty price to sync to Facebook
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -11,6 +11,7 @@
 use SkyVerge\WooCommerce\Facebook\API;
 use SkyVerge\WooCommerce\Facebook\Lifecycle;
 use SkyVerge\WooCommerce\Facebook\Utilities\Background_Handle_Virtual_Products_Variations;
+use SkyVerge\WooCommerce\Facebook\Utilities\Background_Remove_Duplicate_Visibility_Meta;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_4 as Framework;
 
 if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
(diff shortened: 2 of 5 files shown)


N-Molham (approved):
LGTM 🚀

PR #1590, opened 2020-09-23
Implement the Admin\Orders modal messages/button methods

deivamagalhaes:
# Summary

This PRs implements the methods used to generate the modal content for the modals on the edit order page.

## Details 

As already documented on the story, we opted for adding the message and buttons for each modal to the JS params, similarly to how we done for the `Product_Categories` modal. These JS params will be used by `orders.js` when opening the corresponding modals.

I've also moved the cancellation reason constants to the Commerce Orders handler and created a new helper method to return a key-value array of cancellation reason codes and labels.

### Steps

1. Edit an order
1. In your console, open the `wc_facebook_commerce_orders` JS object
    - [x] The `complete_modal_message` param is defined and contains the markup for the complete modal content
    - [x] The `complete_modal_buttons` param is defined and contains the markup for the complete modal buttons
    - [x] The `refund_modal_message` param is defined and contains the markup for the refund modal content (for now it just contains the text above the select, because the select is being added in #1581, as it will be shared by the WC refund form and the refund modal)
    - [x] The `refund_modal_buttons` param is defined and contains the markup for the refund modal buttons
    - [x] The `cancel_modal_message` param is defined and contains the markup for the cancel modal content
    - [x] The `cancel_modal_buttons` param is defined and contains the markup for the cancel modal buttons

### Automated tests

- [x] `vendor/bin/codecept run tests/integration/API/Orders/Cancel/RequestTest.php` tests pass
- [x] `vendor/bin/codecept run tests/integration/APITest.php` tests pass
(no new tests were added, this is just to regression test the cancellation reason refactor)

## Before merge

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -120,6 +120,7 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 				require_once __DIR__ . '/facebook-commerce-messenger-chat.php';
 				require_once __DIR__ . '/includes/Events/Event.php';
 				require_once __DIR__ . '/includes/Commerce.php';
+				require_once __DIR__ . '/includes/Utilities/Shipment.php';
 
 				$this->product_feed            = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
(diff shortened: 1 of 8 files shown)


N-Molham (approved):
Looking good

PR #1607, opened 2020-09-28
Allow Sell on Instagram checkbox to be unchecked

wvega:
# Summary

This PR adds a hidden field fallback for the Sell on Instagram checkbox to ensure a value is posted when merchants uncheck the checkbox.

## Details

The PR also ensures that the checkbox is checked by default when the product is enabled for Commerce only. It was previously always rendered as checked.

## UI Changes

None. Only behavior of the checkbox.

### Setup

Use the following filters to make Commerce available and be able to see the checkbox:

1. Add `add_filter( 'wc_facebook_commerce_is_available', '__return_true' );`
1. Add `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' );`

#### Product not enabled for Commerce

1. Edit a product that is not enabled for Commerce
    - [x] The Sell on Instagram checkbox is unchecked by default
1. Check the Sell on Instagram checkbox
1. Update the product
    - [x] The Sell on Instagram checkbox is checked when the page loads

#### Product enabled for Commerce

1. Edit a product that is enabled for Commerce
    - [x] The Sell on Instagram checkbox is checked
1. Uncheck the Sell on Instagram checkbox
1. Update the product
    - [x] The Sell on Instagram checkbox is unchecked when the page loads

## Before merge

### DIFF
--- a/includes/Admin/Products.php
+++ b/includes/Admin/Products.php
@@ -228,7 +228,7 @@ class Products {
 			<input type="checkbox" class="enable-if-sync-enabled"
 			       name="<?php echo esc_attr( self::FIELD_COMMERCE_ENABLED ); ?>"
 			       id="<?php echo esc_attr( self::FIELD_COMMERCE_ENABLED ); ?>" value="yes"
-			       checked="<?php echo Products_Handler::is_commerce_enabled_for_product( $product ) ? 'checked' : ''; ?>">
+			       <?php checked( Products_Handler::is_commerce_enabled_for_product( $product ) ); ?> />
 		</p>
 
 		<div id="product-not-ready-notice" style="display:none;">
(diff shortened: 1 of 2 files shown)


adrianostanley (approved):
Looks good and QA works.

**Note:** If the product has no price and/or _Inventory/Manage stock?_ is not checked, the checkbox remains unchecked and disabled. I guess this is the normal behavior due to the requisites for a product to be sold on Instagram. but I wanted to point that out in case something needs to be done in this PR. Otherwise, it's good to merge.

wvega:
@adrianostanley yes, that sounds like the expected behavior to me as well. If the product is correctly configured for Commerce, merchants should be able to uncheck or uncheck the checkbox, but if the price is not set, the checkbox will be unchecked and disabled until the product is configured correctly again.

PR #1609, opened 2020-09-28
Limit the allowed order statuses for Commerce orders

ChaseWiseman:
# Summary

Refines the UX a bit to remove status options that would result in failure if attempted.

## Details

API failures would be added as notes as-is, but this removes the potential for that to happen and any resulting confusion. Ensures orders can only make status transitions that are supported by Facebook's Orders API.

## UI Changes

Nothing new added, just some items removed.

## QA

- [x] `acceptance Admin/Orders/CommerceCest:try_order_list_table_actions_removed` tests pass
- [x] `acceptance Admin/Orders/CommerceCest:try_order_preview_actions_removed` tests pass
- [x] `acceptance Admin/Orders/CommerceCest:try_edit_order_statuses` tests pass

## Before merge


includes/Admin/Orders.php
```diff
@@ -107,12 +110,36 @@ public function enqueue_assets() {
 			}, $shipment_tracking );
 		}
 
+		// limit the order status field to statuses that can be handled by Facebook
+		switch ( $order->get_status() ) {
```
unfulvio, line 114: maybe this bit could have been an helper method?

unfulvio (approved):

PR #1610, opened 2020-09-28
Move refund reason field after Order refund items are refreshed

wvega:
# Summary

This PR uses a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to detect when the Order refund items are refreshed and make sure the plugin moves the Facebook refund reason into position.

## Details

Previously, if a refund attempt end up with an error, the Facebook refund reason field was not shown if the merchant attempted to refund the order again.

I tried looking for WooCommerce event that we could use to detect when to move the field again, but [meta-boxes-order.js](https://github.com/woocommerce/woocommerce/blob/73deab28a5535ff049f61c5529e186b8837352f2/assets/js/admin/meta-boxes-order.js#L940-L942) doesn't have one for refunds.

## UI Changes

None.

### Steps

1. Use `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' );` to simulate a Commerce conneciton
1. Set the `created_via` meta of an order to `Facebook` (make sure there is no remote ID set)
1. Set the order status to `complete`
1. Click the Refund button
    - [x] The refund reason dropdown is shown
1. Attempt to refund the order
    - [x] The refund operation fails with `Remote ID for parent order not found.`
1. Click the Refund button
    - [x] The refund reason dropdown is still shown

## Before merge


N-Molham (approved):
LGTM 🎉

N-Molham:
@wvega there is just a merge conflict, I believe it's matter of minifying the `orders.js` file

wvega:
Thank you @N-Molham. You're right, I had to merge the base branch into the PR and update the compiled assets. Conflicts fixed!

PR #1615, opened 2020-09-29
Fetching Google Product Category terms

N-Molham:
# Summary

When adding or editing new product category or product where the GPC field displayed, it shows only one parent term.

## Details

- HTTP GET request to load Google Product Category list timeout is set to short
- Parsing the list using `PHP_EOL` constant which is platform environment dependent causing it not working on Windows OS as Google uses Linux where end of line uses `\n` but on Windows end of line is `\n\r`

## UI Changes

Google Product Category field shows all terms fetched

image

#### Add/Edit Product Category
1. Use `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' );` to simulate a Commerce conneciton
2. Add or edit a product category
    - [x] Default Google product category field shows all terms normally

#### Add/Edit Product 

1. Use `add_filter( 'wc_facebook_commerce_is_connected', '__return_true' );` to simulate a Commerce conneciton
2. Add or edit a simple product
    - [x] Set price
    - [x] Toggle Manage stock on under Inventory tab
    - [x] Toggle Sell on Instagram on under Facebook tab
    - [x] the Google product category field shows all terms normally

## Before merge

## Open Questions
- A question, I noticed that the list get cached for an hour ... shouldn't we increase that for something like a month? the list doesn't get updated that often ... the last update on it was more [than a year ago](https://user-images.githubusercontent.com/3770036/94559115-0a21c280-0261-11eb-940f-bde398965249.png)

### DIFF
--- a/includes/Admin/Google_Product_Category_Field.php
+++ b/includes/Admin/Google_Product_Category_Field.php
@@ -92,7 +92,7 @@ class Google_Product_Category_Field {
 		if ( is_array( $categories_response ) && isset( $categories_response['body'] ) ) {
 
 			$categories_body = $categories_response['body'];
-			$categories_body = explode( PHP_EOL, $categories_body );
+			$categories_body = explode( "\n", $categories_body );
 
 			// format: ID - Top level category > ... > Parent category > Category label
 			// example: 7385 - Animals & Pet Supplies > Pet Supplies > Bird Supplies > Bird Cage Accessories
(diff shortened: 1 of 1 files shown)


N-Molham:
A question, I noticed that the list get cached for an hour ... shouldn't we increase that for something like a month? the list doesn't get updated that often ... the last update on it was more than a year ago

image

ChaseWiseman (approved):

ChaseWiseman:
@N-Molham yeah that's a good idea about the transient - I agree that hourly seems a bit too frequent. I'm thinking more like every week, just to make sure and catch any changes in a somewhat timely manner: cabec91dfb7deb24296cb8734acb34a7fa98eeba

What do you think?

N-Molham:
@ChaseWiseman that also make since. The reason I suggested a month is that the list is parley get changed but one week would do as well just fine, as lone as it doesn't add unnecessarily HTTP call every hour

PR #1624, opened 2020-10-06
Fix a PHP Warning in the feed generation

adrianostanley:
# Summary

This PR will fix a PHP Warning being printed in the generated feed when a variable product is using a deleted term.

### Setup

1. Make sure your PHP is set to show PHP Warnings
2. Go to Products > Attributes and create a Product Attribute with two terms
3. Go to Products > Add New and create a Variable Product
4. Configure the product to use the new attribute for variations (select all terms)
5. Create a variation for each attribute combination
6. Define a price for each variation
7. Publish the product
8. Edit the Product Attribute and delete the two terms

### Steps

1. Go to WooCommerce > Facebook
2. Enter `copy(facebookAdsToolboxConfig.feedPrepared.feedUrl)` in your browser's console
3. Paste the URL in a new tab and append `&regenerate=1`
4. Open the generated file
    - [x] No PHP Warnings are printed in the file
    - [x] No PHP Warnings are printed in the `wp-content/debug.log` file

## Before merge


changelog.txt
```diff
@@ -1,501 +1,502 @@
...
- * Escape HTML tags in product descriptions.
- * Use attribute label name instead of first-letter-capitalized slug.
-
-2017-02-02 - version 0.0.2
-
- * Fix breaking when debug mode is disabled (null check in facebook-settings.js)
+*** Facebook for WooCommerce Changelog ***
+
+2020.nn.nn - version 2.0.4
+ * Fix - Prevent an error during the feed generation when variable products are still using deleted terms
```
adrianostanley, line 4: Not sure why it's considering I've changed the whole file since only line 4 was changed.

Probably PhpStorm messed up with the line breaks or something. If that's a problem, I can remove this file from this PR and add the changelog entry directly into `release-2.0.4`.

N-Molham: Maybe it's the line break is different in your PHPstorm settings ... it happens to me also

N-Molham (approved):

changelog.txt
adrianostanley: @N-Molham Fixed in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1624/commits/4f9267ba2d24f4888834efceea4219843418b4ee. Thanks for reviewing.

Issue #1622, opened 2020-10-06
DB table name "wp_postmeta" hardcoded in file


Gabriellaash:
## 🔎 Isolate the bug

After updating the version of Facebook for WooCommerce to 2.0.3, a large number of errors are generated in error_log located in wp_admin:

WordPress database error Table 'mysqluser.wp_postmeta' doesn't exist for query DELETE FROM wp_postmeta WHERE post_id = 31513 AND meta_key = 'fb_visibility' AND meta_id != 19511  made by do_action('wp_ajax_nopriv_wc_facebook_background_remove_duplicate_visibility_meta'), WP_Hook->do_action, WP_Hook->apply_filters, SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WP_Background_Job_Handler->maybe_handle, SkyVerge\WooCommerce\PluginFramework\v5_5_4\SV_WP_Background_Job_Handler->handle, SkyVerge\WooCommerce\Facebook\Utilities\Background_Remove_Duplicate_Visibility_Meta->process_job, SkyVerge\WooCommerce\Facebook\Utilities\Background_Remove_Duplicate_Visibility_Meta->remove_duplicates

## ✍️ Describe the bug
In version 2.0.3 of plugin Facebook for WooCommerce, database table name "wp_postmeta" is hardcoded in file /includes/Utilities/Background_Remove_Duplicate_Visibility_Meta.php in sql request in line 141:

$sql = "DELETE FROM wp_postmeta WHERE post_id = %d AND meta_key = 'fb_visibility' AND meta_id != %d";

##🚶‍♀️ Steps to reproduce
If you use a different database prefix then wp_, the error will be generated.

simonporter007:
Hey @Gabriellaash!

Thanks for reporting this issue. One of the community has submitted a patch for this issue which our developers have merged into the development branch already! They're working on the next release of the plugin, and this should be included in the 2.0.4 release.

Let me close this off now, but please don't hesitate to reach out again if you're still encountering issues with the new release!

Cheers,

Simon.

PR #1627, opened 2020-10-08
Update Background_Remove_Duplicate_Visibility_Meta.php

Digiover:
Fixes issue https://github.com/facebookincubator/facebook-for-woocommerce/issues/1622 in version 2.0.3 and 2.0.4.


wvega (approved):
@Digiover thank you for fixing that query!

I'm going to merge this and then I'll add a fix for another SQL query that is causing issues and we will release those in 2.0.4 as a quick fix.

PR #1639, opened 2020-10-19
Ensure default variation is carried over to Facebook

N-Molham:
# Summary

This PR will ensure when the default variation changes on the website it syncs (carry over) to Facebook store

### Setup

- Make sure product sync is enabled and all products are synced
- Create a product with two or more variations

### Steps

1. Edit the variable product and change it's default `Default Form Values` to something corresponds to matching variation
    - [ ] Changes get saved normally without error
    - [ ] It reflects on the Shop section in the Facebook page 

image

image

## Before merge


facebook-commerce.php
```diff
@@ -1398,6 +1416,41 @@ function update_product_group( $woo_product ) {
...
+	protected function get_product_group_default_variation( $woo_product ) {
+
+		$default_attributes = $woo_product->woo_product->get_default_attributes( 'edit' );
+
+		if ( empty( $default_attributes ) ) {
+			return null;
+		}
+
+		$default_variation  = null;
+		$product_variations = $woo_product->woo_product->get_available_variations( 'objects' );
```
unfulvio, line 1436: if I am not mistaken in old (yet supported) versions of WooCommerce (see for example WC 3.0.x) this method would only return an array and has no support for `objects` - in this case the foreach loop below would produce a PHP error; we should ensure this not to be a possibility and watch out for types

unfulvio (changes requested):
hey @N-Molham in principle this works but the code needs some changes to account for old WC versions -- thanks

facebook-commerce.php
```diff
@@ -1398,6 +1416,41 @@ function update_product_group( $woo_product ) {
...
 
+
+	/**
+	 * Determine if there matching variation to the default attributes
+	 *
+	 * @param WC_Facebook_Product $woo_product
+	 *
+	 * @return WC_Product_Variation|null
+	 */
+	protected function get_product_group_default_variation( $woo_product ) {
```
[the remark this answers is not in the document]
N-Molham: Yes, because in runtime the product data type passed to the parent calling method `WC_Facebookcommerce_Integration::update_product_group` is by the same data type of `\WC_Facebook_Product` ... so to keep the workflow the same without breaking it, I just continued with the same data type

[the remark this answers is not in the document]
N-Molham: Maybe I just used to make methods like that `protected` by default to avoid changing it down the road in case the the class got extended ... I just need to change this habit

facebook-commerce.php
N-Molham: this 648ea1f should take care of that

unfulvio (approved):
that looks better @N-Molham - given that 2.1.0 is released though (see https://github.com/facebookincubator/facebook-for-woocommerce/releases/tag/2.1.0), we  need to rebase this

N-Molham:
@unfulvio I assume you want to rabase it on 2.1.1 release, right?

wvega:
@N-Molham we deployed 2.1.1 not too long ago to address high impact bug, so we will need to target this for another release. 

We are trying to get all the changes from 2.1.0 and 2.1.1 merged into master so that we can have clean release branches again, so I would say we wait until #1655 is merged before we merging this one into a new release branch (which doesn't exist yet).

wvega:
@unfulvio @N-Molham I just rebased this branch on top of #1658. For reference, the previous set of commits was https://github.com/facebookincubator/facebook-for-woocommerce/compare/c88c9e76...8a937163

facebook-commerce.php
```diff
@@ -1450,6 +1468,63 @@ function update_product_group( $woo_product ) {
 		*/
 	}
 
+
+	/**
+	 * Determines if there a matching variation for the default attributes.
```
wvega, line 1473:
```suggestion
	 * Determines if there is a matching variation for the default attributes.
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Facebook for WooCommerce Changelog ***
 
 2020.nn.nn - version 2.1.2
+ * Tweak - Ensure default variation is carried over to Facebook
```
wvega, line 4: Since the product item ID of the default variation is not available on initial sync (because each variation will be synced asynchronously using the Batch API), I think we should update the changelog entry to indicate when can merchants expect to see the default variation reflected on their catalogs.

Perhaps something like this (should update `readme.txt` as well):

```suggestion
 * Tweak - Default variation selection will be synced to Facebook if the default product variation is already synced 

```

facebook-commerce.php
```diff
@@ -1428,9 +1430,25 @@ function update_product_group( $woo_product ) {
...
-		$product_group_data = array(
+		// figure out the matching default variation
+		$default_product_fbid  = null;
+		$woo_default_variation = $this->get_product_group_default_variation( $woo_product );
+
+		if ( $woo_default_variation ) {
+			$default_product_fbid = $this->get_product_fbid(
+				self::FB_PRODUCT_ITEM_ID,
+				$woo_default_variation['variation_id'],
+				$woo_default_variation
```
wvega, line 1441: @N-Molham looks like `$woo_default_variation` is an array now but `get_product_fbid()` expects a `WC_Facebook_Product` in the third argument. Perhaps we don't need to pass the third argument and should allow `get_product_fbid()` to create the instance instead?

```suggestion
				$woo_default_variation['variation_id']
```

wvega (changes requested):
@N-Molham Changes are looking good!

I took a look at this one trying to prepare the 2.1.2 patch release and left a couple of suggestions to update the changelog entry and fix the parameters to `get_product_fbid()`.

facebook-commerce.php
N-Molham: Yeah, you are right ... the `get_product_fbid()` take care of that if the 3rd argument is `null`

N-Molham:
@wvega Took care of your comments in 13d3f5d & 0119d3b

wvega (approved):
@N-Molham changes look good. Thank you for the updates. I'll merge with a final tweak to update the changelog entry in the `readme.txt` file as well: b2f3097

PR #1641, opened 2020-10-23
Partial fix for search events not firing for product search requests that have a single result

wvega:
# Summary

This PR updates the plugin to inject search events on the product page if the customer was redirected to that page after performing a product searching that ended up with a single result.

## Details

The frontend search event doesn't fire because WooCommerce redirects customers to the product page if the search returns a single result:

https://github.com/woocommerce/woocommerce/blob/1427a39be142da796856b75a1e2a32b913236ad4/includes/wc-template-functions.php#L57-L65

In most cases we can't determine whether a product page visit is the result of a single search result. However, if the customer already has a session (for example after adding one product to the cart), then we can store the event data on the session before the redirect occurs and inject it when the product page loads. That's the approach used in this PR.

Please note that the plugin currently sends a server search event even when there is only one result. That means those events are not completely ignored for customers that have no session. We can't send a pixel event in that scenario, but the event is still tracked.

### Setup

- Activate and configure Facebook for WooCommerce

#### Multiple results search

1. Perform a product search that returns multiple results
    - [x] The Pixel helper reports a Search event

#### Single result (logged in)

1. Perform a product search that returns a single result
    - [x] You are redirected to the product page
    - [x] The Pixel helper reports a Search event
1. Reload the page
    - [x] No search event is reported 
 
#### Single result (logged out)

1. Perform a product search that returns a single result
    - [x] You are redirected to the product page
    - [ ] No search event is reported

#### Tests

## Before merge

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -66,6 +66,7 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 				'pre_get_posts',
 				array( $this, 'inject_search_event' )
 			);
+			add_filter( 'woocommerce_redirect_single_search_result', [ $this, 'maybe_add_product_search_event_to_session' ] );
 
 			// AddToCart events
 			add_action( 'woocommerce_add_to_cart', [ $this, 'inject_add_to_cart_event' ], 40, 4 );
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -253,6 +253,29 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 
 		/**
+		 * Attempts to create an Event instance for a product search event using session data.
+		 *
+		 * @since 2.0.6-dev.1
+		 *
+		 * @return Event|null
+		 */
+		private function get_product_search_event_from_session() {
+
+			if ( ! isset( WC()->session ) || ! is_callable( [ WC()->session, 'get' ] ) ) {
+				return null;
+			}
+
+			$data = WC()->session->get( $this->search_event_data_session_variable );
+
+			if ( ! is_array( $data ) || empty( $data ) ) {
+				return null;
+			}
+
+			return new Event( $data );
+		}
+
+
+		/**
 		 * Triggers Search for result pages (deduped)
 		 *
 		 * @internal
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -276,6 +276,21 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 
 		/**
+		 * Deletes a session variable.
+		 *
+		 * @since 2.0.6-dev.1
+		 *
+		 * @param string $key name of the variable to delete
+		 */
+		private function delete_session_data( $key ) {
+
+			if ( isset( WC()->session->$key ) ) {
+				unset( WC()->session->$key );
+			}
+		}
+
+
+		/**
 		 * Triggers Search for result pages (deduped)
 		 *
 		 * @internal
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -27,6 +27,10 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		const FB_PRIORITY_HIGH    = 2;
 		const FB_PRIORITY_LOW     = 11;
 
+
+		/** @var string name of the session variable used to store search event data */
+		private $search_event_data_session_variable = 'wc_facebook_search_event_data';
+
 		/** @var Event search event instance */
 		private $search_event;
 		/** @var array with events tracked */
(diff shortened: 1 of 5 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -187,6 +193,128 @@ function( $item ) {
...
+		 * @internal
+		 *
+		 * @since 2.0.6-dev.1
+		 *
+		 * @param bool $redirect whether to redirect to the product page
+		 * @return bool
+		 */
+		public function maybe_add_product_search_event_to_session( $redirect ) {
+
+			if ( $redirect && $this->search_event && $this->is_single_search_result() ) {
```
N-Molham, line 216: I think here it needs checking if user logged in or not as WC session gets created anyways if user logged in or not
```php
if ( $redirect && is_user_logged_in() && $this->search_event && $this->is_single_search_result() ) {
```
That way the search event won't be reported if user not logged it which will pass the remaining QA Test

facebook-commerce-events-tracker.php
```diff
@@ -187,6 +193,128 @@ function( $item ) {
...
+		 * Determines whether the current request is a product search with a single result.
+		 *
+		 * @since 2.0.6-dev.1
+		 *
+		 * @return bool
+		 */
+		private function is_single_search_result() {
+			global $wp_query;
+
+			return is_search() && is_post_type_archive( 'product' ) && 1 === absint( $wp_query->found_posts );
```
N-Molham, line 235: I believe if we re-ordered the conditions check would perform better and use less memory
```php
return is_search() && 1 === absint( $wp_query->found_posts ) && is_post_type_archive( 'product' );
```
That way it bail as long as the search result has a single post before checking if it's for a product post type or not. what do you think?

N-Molham (commented):
Looking good @wvega, almost all test passes and the integration test fully passed ... I have a couple of comments to highlight few things to get your feedback on them

facebook-commerce-events-tracker.php
wvega: @N-Molham thank you for reviewing!

If I understand correctly, you are saying that the event was reported when you tested while being logged out, is that right?

Do you remember if you had products in your cart when you ran the test? The session cookie is also added when you add a product to the cart.

In this case having the event sent for a logged out user is good. The ideal solution would be to send the event in all cases.

I used "(logged out)" in the test case description to highlight the behavior of the solution in scenario in which there is no customer session available and we can't pass the event data between requests. I should have made that intention explicit.

Right now, if the user is logged out but a session is available (because the customer added a product to the cart or other reason) we want to use that session and report the event.

facebook-commerce-events-tracker.php
wvega: That makes sense. Thanks! Fixed in b4c05624

facebook-commerce-events-tracker.php
N-Molham: @wvega 

> If I understand correctly, you are saying that the event was reported when you tested while being logged out, is that right?

Yes, correct

> Do you remember if you had products in your cart when you ran the test? The session cookie is also added when you add a product to the cart.

With logged out state, I tried a different browser state (incognito mode on Chrome)

> Right now, if the user is logged out but a session is available (because the customer added a product to the cart or other reason) we want to use that session and report the event.

To be clear, when I tested the logged out state, I didn't add a product to cart before trying the search.

> I used "(logged out)" in the test case description to highlight the behavior of the solution in scenario in which there is no customer session available and we can't pass the event data between requests. I should have made that intention explicit.

I think either ways, it's good that it triggers the event even if user not logged in ... right?

wvega: Yes, I think is good that it triggers the event even if the user is not logged in. 

I'm curious to find out why it was sent for you even though you were logged out and hadn't added a product to the cart. **We don't want** to cause WooCommerce to create a customer session if one doesn't already exists. Hosting providers usually disable cache when they see the session cookie, so we should be careful no to create it for this workaround.

I'll confirm that we are not creating a new session. Everything else seems ok based on my previous tests and your review results.

N-Molham: I think that is different from hosting env to another, Maybe something in my server env triggers WooCommerce to create a session automatically upon the first visit ... let me try to debug it and see why and when a new session get created on my env

wvega:
I just rebased this branch on top of #1658. The previous set of commits was https://github.com/facebookincubator/facebook-for-woocommerce/compare/90b28a64...b4c05624.

wvega:
@N-Molham I can confirm on my end that while I'm logged out, doing a product search that returns a single result doesn't automatically create a customer session. I don't get the event tracked in that scenario. If I add a product to the cart then the event is tracked, but that's expected.

Were you able to check whether something on your end creates a customer session while logged out?

When you log out (or in an incognito window) you shouldn't see a `wp_woocommerce_session_*` cookie set. You shouldn't  see the cookie from [this screenshot](https://cloud.skyver.ge/YEuyelov) unless you add a product to the cart or otherwise trigger WooCommerce to create it.

N-Molham:
@wvega to be honest I don't know why it's doing that with me ... I don't see the cookie in my session but for some reason it still fires the search event, I double checked once again

image

wvega:
@N-Molham thank you for checking again. I also don't know what could be causing it. When I search for "Cap" on my test store, WooCommerce redirects me to the Cap product page and this is what I see on the Pixel Helper: https://cloud.skyver.ge/eDuPr8Nj. The only events reported are the `PageView` and `ViewContent`.

N-Molham:
@wvega I don't believe this would be an issue and we can move forward with it ... Maybe @lgfogle can confirm that as this issue depend on the hosting env

wvega:
@N-Molham I don't think the plugin is creating the session, which was my concern, so I also believe it shouldn't be an issue.

The part of this issue that varies with the hosting is how they respond to the presence of the WooCommerce session cookie but we ran into that because it seemed that the session was being created when you performed a search as a logged out user. However, the session cookie doesn't show up on that screenshot so the unexpected result seems unrelated.

I updated the condition to explicitly check whether the customer has an active session before trying to add the event data. The results are the same for me, but may help with the results on your end: 55cb6992

Other than that, if you agree, I think we can merge.

N-Molham:
@wvega Yeah, I agree, adding the `has_session` check will differently help ... I think it's good to merge

wvega:
Approved in this [comment](https://github.com/facebookincubator/facebook-for-woocommerce/pull/1641#issuecomment-717519962). Merging.

PR #1645, opened 2020-10-26
LWI Ads support

unfulvio:
## Summary

This is the base branch to hold the LWI Ads support in Facebook for WooCommerce.

Currently branched out from #1618 (2.1.0) but could be rebased as necessary.

The FW has been bumped to 5.9.0 to make use in some dependent PRs of some helper methods added in newer versions.

### Stories

TBA

### Final story

All stories in [Epic](https://app.clubhouse.io/skyverge/epic/66229/facebook-for-woocommerce-lwi-ads).

## Before merge

- [ ] I have confirmed the final stories from [Epic](https://app.clubhouse.io/skyverge/epic/66229/facebook-for-woocommerce-lwi-ads) have been merged in this branch
- [ ] The appropriate base branch is being used for releasing the feature

PR #1646, opened 2020-10-26
Move the settings page under "Marketing" menu item

unfulvio:
## Summary 

Moves the plugin settings page under "Marketing".

It also displays the WC Admin top bar in the settings page.

#### Stories: 

- [CH 66402](https://app.clubhouse.io/skyverge/story/66402)
- [CH 66409](https://app.clubhouse.io/skyverge/story/66409)

**Release branch: #1645** 

## Details

Since Marketing may not be available in all WC versions, it only adds this changes from WC 4.0 onward. 

I did not find a way to display the full breadcrumbs in the top bar, but this might be a limitation when enabling WC Admin support for non-REACT pages: https://github.com/woocommerce/woocommerce-admin/blob/main/docs/page-controller.md

## UI Changes

- Updated menu item position
- WC Admin top bar

image

### Setup

Install & activate the plugin.

### Steps

- Visit admin in WC 4.0 and above
   - [x] The Facebook page is listed under `Marketing`
   - [x] The WC Admin top bar is visible
- Visit admin in WC below 4.0
   - [x] The Facebook page is listed under `WooCommerce`
   - [ ] The WC Admin top bar is not visible (unless you have WC Admin feature plugin enabled)
   - [ ] You can click on the WC Admin top bar inbox and this opens correctly
   - [x] No JS conflicts or errors in console
- Both cases
   - [x] The page identifier or admin URL to get to the page hasn't changed

## Before merge


N-Molham (commented):
Almost all test pass except for these 2:
- The WC Admin top bar is not visible (unless you have WC Admin feature plugin enabled)
- You can click on the WC Admin top bar inbox and this opens correctly

The WC Admin bar doesn't show up in my case for some reason, I have WC Admin (latest version) and WC 3.9.3

image

Note: When I switched to this PR branch I got fatal error as this branch uses a different framework version than the branch I was on, so I had to run `composer install`

unfulvio:
closing after the rebase

PR #1663, opened 2020-10-28
Release: 2.2.0

unfulvio:
# Summary

The branch for the v2.2.0 release.

The FW has been bumped to 5.10.0 to make use in some dependent PRs of some helper methods added in newer versions.

### Stories

_This branch should contain the LWI Ads feature._

### Final story

All stories in [Epic](https://app.clubhouse.io/skyverge/epic/66229/facebook-for-woocommerce-lwi-ads).

## Before deploy

- [x] I have confirmed that all stories in epic are merged and acceptance is met.

## Deploy

- Ensure you have no untracked files or uncommitted changes on your working directory
- Use `npx sake deploy --deployAssets=0` to skip the `shell:svn_commit_assets` action

## After deploy

If you deploy from this branch, before merging the changes into master, follow these steps to fix the tag that `saké` creates:

- `git fetch` to get the latest tags
- `git tag -d {version}` to delete the tag that `sake` created
- `git tag {version} {hash of Facebook for WooCommerce versioning commit}`
- `git push --tags -f` to overwrite the tag on GitHub with the modified tag

PR #1665, opened 2020-10-28
Add Advertise screen in settings page

unfulvio:
## Summary 

Adds a new tab in the settings page to manage ads with the LWI Ads script.

#### Stories: 

- [CH 66402](https://app.clubhouse.io/skyverge/story/66419/load-fb-lwi-ads-script-in-connect-page)

**Release branch: #1663** 

## UI Changes

A new tab is added in the settings screen:

Screenshot from 2020-10-28 16-45-16

By default when not connected, it will show a message to prompt the merchant to connect with FB first.

When connected, it will inject the LWI Ads script and two iframes will populate as such

Screenshot from 2020-10-28 17-23-05

(note this may display as Create Ad - I was having a connection issue when I took this screenshot)

When clicking on the button to create ads a popup will open to handle ads.

Meanwhile, in the main connection screen, the Create Ad button is removed once connected:

Screenshot from 2020-10-28 17-16-52

### Setup

- Install & activate the plugin
- Connect with FB

### Steps

- Visit the Facebook for WooCommerce settings page
   - [x] There is no more a `Create Ad` button in the FB box
   - [x] There is a new `Advertise` tab
- Click on `Advertise`
   - [x] If connected, it will load 2 iframes to create and manage ads
   - [x] If not connected with FB, it will load a message to prompt to connect
       - [x] Following the link will redirect you to the connect tab screen
   - [x] If you click on the button to create ads from the iframe, a popup will open

## Before merge


includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -0,0 +1,180 @@
...
+	 */
+	public function __construct() {
+
+		$this->id    = self::ID;
+		$this->label = __( 'Advertise', 'facebook-for-woocommerce' );
+		$this->title = __( 'Advertise', 'facebook-for-woocommerce' );
+
+		add_action( 'admin_head', [ $this, 'output_scripts' ] );
+
+		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_assets' ] );
```
N-Molham, line 41: Maybe it's a good idea to split these 2 lines into their own method, ex: `init_hooks` and call it in the constructor ... to avoid adding logic directly in the constructor

N-Molham (commented):
For some reason, the 2 iframes don't load for me

image

I tried to debug it but I don't see something wrong in the PR code, it's something caused by FB js itself ... I also tried US VPN and got the same problem.

I also have a suggestion below

includes/Admin/Settings_Screens/Advertise.php
unfulvio: ok

unfulvio:
@N-Molham what do you see in console?
I am assuming you are successfully connected with FB?
I wonder how we can improve error catching, since the iframes are entirely handled by FB remote script, we only check if you're connected or not

includes/Admin/Settings_Screens/Advertise.php
unfulvio: see https://github.com/facebookincubator/facebook-for-woocommerce/pull/1665/commits/68fa2477e13a5f6e2cee13c5941ca9abbba0ee49

N-Molham:
@unfulvio there are no errors in the console ... it's the first place I checked ... I will do some additional debugging from my end as well to see what is causing this as other merchants might experience the that problem down the road

N-Molham (changes requested):
I was finally able to figure out the reason for the weird behavior I got ... It is due to the timezone setup of the WP Instance. At first it was Manual UTC offset (UTC+/-) ... when I reviewed the [install instruction file](https://drive.google.com/drive/u/2/folders/1khLTONg0v9gnIO4n-dMHmYv9MjWnIuGV) I noticed that the timezone is different, so I changed the tomezone settings to something similar (America/Los Angeles and Africa/Cairo) it worked fine!

So I think we need to make sure the passed timezone for the `data-fbe-extras` is not UTC based ... it must be a timezone name https://www.php.net/manual/en/timezones.php ... so we need to handle the situation that if the setup has `UTC+/-` to convert it to name based.

Update: I am working on a fix for it now

N-Molham (approved):

N-Molham:
Pushed a commit which should take care of the timezone offset to name conversion 01b2c09

adrianostanley:
> @N-Molham what do you see in console?
> I am assuming you are successfully connected with FB?
> I wonder how we can improve error catching, since the iframes are entirely handled by FB remote script, we only check if you're connected or not
> 
> I'll ask others to have a peek too

I can see both iframes. They take a second to load and I see what @N-Molham posted, but it's working here.

DvtcnlIU6D

unfulvio:
@adrianostanley @N-Molham I have tried to improve the UX of the iframe widget when in a loading state: see https://skyverge.slack.com/archives/CR8VDB4G7/p1604025590199400

adrianostanley:
> @adrianostanley @N-Molham I have tried to improve the UX of the iframe widget when in a loading state: see https://skyverge.slack.com/archives/CR8VDB4G7/p1604025590199400

@unfulvio yeah, it's working here and I can see both frames.

deivamagalhaes:
I've made some changes in https://github.com/facebookincubator/facebook-for-woocommerce/pull/1665/commits/1c7dd184770cdefa34189d71ddf062d673b6ee23 to try and have a full HTML error/loading message. Can someone with a working FB connection test that and see if it displays correctly when the iframe is blocked (and is hidden when the iframe loads properly)?

includes/Admin/Abstract_Settings_Screen.php
```diff
@@ -83,14 +83,33 @@ public function render() {
...
+	 *
+	 * @return bool
+	 */
+	protected function is_current_screen_page() {
+
+		if ( Settings::PAGE_ID !== Framework\SV_WC_Helper::get_requested_value( 'page' ) ) {
+			return false;
+		}
+
+		$tab = Framework\SV_WC_Helper::get_requested_value( 'tab' );
```
wvega, line 107: We need to offer a default here, because the link under the Marketing menu doesn't specify a `tab` parameter:

```suggestion
		// assume we are on the Connection tab by default because the link under Marketing doesn't include the tab query arg
		$tab = Framework\SV_WC_Helper::get_requested_value( 'tab', 'connection' );
```

Without the default the CSS stylesheet is not enqueued when you go to the Facebook for WooCommerce settings page clicking Marketing > Facebook.

includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -0,0 +1,254 @@
...
+	/*
+	 * Converts the given timezone string to a name if needed.
+	 *
+	 * @since 2.2.0-dev.1
+	 *
+	 * @param string $timezone_string Timezone string
+	 * @param int|float $timezone_offset Timezone offset
+	 * @return string timezone string
+	 */
+	private function parse_timezone( $timezone_string, $timezone_offset = 0 ) {
```
wvega, line 145: No need to change it now, but I think the purpose of this method is the same as the `get_timezone_string()` helper in the Connection handler:

https://github.com/facebookincubator/facebook-for-woocommerce/blob/c1b264f0d685e930be91502d111acef374919f70/includes/Handlers/Connection.php#L846-L865

wvega (commented):
@deivamagalhaes unfortunately the first iframe doesn't load for me if we use HTML to show the error message. This is what I get in that case:

https://cloud.skyver.ge/E0urJLGj

The iframe loads if I switch to 1cc5857.

However, I then run into another problem. Clicking the **Get Started** button or the _"You can also view your ads on Facebook"_ link on the second iframe opens a pop up with an error: https://cloud.skyver.ge/NQu1BprY and https://cloud.skyver.ge/RBuWkJOg

includes/Admin/Settings_Screens/Advertise.php
asseti6: @wvega yeah, no need to worry about that one on this PR though

deivamagalhaes:
@wvega , can you please revert my changes after https://github.com/facebookincubator/facebook-for-woocommerce/commit/1cc5857c6451b80a0dfeda90dbacf0ef9cdc0ad0 then?

wvega (approved):
I think we can merge these changes! As far as I can tell, the errors I had during testing are either related to my Facebook's account capabilities or an external issue. I'm not aware of any necessary modifications to the implementation of the initialization code.

These are the errors I saw:

On a fresh WordPress installation I can see the iframes on the Advertise tab with the Create Ad and Manage buttons: https://cloud.skyver.ge/04uORdw0

Clicking the Create Ad button shows: https://cloud.skyver.ge/mXu6zvX0
Clicking the Manage button show a screen saying that the content is not available right now: https://cloud.skyver.ge/nOunJQGW

Once I see that message after clicking the Manage button, the Create Ad button stops working. The popup now shows an error instead: https://cloud.skyver.ge/mXu6zvwg

The next time we try to load the iframes, they render slightly different content. There is Get Started button instead of a Create Ad button and there is no Manage button: https://cloud.skyver.ge/jkuD869R

Clicking the Get Started button or the "You can also view your ads on Facebook" link on the second iframe opens a popup with the error above: https://cloud.skyver.ge/6quPXwd4 and https://cloud.skyver.ge/L1uQyynG.

PR #1667, opened 2020-10-28
Move locale helpers to own handler

unfulvio:
## Summary

Moves some locale function to its own `Locale` handler so that both Messenger and the LWI SDK feature can use it.

### UI Changes

None directly, but FB LWI SDK will possibly load the LWI box according to the merchant's locale, for example (Italian):

image

### Setup

- Install, activate & connect to Facebook

### Steps

1. Go to the settings page > Advertise
    - [x] If you use `en_US` as your WordPress locale, the FB Advertise
1. Change your locale in WordPress > Settings > General (or your current user locale in Profile)
1. Go back to the FB settings page > Advertise
    - [x] The locale used now matches your merchant's / local install
1. [Select an unsupported locale](https://github.com/facebookincubator/facebook-for-woocommerce/blob/57276958c7dd43acc4b650d2d9f44aa1286c464f/includes/Locales.php#L27-L131)
    - [x] The box will default to english

### Tests

- [x] `vendor/bin/codecept run integration LocalesTest.php` passes

## Before merge

- [x]  I have confirmed these changes in each supported minor WooCommerce version


N-Molham:
I am experiencing the iframes not loading as #1665 which I am debugging to be able to review this one ... I re-added the other folks as well

adrianostanley:
> I am experiencing the iframes not loading as #1665 which I am debugging to be able to review this one ... I re-added the other folks as well

@unfulvio it's weird I could see the Advertise tab for #1665 but for some reason in this branch I can't.

I also tried to run it in an incognito window to make sure no caches or cookies were causing the issue.

image

includes/Locale.php
```diff
@@ -0,0 +1,214 @@
...
+	 * Determines if a locale is supported by Facebook.
+	 *
+	 * @since 2.2.0-dev.1
+	 *
+	 * @param string $locale a locale identifier
+	 * @return bool
+	 */
+	public static function is_supported_locale( $locale ) {
+
+		return in_array( $locale, self::get_supported_locales( false ), false );
```
wvega, line 210: @unfulvio I think adding a `$with_labels` boolean parameter decreases readability for the code that uses `get_supported_locales()` and increases the responsibility of the method. Considering that we are only using the boolean attribute in `is_supported_locale()`, what do you think about removing that parameter and moving the call to `array_keys()` here:

```suggestion
		return in_array( $locale, array_keys( self::get_supported_locales() ), false );
```

Also, what's the reason to use `$strict = false` explicitly here?

wvega (changes requested):
@unfulvio QA and tests pass. The code also looks good. Good idea moving that logic into its own object, and I like the final names for the methods and class!

This works for me, but I think we don't need to add a boolean parameter to `get_supported_locales()`. I left a suggestion below.

includes/Locale.php
unfulvio: none really, used to be array_key_exists and I suppose we can restore to that given your suggestion to ignore a bool parameter in the called method

unfulvio: https://github.com/facebookincubator/facebook-for-woocommerce/pull/1667/commits/538ca2c0c8d92ab7b1e68068590222e3b0955ca9

unfulvio:
@wvega thanks for reviewing - I've adopted the suggested changes

@adrianostanley probably you encountered that issue because of this https://github.com/facebookincubator/facebook-for-woocommerce/pull/1665#pullrequestreview-518677132 that was fixed in the other branch, but this one didn't pick up the change yet

wvega:
Thank you for the updates @unfulvio!

PR #1677, opened 2020-10-30
Release 2.1.4

maxdupenois:
# Summary

The branch for the v2.1.4 release. 

## Deploy

- Ensure you have no untracked files or uncommitted changes on your working directory
- Use `npx sake deploy --deployAssets=0` to skip the `shell:svn_commit_assets` action

## After deploy

If you deploy from this branch, before merging the changes into master, follow these steps to fix the tag that `saké` creates:

- `git fetch` to get the latest tags
- `git tag -d {version}` to delete the tag that `sake` created
- `git tag {version} {hash of Facebook for WooCommerce versioning commit}`
- `git push --tags -f` to overwrite the tag on GitHub with the modified tag


wvega:
@maxdupenois thank you for putting this branch together!

Changes look good. However, I noticed that there are other places using the `get_available_product_attributes()` and expecting the return type be to an array of `WC_Product_Attribute` objects. I think it would be good to fix `get_available_product_attributes()` to return `WC_Product_Attribute` for variations too to avoid any other errors down the road.

What do you think?

From the changes here, it seems that when the product is a variation, `get_available_product_attributes()` should get the attributes from the parent instead.

maxdupenois:
> @maxdupenois thank you for putting this branch together!
> 
> Changes look good. However, I noticed that there are other places using the `get_available_product_attributes()` and expecting the return type be to an array of `WC_Product_Attribute` objects. I think it would be good to fix `get_available_product_attributes()` to return `WC_Product_Attribute` for variations too to avoid any other errors down the road.
> 
> What do you think?
> 
> From the changes here, it seems that when the product is a variation, `get_available_product_attributes()` should get the attributes from the parent instead.

Hey @wvega, potentially but, for this use case at least we want to get the variant attributes if they are there. For example, if the variant attribute is color, then we want to send the variant version of the color rather whatever may be present on the parent.

For other cases your probably right, maybe there should be two functions, one that gets from the parent and one that gets from the current variant to make this clear?

wvega:
Hey @maxdupenois, I think having two methods is a good idea. I see them this way:

Other parts of the code expect `get_available_product_attributes()` to return a list of `WC_Product_Attribute`, in that case I still think we should update the method to call `get_attributes()` on the product if **not** a variation, and on the parent product if the product is a variation. I think other uses of that method are more concerned about the name of the attributes that are available, instead of the values, so checking the parent works.

The other method would be the one we use to get the attribute _values_ and in that case I think it only makes sense to call `get_attributes()` on the variation. On my tests, calling `get_attribute( $attribute_name )` on a simple product or a parent product returns a comma separated list of attribute values instead of a specific value.

includes/Products.php
```diff
@@ -1087,6 +1087,10 @@ public static function get_product_pattern( \WC_Product $product ) {
 	 * @return \WC_Product_Attribute[]
 	 */
 	public static function get_available_product_attributes( \WC_Product $product ) {
+		if ( $product->is_type( 'variation' ) ) {
+			$parent_product = wc_get_product( $product->get_parent_id() );
+			return $parent_product instanceof \WC_Product ? self::get_available_product_attributes( $product ) : array();
```
wvega, line 1092: I think you meant to use `$parent_product` here:

````suggestion
			return $parent_product instanceof \WC_Product ? self::get_available_product_attributes( $parent_product ) : array();
```

wvega (approved):
@maxdupenois thank you for the updates. They work great! 

I left a tiny suggestion for `get_available_product_attributes()` and added a new test to test the method with product variations (03a66779).

I think this one is good to go. We are trying to include another tweak in this version, so I'll wait a few hours before doing the deploy. Should be out this afternoon.

includes/Products.php
maxdupenois: Oops, good catch and thank you for the test. Lazy of me to not add one myself.

wvega:
@maxdupenois this fix is now out in Facebook for WooCommerce 2.1.4 🎉

PR #1692, opened 2020-11-11
Rework filter placement to disable pixel tracking

unfulvio:
## Summary

Reworks the way the filter `facebook_for_woocommerce_integration_pixel_enabled` is triggered within `WC_Facebookcommerce_EventsTracker` so that it doesn't cause frontend JS errors.

## Details

Rather than changing the filter hook that would have called the filter added by Facebook, I figured it's simple enough to have a method called from constructor that checks whether the pixel should be enabled or not via `facebook_for_woocommerce_integration_pixel_enabled`. Then, just bail from applying all the filters the constructor is attaching (and move those into a separate method as well).

I checked the references of those methods and the public ones do not look to be called from elsewhere. Anyway, many of them already had some checks to see if the former `$isEnabled` flag was set, I merely replaced that with a call to the new method.

### Setup

- Make sure you are connected with Facebook and your pixel is working
- Install the Facebook Pixel helper from the Chrome web store to check pixel status on a page and inspect your analytics while browsing pages where the pixel is expected to track events
- Keep your console open to check for errors

#### No filter applied

1. Visit front end with a customer user
   - [x] The FB Pixel helper is active and tracking occurs
1. Add product to cart, wait for page reload, visit cart and then checkout
   - [x] Events are correctly tracked
   - [x] No JS error appears in console

#### Apply filter

1. Add a filter `add_filter( 'facebook_for_woocommerce_integration_pixel_enabled', '__return_false' );` with Code Snippets or similar
1. Visit the shop front end and perform some activity (page browsing, search, add to cart...)
    - [x] No pixel is loaded at all, no events are tracked
1. Add product to cart, wait for reload page and visit cart/checkout
    - [x] No JS errors appear in console

## Before merge

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -22,51 +22,86 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 	}
 
 	class WC_Facebookcommerce_EventsTracker {
-		private $pixel;
-		private static $isEnabled = true;
-		const FB_PRIORITY_HIGH    = 2;
-		const FB_PRIORITY_LOW     = 11;
 
+
+		/** @deprecated since 2.3.0-dev.1 */
+		const FB_PRIORITY_HIGH = 2;
+		/** @deprecated since 2.3.0-dev.1 */
+		const FB_PRIORITY_LOW = 11;
+
+
+		/** @var \WC_Facebookcommerce_Pixel instance */
+		private $pixel;
 
 		/** @var string name of the session variable used to store search event data */
 		private $search_event_data_session_variable = 'wc_facebook_search_event_data';
 
 		/** @var Event search event instance */
 		private $search_event;
+
 		/** @var array with events tracked */
 		private $tracked_events;
+
 		/** @var AAMSettings aam settings instance, used to filter advanced matching fields*/
 		private $aam_settings;
 
+
+		/**
+		 * Events tracker constructor.
+		 *
+		 * @param $user_info
+		 * @param $aam_settings
+		 */
 		public function __construct( $user_info, $aam_settings ) {
-			$this->pixel = new WC_Facebookcommerce_Pixel( $user_info );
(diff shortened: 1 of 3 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -87,14 +130,20 @@ public function __construct( $user_info, $aam_settings ) {
 			add_action( 'woocommerce_thankyou',                   [ $this, 'inject_purchase_event' ], 40 );
 
 			// TODO move this in some 3rd party plugin integrations handler at some point {FN 2020-03-20}
-			add_action( 'wpcf7_contact_form', [ $this, 'inject_lead_event_hook' ], self::FB_PRIORITY_LOW );
+			add_action( 'wpcf7_contact_form', [ $this, 'inject_lead_event_hook' ], 2 );
```
deivamagalhaes, line 133: Just out of curiosity, why did you change this priority?

deivamagalhaes (commented):
QA passes, just have a question.

facebook-commerce-events-tracker.php
unfulvio: I did not; it's the same as the constant that was set there - it wasn't a good approach to use constants for setting these priorities as they're not universal; we moved away from those as we started maintaining the plugin, this was still lingering. Since the code block had some changes, I just swapped the constant for a plain number.

dmagalhaes-godaddy: @unfulvio , `FB_PRIORITY_LOW` was 11, from the PR diff. `FB_PRIORITY_HIGH` was 2... Just want to make sure we are not missing something here...

unfulvio: you are very much right @deivamagalhaes ! my bad swapped the two, restored the original value of `11` here: https://github.com/facebookincubator/facebook-for-woocommerce/pull/1663/commits/24dcdcc2d674a019deb3818c5d307218e37d453c

PR #1727, opened 2020-12-10
Fix inventory sync issues

denniskwon:
- Remove commerce enabled flags to ensure inventory syncing
- Fix js errors on Commerce settings screen


wvega (commented):
Hey @denniskwon, I've a question about the removal of the **Sell on Instagram** checkbox. 

I think I overlooked the part of your question on Slack about removing the setting and focused on sending the inventory property regardless of whether the product was ready for Commerce or not.

Are there any plans to let merchants decide to sync some products to the Facebook catalog but not make them available for purchase?

If I remember correctly one of the reasons why we moved the background sync to use the `items_batch` endpoint was because that endpoint has support for the scenario in which merchants want to sync products with enhanced catalog fields but don't want to make them available for Commerce. Is that still something that the plugin should allow?

I understand that right now the checkbox is only controlling whether the inventory field is included or not, but if customers should have the ability to decide which products are available for Commerce, then I think the setting and the validation behind it is still necessary.

denniskwon:
Hey @wvega. Based on initial feedback from testing, we've decided to revisit the design and specific use cases for controlling visibility of products for Commerce. As it is today, we've found that the controls tend to be more broadly confusing for a narrow set of sellers who would need this level of control. We will likely be adding back some level of control in a future release.

wvega:
@denniskwon thank you for the explanation. I asked because a couple of the QA items on our list involved using the `items_batch` endpoint to exclude products from commerce based on the Sell on Instagram status for the product and adjusting the validation for that checkbox on the Edit Product page for variable products. I think that with the new direction those changes no longer apply.

wvega:
I'll check with the team about those QA items and then I'll be back to post another review.

wvega (approved):
@denniskwon I updated the branch with a couple of fixes we added to `release/instagram-checkout` and updated the minified JS files.

Once concern is that we normally don't remove public methods before deprecating them to avoid breaking a merchant's site without notice if they are somehow calling one of those methods from custom code.

We usually:

1. Add a `@deprecated {version}` tag to the method's docblock
2. Add a TODO note to the docblock indicating when (date) or at which version is safe to remove the method
2. Call `wc_deprecated_function()` inside the method
3. Replace the method body with a call to the replacement method or simply remove it if there is no replacement
4. Ensure that the return type stays the same

Here is an example: https://github.com/facebookincubator/facebook-for-woocommerce/blob/180b7f1d4d7438a66a6b2fbec41ddb127d14301a/facebook-commerce-events-tracker.php#L979-L996

I recommend doing the same for the public methods removed here. Other than that this should be ready to merge.

wvega-godaddy (approved):
@denniskwon thank you for adding those deprecation notices. Changes look good and integration tests pass for me. I think we can merge!

PR #1733, opened 2020-12-14
Set request authorization header when setting a new access token

unfulvio:
## Summary

Ensures the authorization header of an API request carries the current access token for the bearer.

## UI Changes

None

### Setup

I am not sure if I could replicate this correctly. Please check the story details.

Two API requests should be logged when you execute the scenario steps. One should be should be for the business config refresh on `init`, then one that's run by `wc_facebook_commerce_fetch_orders` in the same thread. If those are both run, then the `Bearer` in the `Authorization` header may be different because a new access token is passed: the first should be the main access token, and the second should be the page access token, which was generated when connected to Commerce.

Note that to read the `Authorization` value from the headers, you need to temporarily override `get_sanitized_request_headers()` in the `API` handler, otherwise the value will be obfuscated (default normal behavior) as sensitive information.

### Steps

1. Connect the plugin to Commerce using the Connection & orders instructions in epic [64675](https://app.clubhouse.io/skyverge/story/64675/epic-acceptance-criteria-met)
1. Go to the Scheduled Actions page
1. Delete the wc_facebook_business_configuration_refresh transient in the database
1. Manually run the wc_facebook_commerce_fetch_orders action
   - [x] You do not see the "Your connection is invalid" notice

### Tests

1. Run `codecept run integration APITest.php`
   - [x] Integration tests pass

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -47,12 +47,7 @@ class API extends Framework\SV_WC_API_Base {
 	 */
 	public function __construct( $access_token ) {
 
-		$this->access_token = $access_token;
-
-		$this->request_headers = [
-			'Authorization' => "Bearer {$access_token}",
-		];
-
+		$this->set_access_token( $access_token );
 		$this->set_request_content_type_header( 'application/json' );
 		$this->set_request_accept_header( 'application/json' );
 	}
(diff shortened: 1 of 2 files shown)


wvega (approved):
@unfulvio somehow the API requests were working for me without updating the `Authorization` header, so I'm unable to see the "Your connection is invalid" notice on this branch or `release/instagram-checkout`. However, I confirmed that changes in this PR allow the plugin to use the correct token on that header.

Tests pass as well.

unfulvio:
@wvega thanks for reviewing; yeah I wasn't able to replicate properly either, but the change still makes sense. In the logs I saw 2 requests carrying the token/bearer and it was identical all the times.

PR #1736, opened 2020-12-19
Enable custom retailer_id mapping for backwards compatibility

denniskwon:
Allow option to customize value used for retailer_id (sku, product_id, or sku_product_id) for compatibility with existing catalog feeds

### DIFF
--- a/includes/Admin/Settings_Screens/Product_Sync.php
+++ b/includes/Admin/Settings_Screens/Product_Sync.php
@@ -271,6 +271,22 @@ class Product_Sync extends Admin\Abstract_Settings_Screen {
 			],
 
 			[
+				'id'       => \WC_Facebookcommerce_Integration::SETTING_FB_RETAILER_ID_TYPE,
+				'title'    => __( 'Product Identifier', 'facebook-for-woocommerce' ),
+				'type'     => 'select',
+				'class'    => 'product-sync-field',
+				'desc_tip' => __( 'WooCommerce product attribute mapped to retailer_id on Facebook.
+					WARNING: use with caution, as changing this option after uploading products to Facebook may
+					result in having duplicate products in your Facebook catalog.', 'facebook-for-woocommerce' ),
+				'default'  => \WC_Facebookcommerce_Integration::FB_RETAILER_ID_TYPE_SKU_PRODUCT_ID,
+				'options'  => [
+					\WC_Facebookcommerce_Integration::FB_RETAILER_ID_TYPE_SKU_PRODUCT_ID => __( 'SKU + Product ID', 'facebook-for-woocommerce' ),
+					\WC_Facebookcommerce_Integration::FB_RETAILER_ID_TYPE_SKU            => __( 'SKU', 'facebook-for-woocommerce' ),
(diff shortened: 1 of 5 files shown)


includes/fbutils.php
```diff
@@ -90,12 +90,25 @@ public static function make_url( $url ) {
...
 			// products, which have their own variant_ids.
-			return $woo_product->get_sku() ? $woo_product->get_sku() . '_' .
-			$woo_id : self::FB_RETAILER_ID_PREFIX . $woo_id;
+			$woo_id = $woo_product->get_id();
+			$sku = $woo_product->get_sku();
+			$retailer_id = $sku ? $sku . '_' . $woo_id : self::FB_RETAILER_ID_PREFIX . $woo_id;
+
+			switch ( get_option( \WC_Facebookcommerce_Integration::SETTING_FB_RETAILER_ID_TYPE )) {
+				case \WC_Facebookcommerce_Integration::FB_RETAILER_ID_TYPE_SKU:
+					$retailer_id = $sku;
```
wvega, line 101: If the customer selected `FB_RETAILER_ID_TYPE_SKU` but a product doesn't have a SKU the `$retailer_id` variable ends up being empty. I think we should add a fallback here.

Maybe good idea to add a couple of tests to ensure we get a proper value in those cases as well.

includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -270,6 +270,22 @@ public function get_settings() {
 				'default' => 'yes',
 			],
 
+			[
+				'id'       => \WC_Facebookcommerce_Integration::SETTING_FB_RETAILER_ID_TYPE,
+				'title'    => __( 'Product Identifier', 'facebook-for-woocommerce' ),
+				'type'     => 'select',
+				'class'    => 'product-sync-field',
+				'desc_tip' => __( 'WooCommerce product attribute mapped to retailer_id on Facebook.
+					WARNING: use with caution, as changing this option after uploading products to Facebook may
```
wvega, line 279: Perhaps it would be good to move the Warning to a normal description (`desc`) to make it more easy for merchants to see the risks.

wvega (changes requested):
@denniskwon changes work for me, though I caught and edge case and I have a suggestion:

includes/fbutils.php
denniskwon: Yeah, I had thought of this. However, rather than trying to define the behavior here (pick an alt. field, for example, which could potentially cause more confusion) I thought it would actually be better to just let Facebook reject the product at the catalog sync level.

wvega: Got it. I think it makes sense the way you put it, but may not work that way right now because those rejections are very hard to detect for a user. The Batch API doesn't fail right away, instead you have to use the returned handle to check the status of the request and see a possible list of errors, and the plugin doesn't do that.

Probably good idea to add a FAQ entry or something similar in the documentation to instruct both merchants and support teams to check the new settings and the configuration of affected products when someone reports that some products are not syncing.

wvega (approved):
@denniskwon thank you for updating the description for the setting. It looks good!

I think we can merge this one as it is. If there are any updates or follow ups with regards to letting users know when a product cannot be synced because it's missing the SKU they can be handled in separate PRs.

PR #1749, opened 2021-01-18
Maxphi/fix for additional features error

maxdupenois:
Fix to make sure that 'list of strings' params are now converted to actual arrays. This broke as Facebook became stricter about how they ingested this parameter type.

Should fix the issue raised by the user using the 'additional_features' parameter.

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -673,7 +673,12 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 			$all_attributes = $category['attributes'];
 			foreach ( $all_attributes as $attribute ) {
 				$value            = Products::get_enhanced_catalog_attribute( $attribute['key'], $this->woo_product );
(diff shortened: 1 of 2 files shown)


jconroy:
👋 @maxdupenois would you happen to have any more info on the issue trying to be solved here e.g.

> Should fix the issue raised by the user using the 'additional_features' parameter.

Keen to replicate and confirm the problem is solved. Or are you pretty confident?

maxdupenois:
> 👋 @maxdupenois would you happen to have any more info on the issue trying to be solved here e.g.
> 
> > Should fix the issue raised by the user using the 'additional_features' parameter.
> 
> Keen to replicate and confirm the problem is solved. Or are you pretty confident?

@jconroy Pretty confident but easy to replicate :) If you create a product and fill in the additional features with a comma separated list, then (make sure logging is turned on) force a product sync. Check the logs and the call to /items_batch should have the additional_features going through as an array. Facebook used to accept a comma separated string for lists of strings but it appears that's been removed.

PR #1752, opened 2021-01-21
Remove business_management permission from login scopes

denniskwon:
FBE eliminates the need for business_management permission. Remove legacy permission to reduce scopes.


maxdupenois (approved):
Looks good to me

jconroy (approved):
Tested the changes and appears to be working as expected 👍

PR #1776, opened 2021-02-17
Fix product set

sc-ecolasurdo:
- Bug fixed when Product Set is empty.
- Change the version used for Product Set deletion.
- Implement a filter to send the _allow_live_product_set_deletion_ parameter in _true_ by default in Product Set deletion process.
Reference: [Facebook Documentation](https://developers.facebook.com/docs/marketing-api/reference/product-set#Deleting)

### DIFF
--- a/includes/fbgraph.php
+++ b/includes/fbgraph.php
@@ -408,7 +408,7 @@ if ( ! class_exists( 'WC_Facebookcommerce_Graph_API' ) ) :
 
 		public function delete_product_set_item( $product_set_id ) {
 
-			$params = ( true === apply_filters( 'wc_facebook_commerce_allow_live_product_set_deletion', $product_set_id ) ) ? '?allow_live_product_set_deletion=true' : '';
+			$params = ( true === apply_filters( 'wc_facebook_commerce_allow_live_product_set_deletion', true, $product_set_id ) ) ? '?allow_live_product_set_deletion=true' : '';
 
 			$url = $this->build_url( $product_set_id, $params, '9.0' );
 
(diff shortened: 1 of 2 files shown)


jconroy:
Thanks @sc-ecolasurdo !

includes/fbgraph.php
```diff
@@ -407,7 +407,11 @@ public function update_product_set_item( $product_set_id, $data ) {
 		}
 
 		public function delete_product_set_item( $product_set_id ) {
-			$url = $this->build_url( $product_set_id, '', '8.0' );
+
+			$params = ( true === apply_filters( 'wc_facebook_commerce_allow_live_product_set_deletion', $product_set_id ) ) ? '?allow_live_product_set_deletion=true' : '';
```
msaggiorato, line 411: Hey @sc-ecolasurdo , shouldn't the filter be like this? This way we default it to true, and avoid the need for a `__return_true` filter at the top.

```suggestion
			$params = ( true === apply_filters( 'wc_facebook_commerce_allow_live_product_set_deletion', true, $product_set_id ) ) ? '?allow_live_product_set_deletion=true' : '';
```

msaggiorato (changes requested):

msaggiorato (approved):

jconroy:
I've created a release branch https://github.com/facebookincubator/facebook-for-woocommerce/pull/1779

jconroy (approved):
This appears to be testing well for me.  Confirm product set is removed from FB if it becomes empty (I had a product set/category with a single item and removed that item from the category - confirmed on the FB end in my catalog)

Just had a suggestion for rewording one of the inline comments.

Can you re target the 2.3.1 branch and also entries to the readme and changelog in this PR

Cheers!

jconroy:
> Can you re target the 2.3.1 branch and also entries to the readme and changelog in this PR

I've gone and done that for you

PR #1778, opened 2021-02-18
Bump facebook version to v9.0 as v7.0 is close to deprecation

maxdupenois:
Summary: Update Facebook version to v9.0

Differential Revision: D26487283

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -32,7 +32,7 @@ class API extends Framework\SV_WC_API_Base {
 
 
 	/** @var string URI used for the request */
-	protected $request_uri = 'https://graph.facebook.com/v7.0';
+	protected $request_uri = 'https://graph.facebook.com/v9.0';
 
 	/** @var string the configured access token */
 	protected $access_token;
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -395,7 +395,7 @@ class Connection {
 
 		facebook_for_woocommerce()->log( 'Retrieving page access token' );
 
-		$response = wp_remote_get( 'https://graph.facebook.com/v7.0/me/accounts?access_token=' . $this->get_access_token() );
+		$response = wp_remote_get( 'https://graph.facebook.com/v9.0/me/accounts?access_token=' . $this->get_access_token() );
 
 		$body = wp_remote_retrieve_body( $response );
 		$body = json_decode( $body, true );
--- a/tests/integration/APITest.php
+++ b/tests/integration/APITest.php
@@ -574,7 +574,7 @@ class APITest extends \Codeception\TestCase\WPTestCase {
 
 		$response_data = [
 			'paging' => [
-				'next' => 'https://graph.facebook.com/v7.0/1234/products?fields=id,retailer_id&limit=1000&after=ABCD',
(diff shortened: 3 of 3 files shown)


jconroy:
Additional context 

> Your app will be affected by the deprecation of Marketing API versions prior to v8.0.

> Your app WooCommerce Integration, currently accesses a Marketing API version prior to v8.0. We will be deprecating all versions prior to v8.0 on Wednesday, March 3, 2021. To ensure a smooth transition, please migrate all calls to Marketing API v8.0 or higher. We recommend migrating to the most recent version, Marketing API v9.0, to reduce the number of future migrations for developers.

> Visit our [changelog](https://developers.facebook.com/docs/graph-api/changelog) to see all upcoming Marketing API deprecation dates as well as the full list of changes in all versions.

maxdupenois:
> @maxdupenois What's supposed to happen with the requests that are sent through this URL: https://github.com/facebookincubator/facebook-for-woocommerce/blob/release-2.3.1/includes/fbgraph.php#L27
> 
> When I go to edit a product I see a request sent out to the URL:
> `GET https://graph.facebook.com/v2.9/catalog:383672016237451:aW50ZWxsaWdlbnQtd29vZGVuLXdhbGxldC0xMjc5NjM4MF80NA==/?fields=id,product_group{id}`
> 
> There are also some functions there which override the URL version, do we need to maintain those as well to prevent them from eventually getting outdated? See: https://github.com/facebookincubator/facebook-for-woocommerce/blob/release-2.3.1/includes/fbgraph.php#L397-L416

@mikkamp oh interesting. I searched for 'v7.0' did not see that there was something as old as 'v2.9' there. so that url should return information about the product item I believe, specifically its Facebook id and its product group id, however that version was deprecated on July 22 2019 so I suspect it does absolutely nothing useful. Out of interest do you see a response?

With regard to the second part, ah, I suspect that was added in the recent push because the devs didn't want to change the version across the whole plugin but needed the up to date version for those endpoints. Probably best to simply remove the override to make it less confusing. I'll see if I can do that later.

jconroy:
Let's leave these version changes for a follow-up release.

mikkamp:
> so that url should return information about the product item I believe, specifically its Facebook id and its product group id, however that version was deprecated on July 22 2019 so I suspect it does absolutely nothing useful. Out of interest do you see a response?

Yes I get a valid response the first time round (but then it saves the ID in meta so it no longer loads it again). See: https://github.com/facebookincubator/facebook-for-woocommerce/blob/release-2.3.1/facebook-commerce.php#L3619

I haven't really dug into all the code, but it seems requests like this one, all get sent through the old version: https://github.com/facebookincubator/facebook-for-woocommerce/blob/release-2.3.1/facebook-commerce.php#L3581
That gets triggered when I change a product to draft or publish it.

Even though 2.9 is deprecated, all the responses seem to still work, so we might need to take a closer look at all the requests that are being sent through that class to see if we can get it to use a more recent version.

jconroy:
@maxdupenois are you able to get back to this one before March 3? We'll probably want to wrap up a release and ship by the 2nd

maxdupenois:
Put up a new pull request with a more complete approach to managing the version bump

PR #1781, opened 2021-02-19
Ensure that events have an action_source

maxdupenois:
Summary: Events sent via woocommerce plugin don't currently have an action_source, this is to fix that. Opted for a defaults of 'website'

Differential Revision: D26545761

### DIFF
--- a/includes/Events/Event.php
+++ b/includes/Events/Event.php
@@ -88,6 +88,7 @@ class Event {
 	protected function prepare_data( $data ) {
 
 		$this->data = wp_parse_args( $data, [
+			'action_source'    => 'website',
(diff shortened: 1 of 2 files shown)


jconroy:
Additional context

> Basically, every event must have the 'action_source', 'client_user_agent' and 'event_source_url' https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/server-event#action-source. The plugin already sends 'client_user_agent' and 'event_source_url', it just needs to add the action_source parameter.

mikkamp (approved):
Looks good. I confirmed that the `action_source` is included when sending the event. The tests run well.

PR #1783, opened 2021-02-22
FIX for: Logs being recorded even with debug option disabled

wilmerlm11:
Also reported here: https://wordpress.org/support/topic/logs-being-recorded-even-with-debug-option-disabled/

There are are messages spamming the logs, even with the debug option being disabled in the plugin config.

It comes from:
facebook-for-woocommerce\includes\Handlers\Connection.php

Line 394: private function retrieve_page_access_token( $page_id )
One of them is:
facebook_for_woocommerce()->log( ‘Retrieving page access token’ );

What I saw is that the log function has not been overridden to check the debug option as it’s done in the similar function log_api_request.

I'll create a PR with the suggested change to fix that.

Hopefully it can make the next release.
It’s annoying the need to keep cleaning up my logs if I chose to not log anything from this plugin.

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -368,6 +368,21 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 			new WP_Facebook_Integration();
 		}
 
+		/**
+		 * Saves errors or messages to WooCommerce Log (woocommerce/logs/plugin-id-xxx.txt)
+		 *
+		 * @since 2.0.0
+		 * @param string $message error or message to save to log
+		 * @param string $log_id optional log id to segment the files by, defaults to plugin id
+		 */
+		public function log( $message, $log_id = null ) {
+			// bail if logging isn't enabled
+			if ( ! $this->get_integration() || ! $this->get_integration()->is_debug_mode_enabled() ) {
+				return;
+			}
+
+			parent::log( $message, $log_id );
+		}
 
 		/**
 		 * Logs an API request.
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -371,7 +371,7 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 		/**
 		 * Saves errors or messages to WooCommerce Log (woocommerce/logs/plugin-id-xxx.txt)
 		 *
-		 * @since 2.3.1-dev.1
+		 * @since 2.3.3
 		 * @param string $message error or message to save to log
 		 * @param string $log_id optional log id to segment the files by, defaults to plugin id
 		 */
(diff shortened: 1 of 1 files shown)


class-wc-facebookcommerce.php
```diff
@@ -368,6 +368,21 @@ public function add_wordpress_integration() {
 			new WP_Facebook_Integration();
 		}
 
+		/**
+		 * Saves errors or messages to WooCommerce Log (woocommerce/logs/plugin-id-xxx.txt)
+		 *
+		 * @since 2.0.0
```
jconroy, line 374: Can you update to use `2.3.1-dev.1`

jconroy (changes requested):
Thanks so much for taking the time to submit a PR @wilmerlm11 , just one really quick comment before a proper test and review

And just to confirm - the way to replicate your issue is to simply disable debugging (in the plugin) and we should still see a facebook specific log file  generated and populated.

After this patch that won't happen

Is that your expectation?

wilmerlm11 (commented):
Updated "since" version as requested :)

wilmerlm11:
Hi @jconroy ,
thank you for reviewing it!
Your test case is perfect. That's exactly it.
After the fix, you should only see those logs recorded through facebook_for_woocommerce()->log( ), IF the debugging option is enabled in the plugin admin config.
Regards,
Wil

jconroy:
Just making a note that I've replicated the issue

Screen Shot 2021-02-24 at 8 42 52 pm

wilmerlm11:
@jconroy , I've just noticed 2.3.1 is out (without this fix :( ).
Should we keep 2.3.1 in the "since" version?
Any expectation to have this fix merged, since it's a simple one?

Thanks!!

jconroy:
Sorry we didn't get a chance to test properly before last week's release. 

> Any expectation to have this fix merged, since it's a simple one?

Absoolutely, appreciate the contribution - we just need to find some time to double check things.

class-wc-facebookcommerce.php
```diff
@@ -368,6 +368,21 @@ public function add_wordpress_integration() {
 			new WP_Facebook_Integration();
 		}
 
+		/**
+		 * Saves errors or messages to WooCommerce Log (woocommerce/logs/plugin-id-xxx.txt)
+		 *
+		 * @since 2.3.1-dev.1
```
rcstr, line 374: We'll need to update the @`since` tag to `2.3.3` before merging.

rcstr (changes requested):
@wilmerlm11 - I have tested your changes, and it works as expected. Thank you for submitting this PR.

I have requested changes because we need to update the `version` before we merge

class-wc-facebookcommerce.php
wilmerlm11: Hi @rcstr !
Tks for validating the fix.
I've just updated the since first as requested.

rcstr (approved):
thank you @wilmerlm11!

wilmerlm11:
Hi @jconroy !
Does it need you approval yet?
I'm afraid the release train will pass and I'll have to update the "since" version once again :/
Regards

jconroy:
All merged into the 2.3.3 release branch @wilmerlm11 - thank you!

PR #1784, opened 2021-02-23
Fix empty "value" for variable products

alexndlm:
Set "value: to minimal variation price for variable products.


jconroy:
Hi @alexndlm thanks for the contribution - can you include some more detail about the issue and how we can replicate and test the fix please?

alexndlm:
Hello @jconroy 

WooCommerce method `[get_price](https://woocommerce.github.io/code-reference/files/woocommerce-includes-abstracts-abstract-wc-product.html#source-view.266)` return price only for regular product, exactly it return 'price' meta.

For variable products, the variation price is stored in a separate product (variation). If you call the method `get_price` on the main product, it returns an empty string.

For a variable product, there is a special method `[get_variation_price](https://woocommerce.github.io/code-reference/classes/WC-Product-Variable.html#method_get_variation_price)` that returns the minimum or maximum price of a variation.

It seems to me that it makes sense to show the minimum price in the conversion value, so it's better to underestimate expectations than overestimate.

rcstr (commented):
Tested and works as expected, left a few notes.

rcstr (approved):

PR #1787, opened 2021-02-26
Facebook version bump marketing api to 9.0

maxdupenois:
Move plugin to use v9.0 of the marketing version as it's currently on 7.0 (mostly) and will be deprecated soon. Also cleaned up how the version is referenced to make this easier to manage.

Fixed the tests then checked that product sync and product updates still work (the only potentially affected endpoint being the slightly odd way that the plugin attempts to update a product, not the batch sync but the single product update).

## Relevant changelogs:
https://developers.facebook.com/docs/graph-api/changelog/version8.0#marketing-api
https://developers.facebook.com/docs/graph-api/changelog/version9.0#marketing-api

## Version List:
https://developers.facebook.com/docs/graph-api/changelog/versions#marketing-api

### DIFF
--- a/includes/Admin/Settings_Screens/Advertise.php
+++ b/includes/Admin/Settings_Screens/Advertise.php
@@ -93,7 +93,7 @@ class Advertise extends Admin\Abstract_Settings_Screen {
 					appId            : '<?php echo esc_js( $connection_handler->get_client_id() ); ?>',
 					autoLogAppEvents : true,
 					xfbml            : true,
-					version          : 'v8.0',
+					version          : 'v8.0', // Note: This expires on November 1 2022
 				} );
 			};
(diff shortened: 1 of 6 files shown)


includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -93,7 +93,7 @@ public function output_scripts() {
 					appId            : '<?php echo esc_js( $connection_handler->get_client_id() ); ?>',
 					autoLogAppEvents : true,
 					xfbml            : true,
-					version          : 'v8.0',
+					version          : 'v8.0', // Note: This expires on November 1 2022
```
maxdupenois, line 96: The graph api has a different schedule to the marketing one

PR #1791, opened 2021-03-02
Don’t autoload OPTION_GOOGLE_PRODUCT_CATEGORIES

WPprodigy:
This option is quite large. The web request response itself is 483kb and after the parsing I'm seeing it around 708kb on a client site.

Because it's currently autoloaded, this means the `OPTION_GOOGLE_PRODUCT_CATEGORIES` option itself is pushing a site very close to their `alloptions` exceeding 1MB, which is where most object caches stop working. And when `alloptions` can't be stored in the object cache, the database load increases drastically - generally breaking high-traffic sites at that point.

The solution here is quite simple though. Since this option is both only accessed in the admin (I think?) and also accessed primarily from it's own transient, it's an easy win to just not autoload it using the third param here: https://developer.wordpress.org/reference/functions/update_option/

### DIFF
--- a/includes/Admin/Google_Product_Category_Field.php
+++ b/includes/Admin/Google_Product_Category_Field.php
@@ -64,7 +64,7 @@ class Google_Product_Category_Field {
 			if ( ! empty( $categories ) ) {
 
 				set_transient( self::OPTION_GOOGLE_PRODUCT_CATEGORIES, $categories, WEEK_IN_SECONDS );
-				update_option( self::OPTION_GOOGLE_PRODUCT_CATEGORIES, $categories );
+				update_option( self::OPTION_GOOGLE_PRODUCT_CATEGORIES, $categories, 'no' );
 			}
 		}
 


rcstr (approved):
LGTM! thank you @WPprodigy

WPprodigy:
May need to keep an eye on this, as if it starts to near/exceed 1MB then it'll then fail to be stored in the transient when memcached is in use - resulting in the remote request taking place each time. Okay for now though.

jconroy:
I've changed this to the 2.3.3 base branch, merging

mikobig:
Hi all, I recently updated to 2.3.3 but the wc_facebook_google_product_categories option is still autoloaded.
Is there an extra step to exclude it for already active/configured plugins? Thank you!

haszari:
> Hi all, I recently updated to 2.3.3 but the wc_facebook_google_product_categories option is still autoloaded.
Is there an extra step to exclude it from autoload for already active/configured plugins? Thank you!

Thank you @mikobig for reporting!

@rcstr Can you investigate?

WPprodigy:
The transient expires after a week, and then the option will refresh - resulting in it no longer being autoloaded: https://github.com/WPprodigy/facebook-for-woocommerce/blob/2d287f26989f8d2a056571e8f6b4b05f2b381bf6/includes/Admin/Google_Product_Category_Field.php#L66

Could expedite the process by manually deleting the transient with `wp transient delete wc_facebook_google_product_categories` for example, and then I believe reload an admin-side page and it will regenerate.

mikobig:
1 week passed and the option is still autoloaded. There is no such transient in the database.
Thank you!

WPprodigy:
Hmm, that is strange. Can you confirm this line looks the same on the version you are running: https://github.com/woocommerce/facebook-for-woocommerce/blob/77f964b075aed4fd060cc39db1e7a780c8bfdf8c/includes/Admin/Google_Product_Category_Field.php#L67

I suppose another way to force it over to being not autoloaded is to do this just once via WP CLI or similar:

```
$categories = get_option( 'wc_facebook_google_product_categories', [] );
update_option( 'wc_facebook_google_product_categories', $categories, 'no' );
```

mikobig:
I can confirm that line 67 is OK:
`update_option( self::OPTION_GOOGLE_PRODUCT_CATEGORIES, $categories, 'no' );`

Unfortunately no WP-CLI access at this moment.

I check my wp_options table:
SELECT option_name, length(option_value) AS option_value_length FROM wp_options WHERE autoload='yes' ORDER BY option_value_length DESC LIMIT 10;

image

Is there an alternative via phpMyAdmin

WPprodigy:
Ahhhh, I bet it's that the categories list hasn't changed so the update_option() just aborts early here: https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-includes/option.php#L397-L408

You'll want to delete the transient & the option, then load an admin page. That should regenerate it. Alternatively, can just update the `autoload` column in SQL directly for that one option_name.

mikobig:
Deleted the option, all good now. @WPprodigy thanks for the quick replies!

PR #1794, opened 2021-03-04
Sanitize attribute names to correctly auto-match EC attributes

denniskwon:
This change correctly sanitizes WooCommerce attributes to automatically match EC attribute names

### DIFF
--- a/includes/fbutils.php
+++ b/includes/fbutils.php
@@ -418,18 +418,6 @@ if ( ! class_exists( 'WC_Facebookcommerce_Utils' ) ) :
 			return ( self::$store_name ) ? ( self::$store_name ) : 'A Store Has No Name';
 		}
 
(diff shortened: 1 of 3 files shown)


includes/fbutils.php
```diff
@@ -421,23 +421,25 @@ public static function get_store_name() {
 		/*
 		* Change variant product field name from Woo taxonomy to FB name
 		*/
-		public static function sanitize_variant_name( $name ) {
+		public static function sanitize_variant_name( $name, $use_custom_data = true ) {
```
maxdupenois, line 424: Oh nice, good way of making sure we don't break anything.

maxdupenois (approved):
LGTM

rcstr (approved):
LGMT

PR #1795, opened 2021-03-04
Fixes / optimizations for enhanced catalog attribute syncing

denniskwon:
- Add fix to use visible value for attribute instead of slug
- Perf: Only process EC attributes matching a given product


rcstr (approved):
LGTM, thank you @denniskwon

PR #1800, opened 2021-03-10
use container price for Mix and Match products in per-item pricing mode

### LINKED ISSUE #1798: Add support for Mix and Match product price
## Feature requests

I have a customer using Mix and Match to group chocolates together. The price of the box depends on the price of the optional products. But the price showing up at Facebook is $0 (which is the base price of the container itself).

I see there's [explicitly compatibility for Product Bundles and Composites](https://github.com/facebookincubator/facebook-for-woocommerce/blob/05767e3221c993d981b33a00d8780969ab2e0357/includes/Products.php#L429-L440) and since they are very similar to Mix and Match, I was hoping you might add Mix and Match in there too. 

An additional `elseif` should look like this:

```
} elseif ( $product->is_type( 'mix-and-match' ) && $product->is_priced_per_product() ) {
	// If product is Mix and Match product with individually priced items, we rely on their pricing.
	$price = wc_get_price_to_display( $product, [ 'price' => $product->get_container_price() ] );
}

### PR DESCRIPTION
**Pull request by helgatheviking**
Closes #1798.

I think to test you'd need a Mix and Match product with "per-item pricing". Unfortunately, I have no way to test since I don't have a facebook store, but we're pretty closely mimicking Product Bundles/CP here so I would expect it to work the same.

### CODE REVIEW

File: includes/Products.php [hunk 1]
```diff
@@ -437,6 +437,11 @@ public static function get_product_price( \WC_Product $product ) {
 			// if product is a product bundle with individually priced items, we rely on their pricing
 			$price = wc_get_price_to_display( $product, [ 'price' => $product->get_bundle_price() ] );
```
franticpsyx, line 438:
```suggestion
			$price = $product->get_bundle_price( 'min', true );
```

franticpsyx (commented):
Also at https://github.com/facebookincubator/facebook-for-woocommerce/pull/1800/files#diff-b7ec7bedc72ef8ef24a93ec34b91129fe25b983b56f6e31a7420f891d2aa738cR431

Replace:

```
$price = get_option( 'woocommerce_tax_display_shop' ) === 'incl' ? $product->get_composite_price_including_tax() : $product->get_composite_price();
```

With:

```
$price = $product->get_composite_price( 'min', true );
```

includes/Products.php
```diff
@@ -437,6 +437,10 @@ public static function get_product_price( \WC_Product $product ) {
...
 
+		} elseif ( $product->is_type( 'mix-and-match' ) && $product->is_priced_per_product() ) {
+			// If product is Mix and Match product with individually priced items, we rely on their pricing.
+			$price = wc_get_price_to_display( $product, [ 'price' => $product->get_container_price() ] );
+
```
helgatheviking, line 443: Strong pandemic brain move... `get_container_price()` won't exist until 2.0.

haszari:
@danielbitzer Can you give this a review - see #1798 for more info & context.

helgatheviking:
Do you think I should add sub-2.0 support for Mix and Match? `get_container_price()` wont exist until then but it's just a renaming. I'm in the homestretch for 2.0 but gosh that still feels like it could take a while at the rate I am moving.

danielbitzer (changes requested):
@helgatheviking this PR conflicts with the `master` branch. Could you please merge master into this branch and resolve them.

Other thoughts:
- How can I test and view mix and match 2.0 which is required to test this PR? 
- When is the expected release date of 2.0?
- Could this additional pricing logic hook into the `wc_facebook_product_price` filter? Personally, I think it would be best to add this code to the Mix and Match plugin and use a filter instead of here.

rawdreeg:
I will close this PR as stale.

helgatheviking:
Refreshed PR #2437

Issue #1807, opened 2021-03-16
Implement system that will allow automatic re-sync on plugin update.


budzanowski:
## Feature requests

For some of the releases, an update procedure is required. This procedure would consist of various actions like for example re-sync the products to the FB database.

## Developers

We welcome your contributions. 🙌 Here are a few tips to help us work together:

- If you submit a new PR, please ensure the description explains why the change is needed, follows WordPress coding standards, is compatible with PHP 5.6, and contains a small changeset — PRs with massive changesets are likely to be rejected or asked for revisions, as they’re much harder to review for potential issues.
- If you’d like to propose a more substantial changeset, feel free to start a discussion of why and the proposed implementation in this issue! Giving our team a heads up increases the likelihood of changes being accepted quickly.
- Internet points are available for PRs that include test coverage! 🏆 💯 Especially failing + passing tests for bug fixes.

haszari:
I think we can close this now #1808 is merged and shipped in v2.3.3.

PR #1808, opened 2021-03-16
Create 2.3.3. upgrade function.

budzanowski:
This code forces all product resync on plugin update to "2.3.3" version.

Fixes: #1807 

The update procedure is triggered on `wp_loaded` hook so:

> do_action( 'wp_loaded' )
This hook is fired once WP, all plugins, and the theme are fully loaded and instantiated.

This means that we have access to all that we have. 

`facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_all_products();`
Just registers the products for updates. The actual update process is triggered by the `shutdown` hook and later carried on in the background by `$job_handler = facebook_for_woocommerce()->get_products_sync_background_handler();`

### DIFF
--- a/includes/Lifecycle.php
+++ b/includes/Lifecycle.php
@@ -42,6 +42,7 @@ class Lifecycle extends Framework\Plugin\Lifecycle {
 			'2.0.0',
 			'2.0.3',
 			'2.0.4',
+			'2.3.3',
 		];
 	}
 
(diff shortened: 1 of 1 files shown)


denniskwon:
In order to test the changes, I needed to navigate to /wp-admin/options.php and ensure the version # saved under 'wc_facebook_for_woocommerce_version' was set to a version lower than 2.3.3 first. After making those changes, I was able to successfully trigger a sync on upgrade as expected.

rcstr (approved):
LGTM!

haszari:
@budzanowski Do you think we should add an explicit changelog item for this? I think it's a good idea even though it's part of `Fix for WooCommerce variation attribute sync not matching Enhanced Catalog attributes`.

PR #1817, opened 2021-03-22
add connection state (boolean) to tracker snapshot

### LINKED ISSUE #1816: Track initial connection success & errors
There are various reports of issues getting the plugin onboarded and connected to Facebook, often with a generic `Something went wrong` error. Examples: #1711 #1697 

It would be useful to know how often this is happening - e.g. % of sites with connection errors, or log raw connection success/failure. 

This would give us more insight into how this issue is affecting sites, confirm fixes are effective, and give us insight to any new issues / errors that might cause connection problems.

haszari:
Fixes #1816 

This PR adds a new class to add data about this plugin's state to the [WooCommerce tracker snapshot](https://woocommerce.com/usage-tracking/). Note - tracking is opt-in per store; see `WooCommerce > Settings > Advanced > WooCommerce.com`.

The one new property added in this PR is a boolean value indicating whether the plugin has successfully been connected to facebook. This is stored in `$data[extensions][facebook-for-woocommerce][is-connected]` key, for example:

```
  [extensions] => Array
    (
        (
          [is-connected] => true
        )
    )
```

The goal here is that we can use the tracker data to better understand how many sites are experiencing connection problems. 

#### How to test this PR
The easiest way to test this is to use `wp shell` ([WP-CLI](https://developer.wordpress.org/cli/commands/shell/)) and/or add debug logging messages.

1. Ensure tracking is enabled for your test site: `WooCommerce > Settings > Advanced > WooCommerce.com > Enable tracking`.
2. Add debug code to `Tracker::add_tracker_data()` to log tracker data (see example below).
3. Trigger a sync. One way to do this is to use `wp shell`:
  - `wp shell` to get an interactive shell session
  - Call `WC_Tracker::send_tracking_data( true );` to force a snapshot.
  - Note: this function checks a timestamp ( `woocommerce_tracker_last_send` ) to avoid duplicate requests. You may need to delete this option to ensure the data is regenerated.
5. Repeat tests with different settings, e.g. plugin connected, disconnected, or failed connection. Please test a range of scenarios to ensure there aren't any false positives or vice versa. For example, if the plugin cannot sync and `is-connected=true`, that's probably a bug.

```php
	public function add_tracker_data( array $data = [] ) {
		$connection_handler = facebook_for_woocommerce()->get_connection_handler();
		if ( ! $connection_handler ) {
			return;
		}

		if ( ! isset( $data['extensions'] ) ) {
			$data['extensions'] = [];
		}

		$data['extensions']['facebook-for-woocommerce']['is-connected'] = $connection_handler->is_connected();

		// add logging here  
		wc_get_logger()->debug( print_r( $data, true ) );

		return $data;
	}
```

### Changelog entry

> New: Add connection state to [WooCommerce Usage Tracking](https://woocommerce.com/usage-tracking/)

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -88,6 +88,8 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 		/** @var \SkyVerge\WooCommerce\Facebook\Commerce commerce handler */
 		private $commerce_handler;
 
+		/** @var \SkyVerge\WooCommerce\Facebook\Tracker */
+		private $tracker;
 
 		/**
 		 * Constructs the plugin.
@@ -152,6 +154,7 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 				require_once __DIR__ . '/includes/Events/Normalizer.php';
 				require_once __DIR__ . '/includes/Events/AAMSettings.php';
 				require_once __DIR__ . '/includes/Utilities/Shipment.php';
+				require_once __DIR__ . '/includes/Utilities/Tracker.php';
 
 				$this->product_feed              = new \SkyVerge\WooCommerce\Facebook\Products\Feed();
 				$this->products_stock_handler    = new \SkyVerge\WooCommerce\Facebook\Products\Stock();
@@ -184,6 +187,10 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 
 				$this->connection_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\Connection( $this );
 				$this->webhook_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\WebHook( $this );
+
+				if ( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() ) {
+					$this->tracker = new \SkyVerge\WooCommerce\Facebook\Utilities\Tracker();
+				}
 
 				// load admin handlers, before admin_init
 				if ( is_admin() ) {
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -188,9 +188,7 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 				$this->connection_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\Connection( $this );
 				$this->webhook_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\WebHook( $this );
 
-				if ( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() ) {
-					$this->tracker = new \SkyVerge\WooCommerce\Facebook\Utilities\Tracker();
-				}
+				$this->tracker = new \SkyVerge\WooCommerce\Facebook\Utilities\Tracker();
 
 				// load admin handlers, before admin_init
 				if ( is_admin() ) {
--- a/includes/Utilities/Tracker.php
+++ b/includes/Utilities/Tracker.php
@@ -45,8 +45,6 @@ class Tracker {
 
 		$data['extensions']['facebook-for-woocommerce']['is-connected'] = $connection_handler->is_connected();
 
-		wc_get_logger()->debug( print_r( $data, true ) );
-
 		return $data;
 	}
 }
--- a/includes/Utilities/Tracker.php
+++ b/includes/Utilities/Tracker.php
@@ -41,20 +41,19 @@ class Tracker {
 	 * @since 2.3.4
 	 */
 	public function add_tracker_data( array $data = array() ) {
-		$connection_handler = facebook_for_woocommerce()->get_connection_handler();
-		if ( ! $connection_handler ) {
-			return $data;
-		}
-
 		if ( ! isset( $data['extensions'] ) ) {
 			$data['extensions'] = array();
 		}
 
-		$connection_is_happy = $connection_handler->is_connected() && ! get_transient( 'wc_facebook_connection_invalid' );
+		$connection_is_happy = false;
+		$connection_handler = facebook_for_woocommerce()->get_connection_handler();
+		if ( $connection_handler ) {
+			$connection_is_happy = $connection_handler->is_connected() && ! get_transient( 'wc_facebook_connection_invalid' );
+		}
 		$data['extensions']['facebook-for-woocommerce']['is-connected'] = wc_bool_to_string( $connection_is_happy );
 
 		// Debug logging - remove before merge.
-		facebook_for_woocommerce()->log( 'Added Facebook connection state to tracker snapshot: ' . print_r( $data, true ) );
+		wc_get_logger()->debug( 'Added Facebook connection state to tracker snapshot: ' . print_r( $data, true ) );
 
 		return $data;
 	}
--- a/includes/Utilities/Tracker.php
+++ b/includes/Utilities/Tracker.php
@@ -52,9 +52,6 @@ class Tracker {
 		}
 		$data['extensions']['facebook-for-woocommerce']['is-connected'] = wc_bool_to_string( $connection_is_happy );
 
-		// Debug logging - remove before merge.
-		wc_get_logger()->debug( 'Added Facebook connection state to tracker snapshot: ' . print_r( $data, true ) );
-
 		return $data;
 	}
 }
--- a/includes/Utilities/Tracker.php
+++ b/includes/Utilities/Tracker.php
@@ -51,7 +51,10 @@ class Tracker {
 		}
 
 		$connection_is_happy = $connection_handler->is_connected() && ! get_transient( 'wc_facebook_connection_invalid' );
-		$data['extensions']['facebook-for-woocommerce']['is-connected'] = $connection_is_happy;
+		$data['extensions']['facebook-for-woocommerce']['is-connected'] = wc_bool_to_string( $connection_is_happy );
+
+		// Debug logging - remove before merge.
+		facebook_for_woocommerce()->log( 'Added Facebook connection state to tracker snapshot: ' . print_r( $data, true ) );
 
 		return $data;
 	}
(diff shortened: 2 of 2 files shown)


class-wc-facebookcommerce.php
```diff
@@ -185,6 +188,10 @@ public function init() {
 				$this->connection_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\Connection( $this );
 				$this->webhook_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\WebHook( $this );
 
+				if ( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() ) {
+					$this->tracker = new \SkyVerge\WooCommerce\Facebook\Utilities\Tracker();
```
haszari, line 192: I've added this class to `Utility` folder and followed the existing namespace pattern (`SkyVerge`). Keen for feedback on all this and also how the code is loaded.

includes/Utilities/Tracker.php
```diff
@@ -0,0 +1,52 @@
...
+namespace SkyVerge\WooCommerce\Facebook\Utilities;
+
+defined( 'ABSPATH' ) or exit;
+
+/**
+ * Class for adding diagnostic info to WooCommerce Tracker snapshot.
+ *
+ * See https://woocommerce.com/usage-tracking/ for more information.
+ *
+ * @since %VERSION%
```
haszari, line 20: I'm using this `%VERSION%` as a temporary placeholder, will need to be a real version when this ships. 

If there's automation for updating these let me know, otherwise I'll update when we know what version we're targeting. Likely v2.3.4.

includes/Utilities/Tracker.php
```diff
@@ -0,0 +1,52 @@
...
+			return;
+		}
+
+		if ( ! isset( $data['extensions'] ) ) {
+			$data['extensions'] = [];
+		}
+
+		$data['extensions']['facebook-for-woocommerce']['is-connected'] = $connection_handler->is_connected();
+
+		wc_get_logger()->debug( print_r( $data, true ) );
```
budzanowski, line 48: `facebook_for_woocommerce()->log` should be used, it will also respect the debug on/off option.

class-wc-facebookcommerce.php
```diff
@@ -185,6 +188,10 @@ public function init() {
 				$this->connection_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\Connection( $this );
 				$this->webhook_handler = new \SkyVerge\WooCommerce\Facebook\Handlers\WebHook( $this );
 
+				if ( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() ) {
```
budzanowski, line 191: The 
`if( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() )` bit is not needed - this is already happening at line [130](https://github.com/facebookincubator/facebook-for-woocommerce/pull/1817/files#diff-a1e6a6cf0a7255afea85e377040a757915fa8707bb9e88d7348348ee27d79f77R130).

class-wc-facebookcommerce.php
budzanowski: I would love to start using just the `WooCommerce` namespace without `SkyVerge` in the front since we are doing this now.

includes/Utilities/Tracker.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace SkyVerge\WooCommerce\Facebook\Utilities;
+
+defined( 'ABSPATH' ) or exit;
+
+/**
+ * Class for adding diagnostic info to WooCommerce Tracker snapshot.
+ *
+ * See https://woocommerce.com/usage-tracking/ for more information.
```
budzanowski, line 18: We need to add Facebook plugin entry to that document.

budzanowski:
I wonder if we have any experience in adding data collection to an existing plugin. Although this is specific to only those merchants that already have agreed to send us data we still need to make sure that this is not a hidden "feature". Maybe when the user consents it means that we may add additional tracking points without asking again - that would be a good question to someone from legal.

class-wc-facebookcommerce.php
budzanowski: But that is something that we should address in a separate PR with a bigger scope.

includes/Utilities/Tracker.php
haszari: Indeed - will need to ensure that's done when we release. Is there something in the release process about documentation updates?

includes/Utilities/Tracker.php
haszari: Indeed, though I should just remove this debug logging - I left it in by mistake!

haszari:
Thanks for reviewing @budzanowski !

> Maybe when the user consents it means that we may add additional tracking points without asking again - that would be a good question to someone from legal.

Good point - I think we need to be transparent about this. Including on the documentation page and adding this clearly to changelog might be enough; I'll follow up. 

> I wonder if we have any experience in adding data collection to an existing plugin

I suspect we do, AutomateWoo for example. I can find out what process we used previously.

haszari:
Added `blocked` label for now, this should not be merged until we have confirmed how we'll document this change. (And also the logging should be removed before merge.)

class-wc-facebookcommerce.php
haszari: Good spotting! I missed that. Fixed here: ccb29a2fff8662250045fe84e33ba2a5c6f631d4

haszari:
Update - I revisited this and did some more testing, ended up hitting some issues with disconnecting and reconnecting (#1822). 

Since last review I've fixed up various phpcs issues with the code, and improved the check. It wasn't correctly detecting when the site connection was broken. There may be other things we need to check to determine the connection health, or it might be better to add a `last successful sync` date/time or similar data to confirm things are working correctly.

This PR is currently blocked on:

- Confirm that the `is_connected` value is accurate. Specifically when site connection is broken – no false positive. (Needs more testing.)
- Documentation updated and confirm with legal about adding new tracker props to plugin.
- Version numbers updated in docblocks (`%VERSION%`).

includes/Utilities/Tracker.php
```diff
@@ -0,0 +1,58 @@
...
+		if ( ! $connection_handler ) {
+			return $data;
+		}
+
+		if ( ! isset( $data['extensions'] ) ) {
+			$data['extensions'] = array();
+		}
+
+		$connection_is_happy = $connection_handler->is_connected() && ! get_transient( 'wc_facebook_connection_invalid' );
+		$data['extensions']['facebook-for-woocommerce']['is-connected'] = $connection_is_happy;
```
haszari, line 54: I think there's a convention to use underscores in tracker keys, so I might change these to:

`$data['extensions']['facebook_for_woocommerce']['is_connected']`

haszari: Or perhaps it's better to use something simpler like `$data['extensions']['facebook']['connected']`?

haszari: Updated to the simpler `$data['extensions']['facebook']['connected']` for now. This means we're land-grabbing `[extensions][facebook]` but I think that's appropriate, since this is the primary Facebook extension for WooCommerce. 8f149a5dc88a73976b838f5433f0b5cdbdb0ac50

includes/Utilities/Tracker.php
```diff
@@ -0,0 +1,58 @@
...
+		if ( ! $connection_handler ) {
+			return $data;
+		}
+
+		if ( ! isset( $data['extensions'] ) ) {
+			$data['extensions'] = array();
+		}
+
+		$connection_is_happy = $connection_handler->is_connected() && ! get_transient( 'wc_facebook_connection_invalid' );
+		$data['extensions']['facebook']['connected'] = $connection_is_happy;
```
deltaWhiskey, line 54: Instead of `facebook`, I think the plugin's directory name (`facebook-for-woocommerce`) is better because it is more distinctive. This could avoid naming collisions with other Facebook-related plugins in the future.

haszari: I was thinking the same thing. I think I had an irrational fear of `-` in keys, which you have cured 😄

class-wc-facebookcommerce.php
haszari: > start using just the WooCommerce namespace without SkyVerge 

Can you add an issue for that? Or just add to retro board. It's bugging me too :)

includes/Utilities/Tracker.php
haszari: Now back to `extensions_facebook-for-woocommerce_is-connected` which is more explicit and more consistent (with plugin slug).

includes/Utilities/Tracker.php
haszari: (I had to rebase / recreate this PR and the aforementioned commit is no longer on this branch.)

haszari: https://github.com/fbtransfer/facebook-for-woocommerce/pull/1817#discussion_r601667816

haszari:
This is good to go now - all feedback and follow ups have been addressed. The accuracy of `is_connected` may not be perfect - for example situations where site has become disconnected or had trouble reconnecting. This is a good start, we can iterate more if needed.

I'm holding off merging until I've seen a true and false value logged. I've set up two test sites running a zip build of this branch. One has never connected to Facebook, the other was connected already and then "upgraded" plugin to the zip build. Both sites are opted into sharing usage data.

In 24+ hours these should have both sent a snapshot, and I'll confirm the `is_connected` value is as expected, then merge 🎉 

Thanks all for feedback & reviews 😀

haszari:
Switched this to using standard `yes | no` format in 6cc7ea0b1150e564cbcc46cd93a6a26698ff83ce. This is more consistent with other boolean tracker props, and is clearer. My disconnected test sites was showing an empty string for disconnected, so this should improve that.

Also in this commit added debug logging back to aid testing on non-local staging sites.

haszari:
Confirmed the props are working correctly. 

Screen Shot 2021-03-31 at 9 40 40 AM
Screen Shot 2021-03-31 at 9 41 35 AM

This is now ready to merge – @rcstr are you ok with including this in 2.3.4 or should we hold off for 2.3.5? I'm happy either way, I think this is safe and low risk.

haszari:
Merging! Let's include in 2.3.4.

Issue #1804, opened 2021-03-15
FATAL ERROR in Google Product Category in woocommerce products


nygaardc:
FATAL ERROR occurs when facebook-for-woocommerce plugin is active and I choose the Google Product Category: Media->Books->Printed Books

- [X] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [X] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
![fatal-error](https://user-images.githubusercontent.com/80683455/111143094-b2181780-8585-11eb-8f09-168f5611ebe0.jpg)

When I activate facebook for woocommerce and edit my product google category, I get a FATAL ERROR.

It is no longer possible for me to sync this product to facebook. And it is not possible to revert back.

##🚶‍♀️ Steps to reproduce
1. Download facebook for woocommerce
2. Edit your product
3. Choose facebook tab and select Google product category.

## ✔️ Expected behavior
I expected that the plugin would allow me to choose google product category without crashing.

## 🗃 Logs

The Woocommerce test page shows all ok.
![status_1](https://user-images.githubusercontent.com/80683455/111144075-dde7cd00-8586-11eb-8e43-195a77393815.jpg)
![status_2](https://user-images.githubusercontent.com/80683455/111144101-e50edb00-8586-11eb-84b4-1a5d93449e23.jpg)

## Possible solution
If I uninstall facebook for woocommerce plugin, the Facebook options are no longer available in the product screen. Therefore no PHP fatal error. BUT then I cannot use the plugin to sync my products. So it's a catch 22.

HOW CAN I SOLVE THIS?

Any ideas are much appreciated. Thanks a lot.

budzanowski:
I have replicated this on the latest ver. Investigating further.

budzanowski:
Dev Note:
Attributes specification is missing from the Google Product Category to Attributes mapping. The majority of the categories are OK but I have identified a small subset ( 155 items ) of categories that would experience the same error:
[cat_mappings_missing.txt](https://github.com/facebookincubator/facebook-for-woocommerce/files/6142666/cat_mappings_missing.txt)

nygaardc:
Wow, thanks for investigating, budzanowski.

Is there any workaround, I can use? 

[...]

PR #1820, opened 2021-03-23
Check parent category for attributes if current category has none.

budzanowski:
Fixes: #1804

This requires and is based on #1810 - if approved before #1810 lands I will merge it there.

In general, if we don't have attributes for the given category we should add them -  but that may be quite problematic. This is a quick workaround that uses the parent category attributes if they are available.

## Changelog
- Fix: Add fallback for missing category attributes.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,10 +1,11 @@
 *** Facebook for WooCommerce Changelog ***
 
 2021.nn.nn - version 2.3.6
- * Tweak - Improve full sync performance. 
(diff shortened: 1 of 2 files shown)


danielbitzer (approved):
LGTM 👍

danielbitzer:
> if approved before #1810 lands I will merge it there

Personally, I'd prefer hold off merging to avoid adding more changes to #1810

budzanowski:
OK I will wait for #1810 first.

PR #1824, opened 2021-03-24
Javascript filename cleanup

danielbitzer:
In an effort to cleanup the plugin's JS in general I've done an initial audit of the filenames.

## Deleted
- `facebook-settings.min.js` There is no un-minified version of this file and this file is not loaded anywhere within the plugin
- `facebook-for-woocommerce-products-admin.min.old.js` It's named "old" and isn't loaded anywhere
- `facebook-for-woocommerce-settings-messenger.js` This isn't loaded anywhere in the plugin

Feedback needed here since it's possible, however unlikely, that some 3rd party code in a different plugin is loading these JS files.

## Renamed

The name of the plugin is used inconsistently in the JS file names. I think it's best we remove all of these unnecessary references because it just makes it harder to quickly find files in the plugin.

- `facebook-for-woocommerce-product-sets-admin.js` to `product-sets-admin.js`
- `facebook-for-woocommerce-products-admin.js` to `products-admin.js`
- `facebook-for-woocommerce-settings-sync.js` to `settings-sync.js`
- `facebook-for-woocommerce-modal.js` to `modal.js`
- `facebook-metabox.js` to `metabox.js`
- `facebook-infobanner.js` to `infobanner.js`

## Inconsistent use of `admin/` directory

Some JS files were located in `/assets/js/admin` and some where located in `/assets/js`. However when looking closer at where these files are loaded they are _all_ loaded only in the admin area. Because of this I have moved all files into the `admin/` directory.

## Testing

Because these changes can broadly affect the UI I suppose we need to generally test that things are still working.

Issue #1725, opened 2020-12-08
Excluding Categories from Sync doesn't work


amandasjackson:
## 🔎 Isolate the bug

- [x ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x ] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
At Marketing> Facebook> Product Sync, when I select a category to exclude and click `save` nothing happens and I see js errors in the console

##🚶‍♀️ Steps to reproduce

1. Go to Marketing> Facebook> Product Sync
2. Open Browser console
3. Select a category to exclude
4. Click `save`
5. See js errors

```
underscore.min.js?ver=1.8.3:2 Uncaught TypeError: Cannot read property 'replace' of undefined
    at Function.v.template (underscore.min.js?ver=1.8.3:2)
    at wp-util.min.js?ver=5.5.3:2
    at i.render (backbone-modal.min.js?ver=4.7.1:1)
    at initialize (backbone-modal.min.js?ver=4.7.1:1)
    at i.h.View (backbone.min.js?ver=1.4.0:2)
    at new i (backbone.min.js?ver=1.4.0:2)
    at Object.success (facebook-for-woocommerce-settings-sync.min.js?ver=2.2.0:1)
    at i (jquery.js?ver=1.12.4-wp:2)
    at Object.fireWith [as resolveWith] (jquery.js?ver=1.12.4-wp:2)
    at x (jquery.js?ver=1.12.4-wp:4)
```

![Screenshot](https://d.pr/i/8vfKSp+)
Screenshot direct link: https://d.pr/i/8vfKSp

## ✔️ Expected behavior
When I select a category to exclude and click save, it should work without errors

## 🗃 Logs

```
`
### WordPress Environment ###

WordPress address (URL): https://testing-all-the-things.mystagingwebsite.com
Site address (URL): https://testing-all-the-things.mystagingwebsite.com
WC Version: 4.7.1
REST API Version: ✔ 4.7.1
WC Blocks Version: ✔ 3.6.0
Action Scheduler Version: ✔ 3.1.6
WC Admin Version: ✔ 1.6.3
Log Directory Writable: ✔
WP Version: 5.5.3
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: ✔
Language: en_US
[...]

PR #1826, opened 2021-03-24
Setup `@wordpress/scripts` package to compile javascript and remove compiled files from git

### LINKED ISSUE #1814: Setup webpack (@wordpress/scripts) for processing javascript
Currently I don't see any configured task manager responsible for minifying JS or compiling ESNext code to browser-compatible JS. This looks like a quick win since `@wordpress/scripts` is feature packed and easy to set up. I'd also suggest we remove minified and source map JS files from Git and generate them only when building a release.

- https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/

danielbitzer:
Fixes #1814 

## This PR
- **Removes compiled files from Git** This is considered best practice. We'll just need to figure out how this affects Sake.
- Cleans up uses of `wp_enqueue_script()` with more consistent asset URLs
- Sets up some very basic `@wordpress/scripts` usage

## Testing

Because these changes can broadly affect the UI I suppose we need to generally test that things are still working.


haszari (commented):
This looks like a great dev-quality-of-life improvement @danielbitzer  - thanks for diving in to this! I totally agree that we should have a clear front-end build process and avoid committing build outputs 🙌 

> Because these changes can broadly affect the UI I suppose we need to generally test that things are still working.

Indeed - the scope/risk is quite extensive here. Since we're still confirming the key flows and how they work, maybe it's safer to defer merging this until we know how efficiently test everything. 

What do you think about pausing this until cooldown, when we'll have more experience with flows and maybe can dedicate some time to swarm black box testing to address the risk?

danielbitzer:
Thanks @haszari! I'm happy either way but I think we should aim to have this PR and #1824 merged before doing any work on the javascript in this plugin. E.g. the jQuery deprecation changes.

I think assuming all the asset file paths are correct we can be confident that everything will work as it did previously.

haszari:
>  I'm happy either way but I think we should aim to have this PR and #1824 merged before doing any work on the javascript in this plugin. E.g. the jQuery deprecation changes.

That makes sense to me - sorting out the front-end build blocks any changes/fixes in front end. Let's prioritise this when we need to fix a bug with front end changes 👌 

> I think assuming all the asset file paths are correct we can be confident that everything will work as it did previously.

That's a pretty safe bet. Could also do some diffing on the old and new minified files - though I'm guessing this will be messy. Even so I think we should do some timeboxed swarm testing when making a fundamental change like this.

haszari:
I'm working on an issue that needs to be fixed in JS/client code: #1725

So I'm going to test this branch and experiment with fixing on a branch off this branch. I agree that we should get this tested and merged as soon as we can so we can unblock more fixes & improvements on the client-side code.

haszari:
I've merged `master` into this branch and resolved the conflicts. I've also retargeted it to `master` branch, assuming that this branch/PR includes all the client code tooling & refactor changes. We can split this into separate PRs if appropriate; I've merged and retargeted so I can test all these interrelated changes together. FYI @danielbitzer

haszari:
Marking as `blocked` temporarily to discourage merge until we've given this a thorough review and test. This should not be included in 2.4.0 😄

package.json
```diff
@@ -11,9 +11,12 @@
     "url": "https://wordpress.org/support/plugin/facebook-for-woocommerce"
   },
   "devDependencies": {
+    "@wordpress/scripts": "^14.0.0",
     "sake": "github:skyverge/sake"
   },
   "scripts": {
-    "generate:category_attribute_json": "php bin/GenerateCategoryAttributeMapping.php"
+    "generate:category_attribute_json": "php bin/GenerateCategoryAttributeMapping.php",
+    "build": "NODE_ENV=production wp-scripts build",
```
haszari, line 19: We should probably reserve `npm run build` for full build for compatibility with WooRelease - i.e. full clean build, with all dependencies (npm, composer etc). Could use `npm run build:client` (or similar) for the client build specifically.

webpack.config.js
```diff
@@ -0,0 +1,33 @@
...
+const legacyAdminFileEntries = {};
+
+legacyAdminFileNames.forEach( ( name ) => {
+	legacyAdminFileEntries[ `admin/${ name }` ] = `./assets/js/admin/${ name }.js`;
+} );
+
+module.exports = {
+	...defaultConfig,
+	entry: {
+		'admin/index.js': './assets/js/admin/index.js',
```
haszari, line 26: Do we need `admin/index.js`? I figure we can stick with the existing multi entry point approach. If we're planning on migrating to a different approach, e.g. one bundled `admin` script, that's fine, but I wouldn't assume that in this PR. Happy to be educated about best approach here :)

Is there a specific reason the existing entry points are "legacy"? I think this is reasonable, so I'm happy to stick with that approach (i.e. rename `legacyAdminFileEntries` => `adminFileEntries` etc).

haszari (commented):
Gave this a quick review and test to see if there are any obvious problems. Good news - it's looking great so far 🚀! 

Thanks for taking this on @danielbitzer, it's essential to unblock further improvements to the plugin.

### What's changed
Summarising what's changed as this also PR includes changes from #1824.

- File names cleaned up:
  - Remove redundant `facebook-for-woocommerce`.
  - Move all js to admin, as it all seems to be admin-related.
- Deleted some potentially unused js:
  - `facebook-settings`
  - `products-admin-old`
  - `settings-messenger`
- Add `wp-scripts` tooling and relevant config so can build assets. Remove committed minified versions. ❤️ 
- Tweak various `wp_enqueue_script` to load the newly built assets. Also added a function to return assets path.

This all looks great. The only possible concern is that the deleted files are actually useful and supposed to do something – maybe these should be enqueued (i.e. that's a bug). We'll need to do more testing to be sure!

### How I tested
I haven't tested extensively, in particular I haven't tested messenger at all yet. I did a quick pass to ensure the tooling is working and there are no obvious asset 404s. We'll need to broadly test the main flows and do some exploratory testing to be sure. 

I tested:

- Confirmed no asset 404s on all Facebook related admin pages. I didn't extensively test product editing, I believe there's some behaviour there.
- Confirmed no asset 404s on shop page, add to cart, view cart, view checkout. There's lots more to test on front end, e.g. messenger, pixel events (e.g. purchase).
- Tested a debug (`npm start`) and release build (`npm run build`). Everything worked, could make changes in the JS and add breakpoints.

## Next steps
I've added some comments about some details - none are blocking. 

In particular I'd like to better understand if we are proposing a change to entry points (`admin/index.js`), or just improving the tooling for the existing (~~legacy~~) approach - I'd prefer the latter.

The main blockers here are:

1. Extensive testing. Since we're still getting familiar with all the flows, this is tricky. We could look at doing some extensive testing ourselves after v2.4.0 and then in a v2.5.0 RC these changes could be tested by others. 
2. Confirming that the deleted files are no longer needed. If there's anything we can do to be more sure these are no longer needed, let's do it - e.g. looking at history, or looking at what the deleted code is trying to do.

Hopefully we can derisk both of those with a bit of effort after 2.4.0 - obviously we'll need to timebox but might be a good opportunity to explore the features as a merchant/shopper.

haszari:
Bonus! This branch is bug-for-bug (#1725) compatible with `master`. I can trigger the same JS error as #1725 with this branch too 🤣 

Screen Shot 2021-04-12 at 2 30 29 PM

haszari:
I did a little digging to see when unminified `facebook-settings.js` was removed.

Looks like this was part of 2.0.0 - https://github.com/woocommerce/facebook-for-woocommerce/commit/566664d9d610335117811c7a19c8c55c66385278

The [deleted file](https://github.com/woocommerce/facebook-for-woocommerce/blob/df14eb03de63ba8b69efe62b4292eef8da445f7e/assets/js/facebook-settings.js) does indeed look old. Likely superseded by `settings-sync` and other new files. Also since this release is a while ago, if it was doing something important hopefully that would have been noticed & fixed by now.

haszari:
Added this to 2.5.0 - let's do whatever we need to do to get this on the 🚢

haszari:
Removing `blocked` label - I don't think there's anything specific blocking this. Will need good testing but we can do that as part of 2.5.0 RC.

package.json
danielbitzer: Good call. Renamed to `build:assets` (same as Woo core) in 6b5cf87786e8ececce1747056b6a6cbfe1684c6b.

webpack.config.js
danielbitzer: The multiple entry loaded by different `wp_enqueue_script()` calls approach doesn't seem that practical to me. WooCommerce is becoming a fully JS powered interface where separate pages are loaded by JS rather than a new PHP admin request. Meaning there's no opportunity to load separate scripts on different pages by PHP. 

I'll ask for some input from our JS experts in Slack.

"Legacy" as in jQuery dependant rather than React and vanilla JS.

haszari: Sounds reasonable - I don't think this is a blocker at all. If you are proposing a change or improvement - would be good to get more buy in. At face value, moving to a single admin js and front-end js entry point seems like a good plan, but I'm curious to hear forward-looking best-practice recommendations from core, blocks, an wp-admin teams :)

danielbitzer:
I've just tweaked this PR because we'll need to build our assets before we use Sake to deploy or build a ZIP. To handle this I've added two scripts:
- `npm run deploy:sake`
- `npm run build:zip`

Both of these run `npm run build:assets` before deploying of building a ZIP. Building the ZIP works well but I haven't tested deploying for obvious reasons :)

README.md
```diff
@@ -25,8 +25,7 @@ This plugin uses a custom build tool called [`sake`](https://github.com/skyverge
 
 If you have `sake` set up on your system, these commands can be used to generate a production build.
 
-- `npx sake build` builds to `/build/facebook-for-woocommerce`.
```
danielbitzer, line 28: I'm not sure if it's worth setting up a script for this also?

haszari: Yes I think it's worth it - `package.json` is a good place to document the essential workflow scripts (alongside actual docs). Note this may still need a global install of `sake` on dev machine. Feel free to add all these while you're there, and update readme!

danielbitzer: What's the purpose of having a script that build the plugin into a different directory though? This may encourage people to manually ZIP it when doing releases which we should discourage because it often results in unwanted (.DS_Store) files ending up in the release. Not a big issue though since it's listed right next to `npx sake zip`.

haszari: Agreed, the zip command is the more important one. I forget why I documented `sake build` specifically, maybe just to clarify that it _isn't_  the primary build command!?

haszari: I think I remember now - when you check out this repo, you need to build the code somehow. This was my stab at documenting how to do that with sake (without deploying!). This will change when the PR merges - feel free to remove `npx sake build` if it's not a core dev flow :)

danielbitzer: Ok cool. After this PR `npm start` should be used which you have already documented in the README.md 😄

haszari: `npm start` .. ? never heard of it 🤣

danielbitzer:
This all seems to be working well! Requesting a re-review because I need an approval before merging.

Goodbye minified and sourcemap JS files 👋

webpack.config.js
danielbitzer: I agree it's not a blocker. Most likely we can re-visit when we starting doing some more JS related work for this plugin.

sake.config.js
```diff
@@ -13,4 +13,7 @@ module.exports = {
 	framework: 'v5',
 	deployAssets: false,
 	autoload: true,
+	paths: {
+		js: false
```
haszari, line 17: Would be good to document the intention here, it's js file so you can use comments :)

webpack.config.js
```diff
@@ -0,0 +1,33 @@
+// Load the default @wordpress/scripts config object
+const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
+
+const legacyAdminFileNames = [
```
haszari, line 4: Can comment that these are legacy because they use jQuery.

webpack.config.js
haszari: I see the (different) `admin.js` approach is still there, so can you please document the proposed change, or stick with the multi entry point approach? As it stands, you're sneaking in a change in how JS assets are structured, without documenting it.

I see you got some input from the team - great - but since you're changing this, you need to document your design/intention here. Or if you're on the fence, build individual js files as before.

haszari (approved):
Pre-approving but I've left a couple of bits of feedback that may result in changes. In general this is superb and a really important improvement - we need to be able to iterate on our front-end code! 

- `--deployAssets=0` hard-coded option - remove if possible.
- Production build - how confident are you this will work well? I'm somewhat confident, but I've never released with `sake`. If we can really test it or do some dry runs great. At a minimum, let's chat about how to handle 2.5.0 release - if there's any risk or things I should look out for.
- Change to assets/entry points - either document the proposed improvement or keep the same assets/entry points as before. 

I really wanted to merge this today but I want more clarity/confidence about the `sake` / release implications. Might be a good idea to get a review from someone with a bit more experience releasing the plugin - @budzanowski @layoutd perhaps.

package.json
```diff
@@ -11,9 +11,14 @@
...
   "devDependencies": {
+    "@wordpress/scripts": "^14.0.0",
     "sake": "github:skyverge/sake"
   },
   "scripts": {
-    "generate:category_attribute_json": "php bin/GenerateCategoryAttributeMapping.php"
+    "generate:category_attribute_json": "php bin/GenerateCategoryAttributeMapping.php",
+    "build:assets": "NODE_ENV=production wp-scripts build",
+    "start": "wp-scripts start",
+    "deploy:sake": "npm run build:assets && npx sake deploy --deployAssets=0",
```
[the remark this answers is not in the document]
haszari: Update: I see this option is disabled in `sake.config.js`: `deployAssets: false`.

So I suspect this is fine, but there's no need to hard-code or force the option in the npm script - since it's in the config.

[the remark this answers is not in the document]
danielbitzer: > Is --deployAssets=0 needed, what does it do?

It's included in the steps listed here: https://github.com/woocommerce/facebook-for-woocommerce/wiki/Build-&-Release#deploy-via-sake

I'm not sure why this param is included though 🤔

sake.config.js
danielbitzer: Added in 02d753af5c1c0288877ffd87514e279eb17ca4d6

webpack.config.js
danielbitzer: Yep, I tweaked the naming and added some comments in 5fa94a1c420f13ad09f9dbb565428104786a2592.

webpack.config.js
danielbitzer: Fair point, I'm happy to remove this and we can revisit this discussion when we need to add some React based JS to the plugin. I've made a few tweaks in 5fa94a1c420f13ad09f9dbb565428104786a2592.

webpack.config.js
```diff
@@ -0,0 +1,35 @@
...
+
+jQueryUIAdminFileNames.forEach( ( name ) => {
+	jQueryUIAdminFileEntries[ `admin/${ name }` ] = `./assets/js/admin/${ name }.js`;
+} );
+
+module.exports = {
+	...defaultConfig,
+	entry: {
+		// Use admin/index.js for any new React-powered UI
+		'admin/index.js': './assets/js/admin/index.js',
```
haszari, line 28: Spotted an issue with this when testing #1947. The entry point config maybe should be `index` (no `.js` extension); the generated file is `index.js.js` 😆 

Screen Shot 2021-05-13 at 3 25 18 PM

danielbitzer: Oops, will open a PR to fix

haszari:
Looks like we have a JS regression - #1979. I'm working on it, hopefully easy to reinstate!

Issue #1825, opened 2021-03-24
Undefined array key error


WillBrubaker:
## 🔎 Isolate the bug

- [x ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x ] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
As reported in https://wordpress.org/support/topic/php-warning-after-update-v-2-3-3/

>PHP Warning: Undefined array key 0 in /var/www/html/wp-content/plugins/facebook-for-woocommerce/includes/fbproduct.php on line 565

##🚶‍♀️ Steps to reproduce
Presumably try to sync a product to Facebook that hasn't got an image url

## ✔️ Expected behavior
Some logic to check if the value is populated before trying to assign

budzanowski:
I have identified that this happens when the product image is not set but the product has some images in the galery.

budzanowski:
Closed with #1828

PR #1828, opened 2021-03-24
Regenerate indexes after filtering.

budzanowski:
Fixes: #1825

If the product has no image set but the product gallery has images we may end up with URL array that has no index '0'. Parts of the code expect something to always be present at the index `0`. Regenerating indexes solves this issue.

## Testing:
1. Create a product with no image set but with images in the gallery:
image
2. Sync the product.

Outcome: without this PR the error from #1825 is logged. With this PR the error is not logged and a valid image URL is appended.

## Changelog:
- Fix: Undefined array key error for products without 'Product image' set.

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -198,6 +198,11 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 			$image_urls = array_merge( $image_urls, $this->get_gallery_urls() );
 			$image_urls = array_filter( array_unique( $image_urls ) );
 
+			// Regenerate indexes after filtering.
+			// The array_filter does not touches indexes so we may end up with gaps.
+			// Later parts of the code expect something to exist under the 0 index.
+			$image_urls = array_values( $image_urls );
+
 			if ( empty( $image_urls ) ) {
 				// TODO: replace or remove this placeholder - placeholdit.imgix.net is no longer available {WV 2020-01-21}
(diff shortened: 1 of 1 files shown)


includes/fbproduct.php
```diff
@@ -198,6 +198,11 @@ public function get_all_image_urls() {
 			$image_urls = array_merge( $image_urls, $this->get_gallery_urls() );
 			$image_urls = array_filter( array_unique( $image_urls ) );
 
+			// Regenerate indexes after filtering.
+			// The array_filter does not touches indexes so we may end up with gaps.
+			// Later parts of the code expect something to exist under the 0 index.
+			$image_urls = array_values( $image_urls );
```
haszari, line 204: Aha, this is an example of the infamous [PHP unordered numeric array quirk](https://www.oreilly.com/library/view/php-cookbook-3rd/9781449363741/ch04.html)! 

So to clarify, you're regenerating the array numeric indices here - correct?

haszari (approved):
Tested and confirmed the fix - product syncs with no error in logs (I reproduced the error on `master` first to confirm). 🎉 

Fix looks totally safe and robust to me. Could you tweak the comment and PR title to clarify that this refers to PHP array index – when I first read the PR I thought this might be some database index or cache, e.g. `regenerate thumbnails` or similar.

haszari:
> With this PR the error is not logged and a valid image URL is appended.

Just to clarify, I don't see a product image with this PR - it's a placeholder. This seems consistent with Woo though. 

Screen Shot 2021-03-25 at 11 27 10 AM

Screen Shot 2021-03-25 at 11 28 53 AM

includes/fbproduct.php
budzanowski: Yes, array numeric indices.

Issue #1829, opened 2021-03-24
Products without image have no default image.


budzanowski:
The URL that is used for the product default image in the absence of `Product Image` setting is not valid anymore:

```PHP
			if ( empty( $image_urls ) ) {
				// TODO: replace or remove this placeholder - placeholdit.imgix.net is no longer available {WV 2020-01-21}
				$image_urls[] = sprintf( 'https://placeholdit.imgix.net/~text?txtsize=33&name=%s&w=530&h=530', rawurlencode( strip_tags( $this->woo_product->get_title() ) ) );
			}
```

from https://github.com/fbtransfer/facebook-for-woocommerce/blob/master/includes/fbproduct.php#L201-L204

budzanowski:
Closed with #1830

PR #1830, opened 2021-03-24
Add default image placeholder for products without image.

budzanowski:
Fixes: #1829

Because `https://placeholdit.imgix.net` does not work anymore we were sending broken image URLs to the Facebook catalog. This resulted in an empty product image:
image

This fixes it by adding our default image placeholder and linking it with the product:
image

The text in image functionality of `placeholdit` was nice but I don't think that we need to replicate it in any way. Just having a placeholder is way better than having nothing.

For now, I don't give any way to customize it - if people will ask then we can add it.

Why not use use `wc_placeholder_img_src`? Because we lose control over the size of the image. Even if we specify the size option with the function call this can be overwritten by themes and filters. This could potentially lead to problems with the product catalog as Facebook specifies the min and max image size.

## Testing

1. Create a product ( add no images ) using master branch and sync.
2. Switch to this branch, create a product ( add no images )
3. Compare results

## Changelog
- Fix - Add default placeholder for products with no image set.


includes/fbproduct.php
```diff
@@ -199,8 +199,7 @@ public function get_all_image_urls() {
 			$image_urls = array_filter( array_unique( $image_urls ) );
 
 			if ( empty( $image_urls ) ) {
-				// TODO: replace or remove this placeholder - placeholdit.imgix.net is no longer available {WV 2020-01-21}
-				$image_urls[] = sprintf( 'https://placeholdit.imgix.net/~text?txtsize=33&name=%s&w=530&h=530', rawurlencode( strip_tags( $this->woo_product->get_title() ) ) );
+				$image_urls[] = plugin_dir_url( dirname( __FILE__ ) ) . 'assets/images/woocommerce-placeholder.png';
```
danielbitzer, line 202:
```suggestion
				$image_urls[] = facebook_for_woocommerce()->get_plugin_url() . '/assets/images/woocommerce-placeholder.png';
```
I feel like this is simpler...

danielbitzer (approved):
Looks good! Works as expected.

PR #1840, opened 2021-03-26
Optimise memory usage for whole catalog sync scenario.

### LINKED ISSUE #1839: 2.3.3 update routine breaking WP Admin on stores with a high number of products
Internal incident report: pb0Spc-1tW-p2

## ✍️ Describe the bug

Immediately after updating to [version 2.3.3 ](https://github.com/woocommerce/facebook-for-woocommerce/pull/1797) we have had reports of fatal memory exhaustion errors appearing in WP Admin. In these cases it could only be fixed by deactivating the plugin via WP CLI or FTP.

_I have not done any debugging on the affected sites but based on the evidence I’m very confident of the issue…_ I'm currently generating products on my local site. I currently have 1618 products which is using 134 MB of memory to start the catalog sync.

The 2.3.3 update triggers a full product catalog sync to Facebook in the background immediately when a site updates. Note that this is something that happens normally in the plugin. There is a button to trigger a sync: 

<p>This plugin uses SkyVerge's plugin framework to handle the update routine which is hooked to the  <code>wp_loaded</code> action on admin requests (excluding admin-ajax) (<a href="https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/Lifecycle.php#L89">code ref</a>). If the routine throws an error it will continue trying to run the routine on all subsequent admin requests until the plugin is deactivated.</p>

<p>The issue with the product sync code in Facebook for WooCommerce is that it <a href="https://github.com/woocommerce/facebook-for-woocommerce/blob/7af84f0570939eed9d0aad607e27186eae93ce2f/includes/Products/Sync.php#L82-L89">uses an unbounded product query</a> to firstly retrieve every product and then attempts to <a href="https://github.com/woocommerce/facebook-for-woocommerce/blob/7af84f0570939eed9d0aad607e27186eae93ce2f/includes/Products/Sync.php#L176-L187">write a background job for <strong>every product</strong></a> in the same request. Obviously it was just a matter of time until a site has too many products to process in a single request.</p>

@kosiew can also replicate a very similar issue which is reported in #1818

## The solution

We need to remove use of unbounded product database query.

I've observed that there are 2 ways to sync products in this plugin, via **a feed** or **a background job**. Perhaps these do the same thing? If that was the case perhaps one is more scalable and should be used in instances like this. I've asked in slack. If anyone knows please comment.

## Steps to reproduce

- Install [Query Monitor](https://wordpress.org/plugins/query-monitor/)
- Generate a ton of products on your store (use [woocommerce/wc-smooth-generator](https://github.com/woocommerce/wc-smooth-generator)) the exact amount will vary on the server
- Set the `wc_facebook_for_woocommerce_version` option to `2.3.2`
- Load an admin page which will immediately trigger the update routine 
- Check out your memory usage and how many queries the plugin is doing

Screen Shot 2021-03-26 at 3 01 37 pm

## ✔️ Expected behavior
Plugin updates don't kill the WP Admin area.

## 🗃 Logs

TBA

budzanowski:
Partially fixes #1839

This refactors the code that is used to prepare the synchronization of all products to the FaceBook catalog.

### What has been changed.

We split the main query into two queries. One for `product` and one for `product_variation`. For the first, we can optimize the flow and get the `ids` directly. For the second one, the args remain unchanged. What does that give us?

We don't need to test the product for the `product_variation` type - we already have separate arrays for each product type.

We also don't need to test if the parent product is in state `publish` or not. We have all `publish` products from the first query - we need to just check if our variation product parent is in the array.

This significantly reduces the required memory on my machine. For a catalog of ~1000 products and ~6500 variations, the request was originally using around ~110MB. After this change was introduced the request was using ~31MB which is only 2MB higher compared to page load when the mass sync procedure is not triggered.

I have also adjusted the ids generation code in the Feed file. This is the same function but written in a totally different way - very interesting. Now we have the same code responsible for those two places.

### DIFF
--- a/includes/fbproductfeed.php
+++ b/includes/fbproductfeed.php
@@ -892,6 +892,9 @@ if ( ! class_exists( 'WC_Facebook_Product_Feed' ) ) :
 		}
 
 		public function get_product_wpid() {
+
+			wc_deprecated_function( __METHOD__, '2.3.6', '\\WC_Facebookcommerce_Utils::get_all_product_ids_for_sync()' );
+
 			$post_ids = WC_Facebookcommerce_Utils::get_wp_posts(
 				null,
 				null,
--- a/includes/fbproductfeed.php
+++ b/includes/fbproductfeed.php
@@ -446,29 +446,7 @@ if ( ! class_exists( 'WC_Facebook_Product_Feed' ) ) :
 		 * @return int[]
 		 */
 		private function get_product_ids() {
-
-			$post_ids = $this->get_product_wpid();
-
-			// remove variations with unpublished parents
-			$post_ids = array_filter( $post_ids,
-				function ( $post_id ) {
-					return ( 'product_variation' !== get_post_type( $post_id )
-					         || 'publish' === get_post( wp_get_post_parent_id( $post_id ) )->post_status );
-				}
-			);
-
-			$all_parent_product = array_map(
-				function( $post_id ) {
-					if ( 'product_variation' === get_post_type( $post_id ) ) {
-						return wp_get_post_parent_id( $post_id );
-					}
-				},
-				$post_ids
-			);
-
-			$all_parent_product = array_filter( array_unique( $all_parent_product ) );
-
-			return array_diff( $post_ids, $all_parent_product );
+			return \WC_Facebookcommerce_Utils::get_all_product_ids_for_sync();
 		}
 
 
--- a/includes/fbutils.php
+++ b/includes/fbutils.php
@@ -434,6 +434,52 @@ if ( ! class_exists( 'WC_Facebookcommerce_Utils' ) ) :
 			return $term && $term->name ? $term->name : $default_value;
 		}
 
+		/**
+		 * Get all products for synchronization tasks.
+		 *
+		 * @return array IDs of all product for synchronization.
+		 */
+		public static function get_all_product_ids_for_sync() {
+			// Get all published products ids. This includes parent products of variations.
+			$product_args = array(
+				'fields'         => 'ids',
+				'post_status'    => 'publish',
+				'post_type'      => 'product',
+				'posts_per_page' => -1,
+			);
+			$product_ids  = get_posts( $product_args );
+
+			// Get all variations ids with their parents ids.
+			$variation_args     = array(
+				'fields'         => 'id=>parent',
+				'post_status'    => 'publish',
+				'post_type'      => 'product_variation',
+				'posts_per_page' => -1,
+			);
+			$variation_products = get_posts( $variation_args );
+
+			/*
+			* Collect all parent products.
+			* Exclude variations which parents are not 'publish'.
+			*/
+			$parent_product_ids = array();
+			foreach ( $variation_products as $post_id => $parent_id ) {
+				/*
+				* Keep track of all parents to remove them from the list of products to sync.
+				* Use key to automatically remove duplicated items.
+				*/
+				$parent_product_ids[ $parent_id ] = true;
+
+				// Include variations with published parents only.
+				if ( in_array( $parent_id, $product_ids ) ) {
+					$product_ids[] = $post_id;
+				}
+			}
+
+			// Remove parent products because those can't be represented as Product Items.
+			return array_diff( $product_ids, array_keys( $parent_product_ids ) );
+		}
+
 
 		/*
 		* Change variant product field name from Woo taxonomy to FB name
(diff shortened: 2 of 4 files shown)


includes/fbutils.php
```diff
@@ -429,6 +429,52 @@ public function get_variant_option_name( $wp_id, $label, $default_value ) {
...
+
+				// Include variations with published parents only.
+				if ( in_array( $parent_id, $product_ids ) ) {
+					$product_ids[] = $post_id;
+				}
+			}
+
+			// Remove parent products because those can't be represented as Product Items.
+			return array_diff( $product_ids, array_keys( $parent_product_ids ) );
+		}
```
haszari, line 476: This refactored algorithm looks good to me. Reading through the 2x old logic, I believe this is an improvement – single code path for both feed and bulk sync, and the queries are all front-loaded.

haszari (commented):
This looks like a great improvement @budzanowski . 

For now I'm thinking let's finalise what other improvements we plan to make (e.g. batch updates - #1841). Then we can incorporate this optimisation or tweak to suit. This also might give us more opportunity to extensively test the changes and ensure robust performance.

haszari:
FYI @budzanowski I haven't tested this yet - ran out of time today. I plan on testing this and other performance improvements using a standard profiling tool / approach - keen to hear ideas for how best to measure performance. Would be great to get consistent something in place that we can all use on local and on various test environments/hosts.

Tomorrow I'll test this using basic query monitor profiling, see #1845 for info.

galapogos01:
Happy to test this on my ~3500 product site if you can provide a build.

haszari (approved):
Tested this branch and master using the profiling from #1845 . My local test site has ~1000 products, generated by [smooth generator](https://github.com/woocommerce/wc-smooth-generator).

This is how I tested (for each branch):

- Merge an ad-hoc branch from test branch + `try/query-monitor-sync-profiling`, update to that branch.
- Delete/clear all logs.
- Trigger a manual sync from `Facebook > Product sync > Sync products` button.
  - Note I didn't test any other kinds of sync or feed generation. 
- Observe logs.

In both cases, the sync took around 10-15 min to complete (based on the countdown in the `Product sync` tab).

The mem use and time spent for the background sync jobs was similar (as expected). The time and mem use for collating the product IDs to sync was vastly improved on this PR branch:

- 5 seconds => 0.02 seconds
- 81 mb => 6 mb

```
2021-03-31T00:51:00+00:00 DEBUG qm_timing facebook_for_woocommerce__create_or_update_all_products: time=4.974956035614 sec; mem=81 MB
2021-03-31T03:34:05+00:00 DEBUG qm_timing facebook_for_woocommerce__create_or_update_all_products: time=0.025023937225342 sec; mem=6 MB
```

That's a great improvement, nice work @budzanowski !

We'll need to do more testing on this and other improvements as we get closer to release, but this PR's good to go IMO 🚀

haszari:
> Happy to test this on my ~3500 product site if you can provide a build.

Thanks for offering to help out @galapogos01 ! We can look at building an RC or beta when we're closer to release. cc @budzanowski

includes/fbutils.php
```diff
@@ -429,6 +429,52 @@ public function get_variant_option_name( $wp_id, $label, $default_value ) {
 			return $term && $term->name ? $term->name : $default_value;
 		}
 
+		/**
+		 * Get all products for synchronization tasks.
+		 *
+		 * @return array IDs of all product for synchronization.
```
danielbitzer, line 435: I think it's worth adding a warning here to future devs. Something like "If you make any changes to this method please check it will still scale for stores with a massive number of products."

includes/fbproductfeed.php
```diff
@@ -446,29 +446,7 @@ public function sync_all_products_using_feed() {
 		 * @return int[]
 		 */
 		private function get_product_ids() {
-
-			$post_ids = $this->get_product_wpid();
```
danielbitzer, line 450: It's worth deprecating this `\WC_Facebook_Product_Feed::get_product_wpid` method if we're no longer using it.

danielbitzer (approved):
**Before:** 258 MB, 23699 queries, 13.4 seconds
**With this PR:** 28 MB, 112 queries, 1 second

That's an impressive improvement @budzanowski!

On a product with a massive

galapogos01:
> > Happy to test this on my ~3500 product site if you can provide a build.
> 
> Thanks for offering to help out @galapogos01 ! We can look at building an RC or beta when we're closer to release. cc @budzanowski

Did a quick test with this PR and found there was no improvement for the AS sync task. https://github.com/woocommerce/facebook-for-woocommerce/issues/1818#issuecomment-812244863

Happy to provide any other debug logs etc.

Issue #1842, opened 2021-03-29
Temporarily disable full sync on 2.3.3 upgrade


haszari:
We've had reports (#1839) of full product sync (triggered by 2.3.3 upgrade - #1808) causing issues, especially on sites with large product catalogue.

This revealed a performance issue with product sync code that could affect any full sync (e.g. triggered manually from UI, generating product feed file). In the medium term we are looking at how to improve full sync - #1841. Some sites have disabled Facebook for WooCommerce; the goal is to ensure it's safe for those large sites to upgrade and re-enable Facebook for WooCommerce.

In the short term we need to disable the risky full sync to prevent this from impacting sites. Note this will likely be temporary until we have a more robust full sync.

haszari:
Looking back to 2.3.3 release, it was understood that we needed to sync all products as part of updating the plugin. If we disable that due to performance concern, I'd like to understand any sync / functionality implications. 

@denniskwon  Can you comment on this - if we disable full sync on update, how will that affect product attribute syncing? Can you give an example of how this would affect a site's catalogue on Facebook? Thank you!

denniskwon:
@haszari - The main reason to force the full sync was to apply the other attribute related changes shipped as part of 2.3.3. If the full sync does not happen, it means any already synced products (presumably most of the products in an update scenario) would not be synced again, until they are edited. As such, all of those products may be missing values for attributes sync as color, size, etc.

budzanowski:
Closed with #1844.

PR #1844, opened 2021-03-29
Temporarily disable sync on update

budzanowski:
This disables the full catalog sync request that is causing problems on sites with big products catalog.
This will be re-introduced soon after we will optimize the synchronization flow.

Closes: #1842

### DIFF
--- a/includes/Lifecycle.php
+++ b/includes/Lifecycle.php
@@ -42,7 +42,6 @@ class Lifecycle extends Framework\Plugin\Lifecycle {
 			'2.0.0',
 			'2.0.3',
 			'2.0.4',
-			'2.3.3',
 		];
 	}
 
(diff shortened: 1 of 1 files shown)


rcstr (approved):

haszari (approved):
Looks good to me - essentially a straight revert of the upgrade sync job we added 😁 

I tested this by resetting the version option `wc_facebook_for_woocommerce_version=2.3.2` and loading the `Marketing > Facebook > Product Sync` page. I used query monitor and logs to confirm whether the upgrade job happened. Also used [custom query monitor logging](https://github.com/woocommerce/facebook-for-woocommerce/pull/1845) (merged into each branch before test) to profile background job.

I repeated the test twice, once with `master` (~ 2.3.3) and once with this branch.

- 2.3.3 - The upgrade job ran, page load mem use was ~30 mb, logged mem use was ~102 mb for background sync.
- this branch - The upgrade job did not run, page load mem use was ~24 MB.

So that confirms this is working correctly. 

#### Impact - no upgrade sync => no improved product attributes

The other consideration here is if stores are impacted by not having a full sync. 

As far as I understand it, this means that the attributes in the Facebook catalog will not be updated correctly (based on [this comment](https://github.com/woocommerce/facebook-for-woocommerce/issues/1842#issuecomment-809531959) from @denniskwon).

That seems fine for now, as the impact of the performance issue for larger sites is much more serious. It may be possible for those sites to use background sync (click the button in UI) to get attributes updated (I don't know if this could cause similar performance issues).

#### Merge & ship
I say yes!

PR #1845, opened 2021-03-30
add query monitor profiling to three specific product-sync code paths

haszari:
This PR adds ~~[Query Monitor profiling](https://github.com/johnbillion/query-monitor#profiling)~~ profiling to three aspects of product sync. (Now switched to use the profiler from partner PR #1854.)

- Get product ID list for background batch sync / feed `facebook_for_woocommerce__create_or_update_all_products`
- Process background batch sync `facebook_for_woocommerce__background_product_sync__process_job`
- Feed file generation `facebook_for_woocommerce__generate_feed`

~~We may not end up merging this PR. It's intended for helping us measure & optimise product sync. I'm opening this PR for visibility only for now. If this proves useful (and is harmless in production), we could consider merging in future (or something similar).~~ Update: keen to get this shipshape and merged. **When reviewing please consider implications of adding this to the plugin.**

To test:

- [Ensure profiler logging is enabled](https://github.com/woocommerce/facebook-for-woocommerce/tree/master/includes/Debug) `define( 'FACEBOOK_FOR_WOOCOMMERCE_PROFILING_LOG_ENABLED', true );`. I also had debug mode active: `Marketing > Facebook > Connection tab > Log plugin events for debugging`.
- Go to `Marketing > Facebook > Product sync` tab. 
- Click `Sync products` to trigger a sync.
- View logs, e.g. `WooCommerce > Status > Logs`. There should be a `facebook_for_woocommerce_profiling_*` log.

Should see stuff like this in the log file:

```
2021-04-08T02:54:05+00:00 DEBUG create_or_update_all_products - Memory: 6,334.59 KB, time: 0.02s.
2021-04-08T02:54:32+00:00 DEBUG background_product_sync__process_job - Memory: 131,754.52 KB, time: 26.41s.
2021-04-08T02:54:54+00:00 DEBUG background_product_sync__process_job - Memory: 16,114.16 KB, time: 21.52s.
```


haszari:
FYI @budzanowski - I wanted a standard way to evaluate performance improvements such as #1840. Keen for your feedback on this approach and other ideas for how best to measure 📊

budzanowski:
I would go with this approach. It is simple and does not require a lot of alterations in the code. At the same time, it may be quite effective. We could look for other approaches but only if this would not prove to be enough.

danielbitzer:
This is cool! I didn't know Query Monitor had this profiling feature. However, I think there's probably a simpler solution than hacking a QM function to log execution time and memory usage and I wouldn't merge this.

danielbitzer:
> keen to understand our options for profiling and get something in place that we can use as standard

In this example you're are logging memory usage and execution time which can be done more simply with PHP native functions. (`memory_get_usage()`, `microtime()`).

haszari:
> In this example you're are logging memory usage and execution time which can be done more simply with PHP native functions. (memory_get_usage(), microtime()).

Aha! Thanks @danielbitzer, TIL 😁 

Sounds like we could pretty quickly put something together native to the plugin using existing logging & standard APIs 👌 – something to consider @budzanowski.

haszari:
Closing this in favour of native implementation in #1854.

haszari:
Reopening - will repurpose this using the goodness from #1854.

includes/fbproductfeed.php
```diff
@@ -89,6 +89,10 @@ public function schedule_feed_generation() {
 		 * @since 1.11.0
 		 */
 		public function generate_feed() {
+			$profiling_logger = facebook_for_woocommerce()->get_profiling_logger();
+			$profiling_logger->start( 'generate_feed' );
+
+			do_action( 'qm/start', 'facebook_for_woocommerce__generate_feed' );
```
haszari, line 95: There's also some other timing code in this function - see `set_feed_generation_time_with_decay`  on line 107. We should consider combining these into one, and/or see if that feed generation time data is still being used.

rcstr (approved):
awesome! logs looks good :shipit: 

Screen Shot 2021-04-08 at 18 15 15

haszari:
Merging! @budzanowski if you have any feedback or follow-ups let me know or open an issue 🕺

includes/fbproductfeed.php
danielbitzer, line 95: @haszari shouldn't this Query Monitor start action have been removed from this PR?

haszari: Yes it should have! 😱 I'll whip up a quick PR, thanks for catching!

haszari: #1869 🧯 

Whoever reviews that - if it's all good feel free to merge for me 🤞

Issue #1847, opened 2021-03-30
Wrong `Value` Field in AddToCart Events 


abashinfection:
Hello, 

I recently noticed an unexpected behavior related to AddToCart events. 
When triggering multiple AddToCart events from a website, the `value` field contains a cumulative sum of all the previous AddToCart events. 
Based on my understanding, this is due to the following [code](https://github.com/woocommerce/facebook-for-woocommerce/blob/77f964b075aed4fd060cc39db1e7a780c8bfdf8c/facebook-commerce-events-tracker.php#L572):

```php
$product = wc_get_product( $variation_id ?: $product_id );

// bail if invalid product or error
if ( ! $product instanceof \WC_Product ) {
	return;
}
$event_data = [
	'event_name'  => 'AddToCart',
	'custom_data' => [
		'content_ids'  => $this->get_cart_content_ids(),
		'content_name' => $this->get_cart_content_names(),
		'content_type' => 'product',
		'contents'     => $this->get_cart_contents(),
		'value'        => $this->get_cart_total(),
		'currency'     => get_woocommerce_currency(),
	],
	'user_data' => $this->pixel->get_user_info(),
];
```

As per [Facebook documentation](https://developers.facebook.com/docs/facebook-pixel/reference#standard-events), in the case of AddToCart event we have: 
- `value`: The value of a user performing this event to the business.
- `currency`: The currency for the `value` specified.

It seems to me that the actual value for the `value` field should be the **price of the single product, not the cart total**.

This has a significant impact on multiple tools in the Facebook Business Manager platform. 
For example, when monitoring campaign metrics the total value for AddToCart events is much higher than the actual value in WooCommerce carts. 
This is because each new AddToCart event contains the cumulative sum of prices of all products in the cart:
- _First AddToCart_ event: 1 product, value=_X_
- _Second AddToCart_ event: 2 products in cart, value=_X+Y_
- _Third AddToCart_ event: 3 products in cart, value=_X+Y+Z_

The metrics shown by Facebook will sum the value for all the AddToCart events received: 
- _First AddToCart_ + _Second AddToCart_ + _Third AddToCart_ = (_X_) + (_X+Y_) + (_X+Y+Z_)
While the actual metric should simply be _X+Y+Z_. 

In a similar way, also the fields `content_ids` and `content_name` should be only relative to the single product being added to the cart, not the whole cart contents. 

[...]

PR #1848, opened 2021-03-30
AddToCart Events should not include whole cart info

abashinfection:
This PR aims to track progress to address issue #1847

Since I'm really dusty in PHP development, this is only meant to kickstart implementation of the required code changes.
I've updated the AddToCart related routines according to what I believe should be the right behavior but I'm not sure how to retrieve the product info for some of them.

Any help or feedback is really    ###appreciated.


rawdreeg:
Thanks for submitting this @abashinfection ... the issue has been resolved by https://github.com/woocommerce/facebook-for-woocommerce/pull/2215. I will close this PR.

Issue #1849, opened 2021-03-31
2.3.4 Fatal error if WooCommerce Admin not active


seldimi:
When updating to 2.3.4 and you don’t have WooCommerce Admin installed, you get a fatal error:

PHP Fatal error: Uncaught Error: Class ‘Automattic\WooCommerce\Admin\Features\Features’ not found in /home/www/…./wp-content/plugins/facebook-for-woocommerce/includes/Admin/Settings.php:62

iMazed:
This was also reported in 3869417-hc, and can be replicated on a fresh install. Setting to critical.

iMazed:
Also reported on WP.org: https://wordpress.org/support/topic/fatal-error-if-woocommerce-admin-not-active/

jedkirby:
WooCommerce Admin is packaged up within WooCommerce directly, so, this issue is present because the version 2.3.4 relies on at least WooCommerce 5.0.0 > being installed.

WooCommerce introduced the class `Automattic\WooCommerce\Admin\Features\Features` after version 5.0.0, as see below:

WooCommerce v4.9.2: https://plugins.trac.wordpress.org/browser/woocommerce/tags/4.9.2/packages/woocommerce-admin/src/Features
WooCommerce v5.0.0: https://plugins.trac.wordpress.org/browser/woocommerce/tags/5.0.0/packages/woocommerce-admin/src/Features?order=name

To resolve your instance, you'll need to either downgrade Facebook for WooCommerce, or upgrade WooCommerce to 5.0.0 or greater.

seldimi:
Well, that doesn't make sense. There are a lot of websites using WC 4.x or need some more time to be upgraded to 5.x version. 
Also, there was neither notice, nor minimum requirements about depreciating WC 4.x version. 

Lastly, have a look at the plugin requirements
WC requires at least: 3.5.0

You should be more prepared if you want to discontinue WC 4.x

jedkirby:
I totally agree, it's a breaking change. I don't work on this plugin, FYI - i'm just resolving broken WooCommerce website's as well as you.. Hopefully they'll resolve the issue soon, but, in the mean time downgrading as I mentioned above fixes it for the short term.

iMazed:
@jedkirby - the website I'm working on is running WooCommerce 5.1.0 so I'm pretty sure that there's an additional problem going on. Thanks for your input, though!

jedkirby:
[...]

PR #1851, opened 2021-03-31
Fix critical for pre 5.0.0 WC sites.

budzanowski:
Diable the problematic feature for now as we figure out the version limit.

Fixes: #1849


rcstr (approved):
Tested using `WC Version 4.9.2` and was able to duplicate the bug. Updated the PR with a fix that will allow us to keep `wc-admin` nav items

rcstr:
Updated base branch to `release-2.3.5`, merging

PR #1857, opened 2021-04-01
Handle and log connection errors

### LINKED ISSUE #1805: Improve support for diagnosing connection issues - show errors mode
Sometimes the initial connection between the Woo plugin and Facebook fails, with cryptic errors like `Something went wrong`. It might be useful to add an option to log or display more error info to help diagnose what's causing the problem. 

Previously there was an alternative connect server with more logging enabled. This is no longer available; the goal of this issue is to meet that need, i.e. expose more details about connection errors so merchants and support team can easily resolve connection problems.

### Possible solution
- Add an option to plugin e.g. `show errors` or `log errors`. This could be activated via UI, or via code (e.g. filter or wp options). Or potentially just always log errors by default if practical.
- When this option is set, pass it to connect server.
- In connect server, if option is present, capture and return any errors to site.
- Plugin logs or displays errors in site admin UI (could use `WooCommerce > Status > Logs`).

related: 118-gh-Automattic/connect.woocommerce.com

rcstr:
Fixes #1805

### Description
We want to log connection errors, this PR allows `connect.woocommerce.com` to send errors when connection to FB fails and logs them.

### Changes
- defines proxy error (`connect.woocommerce.com`) codes and their messages
https://github.com/woocommerce/facebook-for-woocommerce/blob/2830e71d62cb8a25d0cc910140fac902fa47adc3/includes/Handlers/Connection.php#L121-L135

- Handles connections errors and logs them
- Allows connection errors to be log when debug is disabled **but** the site is not connected (**this is because `debug` mode can't be enabled if the site is not connected**)

### Testing instructions
You need an instance of `https://connect.woocommerce.com/` to test this PR end-to-end, if you don't you can test following these instructions:

- Comment these lines in order to pass _nonce_ validation
https://github.com/woocommerce/facebook-for-woocommerce/blob/2830e71d62cb8a25d0cc910140fac902fa47adc3/includes/Handlers/Connection.php#L289-L291
- Now from your local instance go to URL 'https://HOST/?wc-api=wc_facebook_connect&err=1&err_code=token_permission_missing_ads_read' where HOST is your WC instance URL (`fb-wc.local` in my case)
- This request will be redirected to `/wp-admin/admin.php?page=wc-facebook`
- You'll see an error message on your screen:
Screen Shot 2021-04-14 at 03 29 13

- Now go to `WooCommerce > Status > Log` and select the last one with name `facebook_for_woocommerce-connect` and confirm error is being logged:
Screen Shot 2021-04-14 at 03 32 02

- Test changing the query param `err_code=token_permission_missing_ads_read` with other errors like: `business_manager_id_missing`, `page_id_missing`, `user_access_token_missing` and validate that the message is correct one


budzanowski (approved):

includes/Handlers/Connection.php
```diff
@@ -115,6 +117,22 @@ public function __construct( \WC_Facebookcommerce $plugin ) {
...
+				'catalog_id_missing'                => __( 'Catalog ID is missing', 'facebook-for-commerce' ),
+				'pixel_id_missing'                  => __( 'Pixel ID is missing', 'facebook-for-commerce' ),
+				'page_id_missing'                   => __( 'Page ID is missing', 'facebook-for-commerce' ),
+				'permissions_response_data_invalid' => __( 'Permissions response data invalid', 'facebook-for-commerce' ),
+				'user_data_invalid'                 => __( 'System User response data invalid', 'facebook-for-commerce' ),
+				'user_access_token_missing'         => __( 'System User access token is missing', 'facebook-for-commerce' ),
+				'fbe_user_data_invalid'             => __( 'Get FBE System User response data invalid', 'facebook-for-commerce' ),
+				'user_access_token_missing'         => __( 'System User access token is missing', 'facebook-for-commerce' ),
+				'user_id_missing'                   => __( 'System User ID missing', 'facebook-for-commerce' ),
+				'token_permission_missing_'         => __( 'The following access token permissions are missing: %s', 'facebook-for-commerce' ),
```
haszari, line 133: Do we need to map these error codes to an error message? Since we're logging, I think it's more robust and simpler to just log the actual error code. 

If we map these to messages, we'll need to maintain the map - the error codes and their meaning may change.

includes/Handlers/Connection.php
```diff
@@ -272,20 +290,24 @@ public function handle_connect() {
...
 
-			$merchant_access_token = ! empty( $_GET['merchant_access_token'] ) ? sanitize_text_field( $_GET['merchant_access_token'] ) : '';
+			$is_error                 = ! empty( $_GET['err'] ) ? true : false;
+			$error_code               = ! empty( $_GET['err_code'] ) ? sanitize_text_field( $_GET['err_code'] ) : '';
+			$merchant_access_token    = ! empty( $_GET['merchant_access_token'] ) ? sanitize_text_field( $_GET['merchant_access_token'] ) : '';
+			$system_user_access_token = ! empty( $_GET['system_user_access_token'] ) ? sanitize_text_field( $_GET['system_user_access_token'] ) : '';
+			$system_user_id           = ! empty( $_GET['system_user_id'] ) ? sanitize_text_field( $_GET['system_user_id'] ) : '';
+
+			if ( $is_error && $proxy_error_message = $this->get_message_for_error_code( $error_code ) ) {
+				throw new SV_WC_API_Exception( $proxy_error_message );
```
haszari, line 300: Ah I see - you want to pass a sentence/english version of the error to `SV_WC_API_Exception` constructor.

I'd suggest just adding a new exception type which handles the logging and all the logic related to these errors (using other exception classes as a base if useful). If we need to map to UI text/sentence form in future, could add this behaviour to exception class. 

So the interface would be something like:

```php
if ( $is_error && $error_code ) {
   throw new FacebookApiException( $error_code );
}

File: includes/Handlers/Connection.php [hunk 3]
```diff
@@ -272,20 +290,24 @@ public function handle_connect() {
...
+			}
 
 			if ( ! $merchant_access_token ) {
 				throw new SV_WC_API_Exception( 'Access token is missing' );
 			}
 
-			$system_user_access_token = ! empty( $_GET['system_user_access_token'] ) ? sanitize_text_field( $_GET['system_user_access_token'] ) : '';
```
haszari, line 281: I see this is an existing pattern mapping a missing field to a human-readable message, e.g. `! system_user_access_token => 'System User access token is missing' - and this is why you did the map of code to message above. I don't think that mapping is adding much value to the log messages - it's a layer of indirection, which is unhelpful IMO.

It would be useful to log the exact field that's missing for these errors, e.g. something like:

`Failed to connect to Facebook. No system_user_access_token provided.`

It looks like this code may have previously shown errors in the UI? I don't understand why the missing fields are converted to log messages like `System User access token is missing` - it's clearer to log the missing field / attribute IMO. 

Possible follow up - simplify and standardise this error handling, with a generic message (`Failed to connect to Facebook. No ${ field_name } provided.`. Or if we decide we still need the "translated" log messages as sentences, use a map/lookup table (similar to what you did for the new ones). I'm not clear on the implications here, so not a blocker, needs more discussion before follow up.

haszari:
> You need an instance of https://connect.woocommerce.com/ to test this PR end-to-end

This is the best way to test! Ideally, follow instructions in this PR: https://github.com/Automattic/connect.woocommerce.com/pull/119

The instructions in this PR description are a good fallback if you're having issues getting connect server set up.

haszari (commented):
I tested this with local connect server running the "show connection errors" PR: https://github.com/Automattic/connect.woocommerce.com/pull/119 

I did a successful connect, disconnect and then a failed connect: I unchecked `ads_read` permission in Facebook. 

The issue was that I was never redirected back to the Woo site. I'm guessing the error code is passed as a redirect param, so the new logging code doesn't have a chance.

I think the blocker here is that [we need a redirect button on the connect server error view](https://github.com/Automattic/connect.woocommerce.com/pull/119#pullrequestreview-636245379).

Aside from that, I think we should keep the error logging code as simple as possible - just log the raw error code. I don't think it's worth mapping the code to a message string for the log. (Unless I misunderstand and these errors are shown in Woo UI somewhere.)

I added some comments about the existing error handling code - similar to above, I think we need to log the basic details, e.g. raw field names and error codes. We can add value by generating nice error messages, but I don't know if that's worth the maintenance. (It might be needed?)

haszari:
Tested this with the corresponding changes in connect server (122-gh-Automattic/connect.woocommerce.com). It's working now 🎉 

- I connected my site to Facebook normally, and enabled debug logging (so I can test that part of it), then disconnected.
- Hacked my Facebook code to use `connect.woocommerce.local` instead of production.
- Went through onboarding and disabled `ads_read` permission.
- Saw the error message on screen:

Screen Shot 2021-04-16 at 2 50 41 PM

- Went back to my Woo site and saw the info in the logs:

Screen Shot 2021-04-16 at 3 06 00 PM

haszari:
FYI @rcstr  I'm still keen for your thoughts on this feedback from my review:

> Aside from that, I think we should keep the error logging code as simple as possible - just log the raw error code. I don't think it's worth mapping the code to a message string for the log. (Unless I misunderstand and these errors are shown in Woo UI somewhere.)

> I added some comments about the existing error handling code - similar to above, I think we need to log the basic details, e.g. raw field names and error codes. We can add value by generating nice error messages, but I don't know if that's worth the maintenance. (It might be needed?)

Not a blocker, just keen to keep this code as simple and direct as possible.

rcstr:
> > Aside from that, I think we should keep the error logging code as simple as possible - just log the raw error code. I don't think it's worth mapping the code to a message string for the log. (Unless I misunderstand and these errors are shown in Woo UI somewhere.)
> 
> > I added some comments about the existing error handling code - similar to above, I think we need to log the basic details, e.g. raw field names and error codes. We can add value by generating nice error messages, but I don't know if that's worth the maintenance. (It might be needed?)


@haszari  -  thanks for the suggestions, I thought about it, as you mentioned it makes sense to keep it simple and avoid adding another thing to maintain that won't provide much value. I have updated the implementation 75ba39c...373700e.

let me know your thoughs

class-wc-facebookcommerce.php
```diff
@@ -383,8 +384,8 @@ public function add_wordpress_integration() {
 		 * @param string $log_id optional log id to segment the files by, defaults to plugin id
 		 */
 		public function log( $message, $log_id = null ) {
-			// bail if logging isn't enabled
-			if ( ! $this->get_integration() || ! $this->get_integration()->is_debug_mode_enabled() ) {
+			// bail if logging isn't enabled, always enabled if is not connected.
```
haszari, line 387: Confusing logic and comment! Does something like this make it clearer?

`Bail if site is connected and user has disabled logging`

Probably worth explicitly documenting that we force logging if site is disconnected so it's easier to support/diagnose. So maybe…

```php
// Bail if site is connected and user has disabled logging.
// If site is disconnected, force-enable logging so merchant can diagnose connection issues.

File: includes/Exceptions/ConnectWCAPIException.php [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+
+namespace SkyVerge\WooCommerce\Facebook\API\Exceptions;
+
+defined( 'ABSPATH' ) or exit;
+
+/**
+ * Class Connect_WC_API_Exception
+ * @package SkyVerge\WooCommerce\Facebook\API\Exceptions
+ */
+class Connect_WC_API_Exception extends \Exception {
```
haszari, line 11: That doc comment is a great place to document the purpose / intention of the exception! Pretty clear from name but maybe could be clearer.

includes/Handlers/Connection.php
```diff
@@ -313,6 +320,11 @@ public function handle_connect() {
 
 			facebook_for_woocommerce()->log( sprintf( 'Connection failed: %s', $exception->getMessage() ) );
 
+			set_transient( 'wc_facebook_connection_failed', time(), 30 );
+		} catch ( Connect_WC_API_Exception $exception ) {
+
+			facebook_for_woocommerce()->log( sprintf( 'Failed to connect to Facebook. No %s provided', $exception->getMessage() ), 'facebook_for_woocommerce_connect' );
```
haszari, line 326: Is that message make sense for all possible error codes? I'd make it general, for example `Failed to connect to Facebook. Facebook API returned error code %s`

haszari (approved):
This looks great - code is much simpler too - nice bonus.

I added some comments about make things clearer, totally optional if you want to follow up.

I haven't retested this – assuming someone is going to give this another test at some point. If you need me to retest I'm happy to - let me know.

One tip - the PR description is now incorrect, which could confuse other reviewers (or future developers). Not a big deal but if you are getting another review I'd tweak it so it's clear on what's implemented (matches current code) and how to test.

includes/Handlers/Connection.php
rcstr: updated in 31d7649

class-wc-facebookcommerce.php
rcstr: thank you for the suggestion, updated in ff9b59f

class-wc-facebookcommerce.php
```diff
@@ -383,8 +384,9 @@ public function add_wordpress_integration() {
 		 * @param string $log_id optional log id to segment the files by, defaults to plugin id
 		 */
 		public function log( $message, $log_id = null ) {
-			// bail if logging isn't enabled
-			if ( ! $this->get_integration() || ! $this->get_integration()->is_debug_mode_enabled() ) {
+			// Bail if site is connected and user has disabled logging.
+			// If site is disconnected, force-enable logging so merchant can diagnose connection issues.
+			if ( ( ! $this->get_integration() || ! $this->get_integration()->is_debug_mode_enabled() ) && $this->get_connection_handler()->is_connected() ) {
```
budzanowski, line 389: Do we have other places that would start logging stuff when there is no connection?

budzanowski (approved):
I have left a small question but otherwise, this looks great!

budzanowski:
One more question. I have noticed that connection errors are logged in a different log file name - was this intentional?
In the description we have:
> Now go to WooCommerce > Status > Log and select the last one with name facebook_for_woocommerce- and confirm error is being logged:

This is no longer true. One needs to look for the `facebook_for_woocommerce_connect-` instead.

class-wc-facebookcommerce.php
rcstr: Did a quick search and didn't anything rang a bell, I wonder why you ask? Will merge this PR and we can expand on this if necessary on another PR 💯

rcstr:
thanks for the reviews team, time to merge 🙌🏽

Issue #1873, opened 2021-04-12
Add setting that allows enabling/disabling feed file generation.


budzanowski:
For some of the sites feed file is unnecessary. Besides it not being used it can slow down the site and cause performance issues. We need to add a setting that allows disabling the feed file generation. We also need to add a documentation section that explains when the feed file should be used.﻿

haszari:
To clarify some more context here:

- There are two mechanisms for sync implemented in this plugin - feed (FB pull) and batch API (Woo push). 
- The feed approach is no longer recommended in favour of the batch API, which allows incremental updates. Batch API is used by default.
- It's possible to configure Facebook to pull from the feed file, and some existing sites may still be using it.
- The feed file is generated on all sites, even if it's not needed.
- The feed file includes all products, so can be expensive to generate and transmit (CPU/bandwidth).

So we're proposing adding an setting to disable, so sites that don't need the feed file can disable it.

haszari:
Update: after discussion with WooCommerce and Facebook developers, we've confirmed that the feed-based sync is still an important component and will need to be maintained. 

Specifically:

- Both sync mechanisms are an important part of syncing the store from Woo => Facebook.
- Feed is used for periodic update of full catalogue and inventory. There are some performance/scaling implications which we're looking at.
- Batch API is used to incrementally update products as they are edited. 

@budzanowski I'm closing this in favour of reducing the feed generation interval (#1882) and ongoing work to optimise for larger stores.

haszari:
Reopening.

Although feed is an important part of product sync, the current implementation may cause issues for larger stores. We're working on improving feed code so it's reliable for all stores.

In the mean time, it would be useful for larger stores to have the option to disable the feed, or specifically disable the existing/legacy feed code.

haszari:
Related issue: #1871

PR #1874, opened 2021-04-12
Introduce option to disable feed file generation.

budzanowski:
Fixes: #1873.

This PR adds an option to disable feed file generation.﻿
A follow-up to this item will be a documentation update that will describe when to disable or enable feed file generation.

In addition to adding the option, we also track it.

I have not updated the changelog since I target this for 2.5.0
Changelog:

Add - Settings option to disable feed file generation.
Add - Track feed file generation option in the tracker.

### DIFF
--- a/includes/Admin/Settings_Screens/Product_Sync.php
+++ b/includes/Admin/Settings_Screens/Product_Sync.php
@@ -318,6 +318,27 @@ class Product_Sync extends Admin\Abstract_Settings_Screen {
 			],
 			[ 'type' => 'sectionend' ],
 
+			[
+				'name' => __( 'Feed File Settings', 'woocommerce' ),
+				'type' => 'title',
+				'id'   => 'feed_file_settings',
+			],
+
+			[
+				'id'      => \WC_Facebookcommerce_Integration::SETTING_FEED_FILE_GENERATION_ENABLED,
+				'title'   => __( 'Enable feed file generation.', 'facebook-for-woocommerce' ),
+				'desc'    => sprintf(
+					/* translators: url to documentation section. */
+					__( 'Feed file is used for cyclic catalog content synchronization. Please check the documentation at %s to learn if this option should be enabled for your store.', 'facebook-for-woocommerce' ),
+					'<a href="' . facebook_for_woocommerce()->get_documentation_url() . '#feed-file-settings">Feed File Settings</a>',
+				),
+				'type'    => 'checkbox',
+				'label'   => ' ',
+				'default' => 'yes',
+			],
+
+			[ 'type' => 'sectionend', 'id' => 'feed_file_settings' ],
+
 		];
 	}
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -68,6 +68,9 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var string the "enable product sync" setting ID */
 	const SETTING_ENABLE_PRODUCT_SYNC = 'wc_facebook_enable_product_sync';
 
+	/** @var string the product description mode setting ID */
+	const SETTING_FEED_FILE_GENERATION_ENABLED = 'wc_facebook_feed_file_generation_enabled';
+
 	/** @var string the excluded product category IDs setting ID */
 	const SETTING_EXCLUDED_PRODUCT_CATEGORY_IDS = 'wc_facebook_excluded_product_category_ids';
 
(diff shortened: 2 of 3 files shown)


includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -318,6 +318,27 @@ public function get_settings() {
 			],
 			[ 'type' => 'sectionend' ],
 
+			[
+				'name' => __( 'Feed File Settings', 'woocommerce' ),
```
haszari, line 322: Should be sentence case. 

Though we might want to review the UI for this - having it as a regular option in the UI may be confusing if we're trying to encourage merchants to turn it off.

includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -318,6 +318,27 @@ public function get_settings() {
...
+				'type' => 'title',
+				'id'   => 'feed_file_settings',
+			],
+
+			[
+				'id'      => \WC_Facebookcommerce_Integration::SETTING_FEED_FILE_GENERATION_ENABLED,
+				'title'   => __( 'Enable feed file generation.', 'facebook-for-woocommerce' ),
+				'desc'    => sprintf(
+					/* translators: url to documentation section. */
+					__( 'Feed file is used for cyclic catalog content synchronization. Please check the documentation at %s to learn if this option should be enabled for your store.', 'facebook-for-woocommerce' ),
```
haszari, line 332: Assuming we stick with a checkbox in UI for this, let's iterate on the wording to make it clear what we recommend. We can also use a tooltip to declutter the user interface.

haszari (commented):
Thanks for prototyping this @budzanowski – this is a good start and maybe part of the process of moving away from feed-based sync. 

However I think we need to step back and decide how we want this process to work - do we want to deprecate the option, or gather usage info, or disable in new stores (etc - lots of possibilities here). If we are planning on deprecating, we'll need to do more work on the UI and notices/guidance (maybe even emails) to work through this with merchants.

I tested this PR and it's working as intended. There's a new checkbox in the `Product sync` UI:

Screen Shot 2021-04-13 at 11 23 34 AM

Disabling this checkbox or the underlying option disables the async task for generating the feed. Here's a screenshot of the task after I re-enabled it:

Screen Shot 2021-04-13 at 11 40 16 AM

### Next steps
I'm marking this as blocked until we've had discussion about how this fits in and what UI works best for our needs.

I can think of a few ideas for how we might handle this transition:

- Add the option, with no UI. This is safe, but doesn't change anything, and allows merchants a way to disable feed on larger stores. We could add a notice or UI for larger stores (many products).
- Add an option with flipped logic: make the feed disabled by default, and only enable it if stores opt-in. We might need to work through this over a few releases, with docs/guidance, so we don't break stores that rely on the feed.
- Reduce the frequency of feed generation. It's every 15 minutes at present, if we do it less often, sites won't be "broken" but impact would be reduced. (This doesn't seem very useful.)
- Disable by default, and track [requests for the feed (`handle_feed_data_request()`)](https://github.com/woocommerce/facebook-for-woocommerce/blob/master/includes/Products/Feed.php#L74) in the store in a transient or store setting somewhere. Then if the feed is requested we can either enable, or guide merchant to disable setting on Facebook end.

Those are just some raw ideas - we need to discuss this and work through the product implications to decide what to ship.

haszari:
Side note: @budzanowski is there a reason why this branch is on a fork of the repo? 😄

haszari:
@budzanowski  are you ok with closing this PR? As far as I understand, we're not planning on allowing disable of feed, as it is a necessary part of sync system. Closing - let me know or comment if there's changes on here you want get ship shape 😁

haszari:
Reopening.

Although feed is an important part of product sync, the current implementation may cause issues for larger stores. We're working on improving feed code so it's reliable for all stores. 

In the mean time, it would be useful for larger stores to have the option to disable the feed.

Reopening this PR to add that option (may shift to a new PR later). Note this option should not have UI - this option is a support tool for large stores, not a regular option.

danielbitzer:
Are you thinking that this option would be supported long-term or is it just a stop-gap solution until we complete the feed rebuild?

haszari:
> Are you thinking that this option would be supported long-term or is it just a stop-gap solution until we complete the feed rebuild?

@danielbitzer definitely temporary. Need to consider what happens when we ship the improved feed generation engine - do we want the option to disable that?

haszari:
Closing, see issue #1873 and new PR #1932

PR #1885, opened 2021-04-16
use existing / current tab for connection flow `Get Started` button

### LINKED ISSUE #1863: `Get Started` connection button opens in new tab, leaving orphaned extra tab
## ✍️ Describe the bug
The `Get Started` button opens a new tab by default (`target="_blank"`). When the connection flow completes, the redirect back to Woo dashboard uses the same tab, leaving a redundant WP dashboard tab open. 

Screen Shot 2021-04-08 at 12 38 47 PM

Not a major but might be simpler to not force new tab (user can always cmd/ctrl-click or right click to open separate tab). cc @j111q - do we have guidelines in Woo/WP for when to use new tab?

If we do decide to keep the new tab behaviour, consider adding an "external link" icon to the button to indicate it will launch a new tab (and on a separate domain).

## 🚶‍♀️ Steps to reproduce
1. Start with a disconnected Facebook for WooCommerce plugin.
2. Go to `WooCommerce > Marketing > Facebook`.
3. Click `Get Started` and connect site.

At the end of the connection flow, you have an extra (stale) tab in "disconnected" state.

## ✔️ Expected behavior
Use same tab for simplicity and to avoid stale UI state.

haszari:
Fixes #1863 

The connection flow is designed to guide merchant through connection settings on Facebook and back to their Woo dashboard.

The `Get Started` currently opens a new tab. At the end of onboarding, the previous tab has stale/old content and could cause confusion.

This PR ensures the button opens in current tab by removing `target="_blank"` attribute.

### How to test
- Disconnect your Woo site from Facebook, or start with a disconnected site.
- Go to `wp-admin > Marketing > Facebook` and click `Get Started` and follow the prompts to connect.

Should not launch a new tab, and at the completion of onboarding, there should be no spare or redundant tabs.

Please also test various user scenarios:

- Using `command` or `control` key to force a new tab anyway (browser/OS dependent) - we should not break this for users that manually override.
- Different browsers or mobile devices.
- Browser settings for new tab/same tab (I think some browsers have a preference/option).
-

### DIFF
--- a/includes/Admin/Settings_Screens/Connection.php
+++ b/includes/Admin/Settings_Screens/Connection.php
@@ -281,7 +281,7 @@ class Connection extends Admin\Abstract_Settings_Screen {
 
 				<?php else : ?>
 
-					<a href="<?php echo esc_url( facebook_for_woocommerce()->get_connection_handler()->get_connect_url() ); ?>" class="button button-primary" target="_blank">
+					<a href="<?php echo esc_url( facebook_for_woocommerce()->get_connection_handler()->get_connect_url() ); ?>" class="button button-primary">
 						<?php esc_html_e( 'Get Started', 'facebook-for-woocommerce' ); ?>
 					</a>
 


rcstr (approved):
change looks good and works as expected.

PR #1887, opened 2021-04-20
Implement/feed generator

budzanowski:
This is a draft PR.

Implementation of new style Feed generation.
Most of the things are subject to discussion and may change.


Sifat1998 (approved):

danielbitzer:
Thanks for all the initial work done here @budzanowski!

I know this is unfinished but I just wanted to suggest something as I've been working on the AS side of things. I think we should split up the `FB_Feed_Generator` class here becase it's responsible for too many things. I think we can separate out the code responsible for:
- the Action Scheduler job batches, [perhaps into a job](https://github.com/woocommerce/facebook-for-woocommerce/pull/1913)
- the ajax endpoints, could become REST endpoints
- handling nad writing to the the .csv file

includes/Products/class-fb-feed-generator.php
```diff
@@ -0,0 +1,579 @@
...
+			$this->get_feed_directory() . $this->get_temp_filename(),
+			$this->get_feed_directory() . $this->get_filename()
+		); // phpcs:ignore WordPress.VIP.FileSystemWritesDisallow.file_ops_rename, Generic.PHP.NoSilencedErrors.Discouraged,
+	}
+
+	public function prepare_feed_generation() {
+		$this->prepare_feed_folder();
+		$products_ids = \WC_Facebookcommerce_Utils::get_all_product_ids_for_sync();
+		//$products_ids = array_slice( $products_ids, 0, 150, true );
+		update_option(
```
danielbitzer, line 352: In my experience with AW I've found reading and writing a large options value for every batch is slow and not as scalable as using querying items in chunks using an offset which is what I'm suggesting here: https://github.com/woocommerce/facebook-for-woocommerce/pull/1913/files#diff-441c6538f1e838f65093411744cb6e11279e449a05b9968ee30dea71ee82e288R50-R51

budzanowski: Well, I think that this may be actually the opposite way 🤣 It all depends on the size of the items in the batch. Please see https://github.com/woocommerce/facebook-for-woocommerce/pull/1913/files#r622851965 comment.

haszari:
I'm going to test the generated feed file and log an issue for any problems revealed by Facebook validator - e.g. missing fields etc.

haszari:
Note – I see various UI code on this PR (maybe temporary - this is a draft). Can we move that off this branch so we can focus on shipping the feed engine? We'll need to look at the UI in the future sprint, so I'd like to avoid pre-empting that (design discussion) by implementing any UI now. 

If we absolutely need UI to ship this, let's keep it super basic, and hidden by default.

haszari:
Looking at the UI, we can probably just log the status info (IDs etc), and use CLI, Action Scheduler or wp shell to trigger generating a feed for testing purposes.

Screen Shot 2021-05-06 at 4 17 50 PM

haszari:
I've given this an end-to-end test and 😿 it's working great! There's a bit more work and coordination needed to ship the improved feed system - this gets us closer 🙌 

### How I tested
- Checked out this branch and ensured the code was built (using .. `npx run sake` - may not be needed).

- **Configure a Facebook catalogue data source**
- Go to `wp-admin > WooCommerce > Marketing > Facebook > Feed Status` (tab).
- You should see some info about a data source (feed) configured in Facebook, specifically a `Feed id`:

Screen Shot 2021-05-06 at 4 17 50 PM

- Go to `business.facebook.com` `Commerce > Catalogue > Data sources`.
- You should see a data source with the same ID.

Screen Shot 2021-05-06 at 4 29 42 PM

- **Manually generate a feed file**
- Go back to `wp-admin` and generate a feed file - click `Generate Feed` button.
- You'll see a progress UI while the feed generates.
- When it completes, it should show a path to the new feed file:

Screen Shot 2021-05-06 at 4 57 54 PM

- Get that file and look inside - it's a good ol' csv with your product info.

- **Manually upload the feed file to Facebook** (to verify it works correctly)
- Go to `business.facebook.com` `Commerce > Catalogue > Data sources`.
- Click on the link (right at end of row) to open details UI for the correct feed.
- Click `Settings` tab.

Screen Shot 2021-05-06 at 4 35 49 PM

- Scroll down to `Single upload` and upload the file.

Screen Shot 2021-05-06 at 4 35 56 PM

- You'll see progress…

Screen Shot 2021-05-06 at 4 36 15 PM

- ...and when it completes you'll see a summary.

Screen Shot 2021-05-06 at 5 02 21 PM

- **Check issues/warnings with the feed file**
- Click `View report` to see all the issues or download a report.
- My issues report is attached below - looks like minor warnings, but there are some things we can improve in the feed file (e.g. product image urls). We can log these as follow up issues.

[errors_for_feed_upload_06_05_2021, 16_36.csv](https://github.com/woocommerce/facebook-for-woocommerce/files/6431862/errors_for_feed_upload_06_05_2021.16_36.csv)

### Testing results
- This feed file works!
- There are some file details we need to tweak/improve (I suggest we handle that as follow up).

### Feedback
There are a few components that this prototype PR implements. I think we should break the overall feature down into smaller components that we can merge and test independently, and also ship in releases (behind a feature flag) so we can test in a variety of stores. 

These are some of the "component" goals as I see it:

1. Remove/kill bad old feed code so it doesn’t cause problems
1. Generate a new feed (CLI command/hidden feature) (the core of this PR)
1. Generate feed file on a schedule with Action Scheduler
1. Serve feed file (expose it so FB can http-get it)
1. Register/configure a data source in Facebook Commerce (via API) - as part of onboarding, or on demand, maybe UI in product sync settings
1. Options & UI for configuring schedule and other settings to optimise for different number of products
1. Options for tweaking the details of what’s in the feed file (if needed)

I've numbered these as I think they depend on each other - but we can merge each independently IMO (and may be able to break it up further).

I don't think the ones near the bottom are as urgent as the ones at the top of the list - we may not need to ship a UI for this in 2.4.0 or 2.5.0. I definitely think we need to decouple the tech/engine from the UI now - that's essential to get this PR merged IMO.

### Next steps

As mentioned above, I think the most important blocker is that we need to agree on how to break this down into smaller pieces that we can test and ship individually, so we don't have to ship a big new feature in one (scary) release.

Here are some ideas for next steps:

- Split off a separate PR which removes code associated with the old feed generation. This is essentially a bug fix as it can impact larger stores. Merge and ship that ASAP.
- Prepare a PR that implements feed generation with no UI - possibly adapt this PR.

danielbitzer:
I agree about the UI. We should focus on shipping the engine and removing any UI related code would de-clutter the PR.

> we can probably just log the status info (IDs etc), and use CLI, Action Scheduler or wp shell to trigger generating a feed for testing purposes.

Agreed

haszari:
> Split off a separate PR which removes code associated with the old feed generation. This is essentially a bug fix as it can impact larger stores. Merge and ship that ASAP.

Follow up on this.

As discussed in team chat:

- There are stores using the existing feed generation code without problems. The existing code works, it just has issues with larger stores (many products) X host environments with lower resources (uses a lot of memory).
- So we shouldn't remove the existing feed generation code until we have a replacement - we don't want to remove a capability that sites are using.

Alternatively, we'll add a new option to allow sites to selectively disable the old feed generation if it's causing issues. @haszari  working on this, picking up from #1874

haszari:
@budzanowski I've retitled this PR to clarify that it's a prototype - apologies if I misunderstand.

Since this is accumulating conflicts, I'm assuming we're keeping the PR around as a reference but not planning on merging (`blocked`).

#1924 is not conflicted and contains similar code, along with other work from @danielbitzer , so I'm assuming that PR is the focus.

PR #1908, opened 2021-04-28
Add “Requires PHP” plugin header

### LINKED ISSUE #1907: Set a minimum required PHP version
This was originally discussed in https://github.com/woocommerce/facebook-for-woocommerce/issues/1906

## Reasoning:

We should define a minimum required PHP version so we all know what language features we can utilize. We also want to make it easier for users by not triggering fatal errors on sites that use old PHP versions. Instead, we want to fail gracefully when encountering an incompatible version of PHP.

I believe we should use PHP 7.0 to align with WooCommerce core and WC Admin.

## Implementation:

Initially I thought the best implementation would be to "define a minimum Woo version and inherit the min PHP version from that Woo version".

However, after some more investigation, I think we should use the `Requires PHP` WordPress plugin header since this is a WordPress.org hosted plugin. This is also what WooCommerce core and WC Admin uses. A handy feature of this header is that incompatible sites will be prevented from updating the plugin and shown a warning (I couldn't find any docs for this). 

More on the `Requires PHP` header:
- https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/
- https://core.trac.wordpress.org/ticket/40934
- https://developer.wordpress.org/plugins/plugin-basics/header-requirements/

danielbitzer:

Sets a minimum PHP version for the plugin as proposed in #1907

Closes #1907 

You can't test the changes until the plugin is deployed to WordPress.org.

### Changelog entry

> New - Set minimum PHP version to 7.0


budzanowski (approved):
:shipit:

rcstr (approved):
🥇

haszari (approved):
Looks good!

> You can't test the changes until the plugin is deployed to WordPress.org.

Trigger warning for me ("can't test")! .. can we try building a zip with the header, and uploading to a test environment running older PHP? I'm guessing you tried that and nothing happens.

Also, can you add info about how we can test/confirm once this change is released and deployed?

danielbitzer:
Once the plugin is deployed to WP.org below are notices that users on incompatible PHP versions will see when they attempt to update or install.

It's quite a neat solution WP.org has implemented.

Screen Shot 2021-04-29 at 10 32 53 am


Screen Shot 2021-04-29 at 10 46 46 am

PR #1910, opened 2021-04-28
Generate PHP7.0 compatible code.

### LINKED ISSUE #1906: Fatal plugin error with `GoogleProductTaxonomy.php` on PHP < 7.1
I received the following email from Wordpress

"ince WordPress 5.2 there is a built-in feature that detects when a plugin or theme causes a fatal error on your site, and notifies you with this automated email.

In this case, WordPress caught an error with one of your plugins, Facebook for WooCommerce. Error Details
=============
An error of type E_PARSE was caused in line 6 of the file /home/sites/14b/4/43b0323a81/public_html/wp-content/plugins/facebook-for-woocommerce/includes/Products/GoogleProductTaxonomy.php. Error message: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE)"

Which led to the product descriptions being unable to view in the edit pane but they were still present in white font and had changed tier font style. 

Once I deactivated the plugin this resolved. 

But can I reactivate the plugin or what is the best step. We use this plugin on a number of sites so are worried about it 

## 🔎 Isolate the bug

- [ ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.

## ✍️ Describe the bug

##🚶‍♀️ Steps to reproduce

1. The first step
1. The next step…

budzanowski:

Explicit scope quantifier on const is not supported in PHP<7.1. Since we required public and public is implicit removing the `const` keyword is all that is needed to fix #1906.

Closes #1906.

1. Edit a product.
2. Open the Facebook tab.
3. Categories setting should work
image

### Changelog entry

 * Fix - PHP<7.1 incompatible code for Google Taxonomy Setting in products.


danielbitzer:
Is the plan to ship this fix in 2.4.1? If so, I'd suggest creating a new release branch starting from the 2.4.0 tag rather than `master`. That way we don't ship anything intended for 2.5.0.

Something strange happened to the .pot file and and some JS files in this PR.

budzanowski:
I don't know what the plan is, we will discuss this on today's dev call.
yep, the .pot and js are my faults. I generated a zip for testing on Jurassic ninja and accidentally pushed all the change files.=.

budzanowski:
Unnecessary changes removed.

rcstr:
Tested this PR with `PHPCompatibility` and didn't get any error on our source files but I got some warnings from the `vendor/` folder, setting also the PHP version for composer could help us solve these and get only dependencies will be PHP 7.0 compatibility.

[This is the full report](https://gist.githubusercontent.com/rcstr/a9521ab3d6c3762ee01c62c139289b8a/raw/210b2a009358d48bb0e45773df9c187e4589cce4/report.txt)

budzanowski:
I think that we require PHP 7.1+ for dev-related libraries.

haszari:
> Tested this PR with PHPCompatibility and didn't get any error on our source files but I got some warnings from the vendor/ folder, setting also the PHP version for composer could help us solve these and get only dependencies will be PHP 7.0 compatibility.

Side question: Should we exclude `/vendor` from PHPCompatibility? Seems like we should just run it on our own code.

bin/GenerateCategories.php
```diff
@@ -26,7 +26,7 @@ public function prepare_categories( $file ) {
 		'// This file was generated using GenerateCategories.php, do not modify it manually' . PHP_EOL .
 		'// php GenerateCategories.php taxonomy-with-ids.en-US.txt' . PHP_EOL .
 		'class GoogleProductTaxonomy {' . PHP_EOL .
-		'	public const TAXONOMY = %s' . PHP_EOL .
+		'	const TAXONOMY = %s' . PHP_EOL .
```
haszari, line 29: > Since we required public and public is implicit removing the const keyword is all that is needed

Just to clarify - the fix is to remove `public`, not remove `const` as mentioned in description!

haszari (approved):
Looks good to me (note I haven't tested). 

I think it's probably easiest to deploy a 2.4.1 patch release with this change. We could do that by retargeting this PR to a new `release/2.4.1` branch or just merge to master and cherry pick to release branch.

rcstr:
> Side question: @rcstr should we exclude `/vendor` from PHPCompatibility? Seems like we should just run it on our own code.

Based on our [`composer.json`](
https://github.com/woocommerce/facebook-for-woocommerce/blob/master/phpcs.xml#L26-L27) we are already doing that but stills checks that folder and I couldn't find a way to (really) ignore it.

haszari:
>  > Side question: @rcstr should we exclude /vendor from PHPCompatibility? Seems like we should just run it on our own code.

> Based on our composer.json we are already doing that but stills checks that folder and I couldn't find a way to (really) ignore it.

@rcstr .. you linked to phpcs.xml, which configures phpcs, I don't think PHPCompatibility uses that config. 🤷

rcstr:
> @rcstr .. you linked to phpcs.xml, which configures phpcs, I don't think PHPCompatibility uses that config. 🤷

It does actually, PHPCompatibility it's just a code standard for PHPCS.

haszari:
Added to 2.5.0 for visibility for now - obviously feel free to shift if we do a 2.4.1.

bin/GenerateCategories.php
budzanowski: Yes, sorry.

budzanowski:
Closing in favour of #1914

PR #1911, opened 2021-04-28
Add php coding standards checks.

budzanowski:
This adds phpcs testing according to the woocommerce-core-sniffs.
It also checks for a minimum PHP 7.0 version


budzanowski:
Merging since I can't this can't be tested from branch.

PR #1913, opened 2021-04-29
Introduce generate feed job structure

danielbitzer:

I'm proposing we introduce a simple Action Scheduler "job" structure for the **generate product feed** background job we need to fix/improve.

Since a lot of our plugins use Action Scheduler to run jobs I think we should consider building this in a framework model that we can use across multiple plugins like FB, AutomateWoo, GLA and most likely other Woo extensions. I've initially set this up as over at https://github.com/woocommerce/action-scheduler-job-framework and installed it here via Composer. We could also simply rename the namespace and go with the old copy/paste approach if that's preferred.

My thinking is that the `GenerateProductFeed` class will contain a subset of the logic in @budzanowski's prototype `FB_Feed_Generator` class in #1887.

I've included some info in the `README` about how the AS job framework works: https://github.com/woocommerce/action-scheduler-job-framework#chained-jobs

TBA

### Changelog entry

None


danielbitzer:
I've requested a review to get some feedback.

includes/Jobs/GenerateProductFeed.php
```diff
@@ -0,0 +1,112 @@
...
+	 *
+	 * NOTE: when using an OFFSET based query to retrieve items it's recommended to order by the item ID while
+	 * ASCENDING. This is so that any newly added items will not disrupt the query offset.
+	 *
+	 * @param int   $batch_number The batch number increments for each new batch in the job cycle.
+	 * @param array $args         The args for the job.
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
+	 */
+	protected function get_items_for_batch( int $batch_number, array $args ): array {
```
budzanowski, line 45: This approach has the potential to slow things down significantly and increase memory footprint. There is logic inside the `get_all_product_ids_for_sync` that prepares the item for batching later. I understand your concern about the size of the option which stores the items for processing but without it, you will be forced to do additional queries to figure out if the product variation that you are processing has a parent in the published state.

includes/Jobs/GenerateProductFeed.php
```diff
@@ -0,0 +1,112 @@
...
+			'fields'         => 'ids',
+			'post_status'    => 'publish',
+			'post_type'      => [ 'product', 'product_variation' ],
+			'posts_per_page' => $this->get_batch_size(),
+			'offset'         => $this->get_query_offset( $batch_number ),
+			'orderby'        => 'ID',
+			'order'          => 'ASC',
+		];
+
+		$query = new WP_Query( $product_args );
```
budzanowski, line 56: Fetching by id felt much faster - but this is not something I have measured.

includes/Jobs/GenerateProductFeed.php
```diff
@@ -0,0 +1,112 @@
...
+	 * Process a single item.
+	 *
+	 * @param string|int|array $item A single item from the get_items_for_batch() method.
+	 * @param array            $args The args for the job.
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
+	 */
+	protected function process_item( $item, array $args ) {
+		try {
+			$product = wc_get_product( $item );
```
budzanowski, line 70: Individual queries for single products are slow and bloat memory. It looks like the approach using `wc_get_products` could be better.

budzanowski:
Looks cool. I don't really see this being much different at the core of operations compared to what is in #1887 except this being based on a framework.

The downside that I see is that we would need to abandon the WC_Product_CSV_Exporter entirely and copy what is needed into a new structure. We would also need to figure out what is the most performant path for processing the items and how to build the batch and query items correctly.

includes/Jobs/GenerateProductFeed.php
budzanowski: I would argue that the reading option with prepared items not only solves the published variation parents issue but it would be faster than doing a `new WP_Query` every time.
What we could do to optimize a bit is to have the batch number in a separate option so we would only need to read the option once and we would store the batch number after batch processing.

danielbitzer:
> I don't really see this being much different at the core of operations compared to what is in #1887 except this being based on a framework.

Yes the main differences are the code structure and the approach to querying both of which can be refined. My main thinking is to introduce some structure to the code and have separate class that are have clear responsibilities rather than one class that does everything related to the feed.

> The downside that I see is that we would need to abandon the `WC_Product_CSV_Exporter` entirely and copy what is needed into the new structure

I don't think this is true and yes I don't think we should abandon that class. A "job" class would only be responsible for the Action Scheduler side of things and would call functions in a separate class that handles the CSV exporting.

includes/Jobs/GenerateProductFeed.php
danielbitzer: This is something we'll have to test I think. I just know when we used this approach of doing large option table read/writes for each batch when using AW we had user complaints.

> What we could do to optimize a bit is to have the batch number in a separate option so we would only need to read the option once and we would store the batch number after batch processing.

Do you mean we would only need to "write" the option once?

includes/Jobs/GenerateProductFeed.php
danielbitzer: That's true, I will adjust

includes/Jobs/GenerateProductFeed.php
danielbitzer: Not according to my tests, I've got 20,000 products on my local site and this is what I observe:

Calling `get_all_product_ids_for_sync` and writing all the product IDs to an option:

Screen Shot 2021-04-30 at 9 29 30 am

Here I've used `wc_get_products()` (as you suggested in another comment) to select full product objects using an offset.

Screen Shot 2021-04-30 at 9 28 03 am

The actual code used:
```php

wc_get_products(
	[
		'status'  => 'publish',
		'type'    => [ 'simple', 'variation' ],
		'limit'   => 10,
		'offset'  => 0,
		'orderby' => 'ID',
		'order'   => 'ASC',
	]
);
```

budzanowski: That is true but you do the query in `get_all_product_ids_for_sync`  only once and that only once and later you just read the option.
But the problem here is different. You don't know if the products that you have collected are valid! Variation in the `publish` state should not be synced unless its parent is also in publish state. So if you happen to have in your query 10 variation. You will need to figure out their parent id and query for their parent's state. With bigger batch numbers this could big a long process. This is exactly why `get_all_product_ids_for_sync` uses two separate queries and combines the results - this was the trick that allowed us to reduce ids collection time significantly.

includes/Jobs/GenerateProductFeed.php
budzanowski: Reading big options is costly and writing it is costly. If we would go with ids in an option approach we could use a separate option for tracking the current batch number. This way, we only need to read the big option and we don't need to write it just to update the batch number.

includes/Jobs/GenerateProductFeed.php
danielbitzer: Gotcha, that's annoying that you can't query variations by status without referencing the parent product. 

I guess my concern is more about scalability rather than optimization. e.g. if a store has 100,000 products the option is going to be ~2 MB by my calculations. That is kind of an extreme amount of products but the more products a store has the high chance there is of the feed crashing with the options approach. If we use offset queries then the feed just takes longer to generate.

includes/Jobs/GenerateProductFeed.php
danielbitzer: > If we would go with ids in an option approach we could use a separate option for tracking the current batch number.

Yeah avoiding writes, might be faster.

> This way, we only need to read the big option and we don't need to write it just to update the batch number.

Actually we don't need to store the batch number as an option since it's passed as an argument to the scheduled action by the job framework.

budzanowski: Even better :)

includes/Jobs/GenerateProductFeed.php
budzanowski: What I am happy about is that the things that we discuss are interchangeable pieces. We could probably skip all of the discussions now and just push forward. When we will have most of this build we will just test. We will know what is the right approach without guessing.

danielbitzer: Yes that's true, we can adjust this later and work on other things.

But if we need to determine the variation status from the parent I think a JOIN is all we need.

```sql
SELECT 
	post.ID, 
	parent.ID as parentID,
	parent.post_status as status
FROM wp_posts as post
LEFT JOIN wp_posts as parent ON post.post_parent = parent.ID 
WHERE post.post_type = 'product_variation'
AND parent.post_status = 'publish'
ORDER BY post.ID ASC
LIMIT 10
``` 

I don't think there's any need to read and write big options values if we use a join like this.

budzanowski: The easiest way to test it is to remove the limit and compare the total ids.

danielbitzer: We can actually select both the variations and the simple products in the same query...
```sql

SELECT post.ID
FROM wp_posts as post
LEFT JOIN wp_posts as parent ON post.post_parent = parent.ID 
WHERE 
	( post.post_type = 'product_variation' AND parent.post_status = 'publish' )
OR 
	( post.post_type = 'product' AND post.post_status = 'publish' )
ORDER BY post.ID ASC
LIMIT 50 OFFSET 0
```

Seems to perform quite well on my site

Screen Shot 2021-04-30 at 3 07 40 pm

composer.json
```diff
@@ -11,16 +11,26 @@
         {
             "type": "git",
             "url" : "git@github.com:facebookincubator/facebook-for-woocommerce.git"
+        },
+        {
+            "type": "vcs",
+            "url": "https://github.com/woocommerce/action-scheduler-job-framework"
```
haszari, line 17: Is this versioned - how do we know what version of the code we're getting? Does this get whatever is latest at time of `composer install`?

class-wc-facebookcommerce.php
```diff
@@ -128,6 +131,7 @@ public function init() {
 			add_filter( 'woocommerce_navigation_get_breadcrumbs', [ $this, 'wc_page_breadcrumbs_filter' ], 99 );
 
 			if ( \WC_Facebookcommerce_Utils::isWoocommerceIntegration() ) {
+				require_once __DIR__ . '/vendor/autoload.php';
```
haszari, line 134: I see we have a related PR migrating to autoloader (#1919) - assuming that will get merged before this. Not sure if it matters, just clarifying that this PR doesn't introduce autoloader.

danielbitzer: I'm planning to merge #1919 and then remove the autoloader addition from this PR.

composer.json
danielbitzer: Yeah if we want to go ahead with this separate job framework approach we should definitely tag a version of it before we merge this PR. Right now it's pulling from the `trunk` branch. I wanted to get a consensus about whether this is something we are willing to pursue as a squad. 

The framework is fairly light. It only contains the features we need for the FB extension so it could easily become part of this repo. However, the advantage of a separate framework is any other extension can use it. E.g. GLA, AutomateWoo, Pinterest...

haszari: Sounds good - probably worth a discussion (P2, feedback from Action Scheduler maintainers). If other projects/teams are interested, probably good to start with the separate repo. I'm happy with either option👌 

I guess there's a third option - add it to Action Scheduler directly. (Maybe that's longer-term goal if it proves useful?)

danielbitzer: A P2 sounds good, will post...

> I guess there's a third option - add it to Action Scheduler directly

Potentially yes but it's very difficult to get anything added to AS. They're seem to be only accepting bug fixes for the last several months.

budzanowski (approved):

haszari:
Great to see this progressing (now merged) 🎉 

@danielbitzer can you update the description to clarify if this changes any behaviour or background processing in the plugin? I.e. will the new `GenerateProductFeed` job run or show up as tasks in Action Scheduler status, or is this code "dormant" until new feed engine code makes use of it?

danielbitzer:
Done, updated PR description and title.

PR #1919, opened 2021-04-30
Use PSR-4 composer autoloader rather than requiring all classes upfront

danielbitzer:
### Changes proposed in this Pull Request:

Autoload classes that are already following the PSR-4 naming convention with composer rather than requiring every class upfront.

I removed `require_once` statements for 20 classes, listed below.

Why? It's more efficient, it's quicker for devs and it's less error-prone since the loading happens automatically.

### How to test the changes in this Pull Request:

1. Run `composer install`
2. Check plugin works
3. Run the following code to confirm the 20 classes still load ok:
```php
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Locale::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\AJAX::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Handlers\Connection::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Integrations\Integrations::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Product_Categories::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Commerce::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products\Feed::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products\FBCategories::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products\Stock::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products\Sync::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Products\Sync\Background::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\ProductSets\Sync::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Debug\ProfilingLogger::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Debug\ProfilingLoggerProcess::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Utilities\Shipment::class ) );
var_dump( class_exists( \SkyVerge\WooCommerce\Facebook\Utilities\Tracker::class ) );
[...]

PR #1921, opened 2021-04-30
Fix coding standards with phpcbf and WooCommerce-core sniffs.

budzanowski:
Changes have been automatically generated using phpcbf and coding standards for WooCommerce.


budzanowski:
The test is failing because it has found additional issues not fixable with phpcbf.

Ismail-elkorchi:
The failing sniffs can be disabled individually in phpcs.xml to let the test pass. Then, in other PRs those disabled sniffs can be enabled as the errors related to it get fixed.

budzanowski:
The plan is to live with the failing test for now but at the same time fix as many issues as possible until we don't have any.

budzanowski:
Since this is auto-generated based on WooCommerce Core standard there is no reason, and sense ( given the size of the issue ) to review this. Merging.

haszari:
Added to 2.5.0 - another reason to so some solid testing of the release (alongside autoloading #1919 and improving assets build #1826 #1824).

Issue #1909, opened 2021-04-28
Attribute values with commas break product sync


slaxxarn:
## 🔎 Isolate the bug

- [ x] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x ] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug

We have comma ‘,’ in our attribute values and by looking at the source code for the plugin (facebook-for-woocommerce/includes/Products/Sync/Background.php line 299 to 304) you guys do an implode on attributes by comma sign to convert all attributes to a string for “additional_variant_attribute”. This breaks our import by, i guess, on facebooks end, explode by comma sign aswell.

An example how are attribute values are written: processor:1,8 GHz Fyrk\u00e4rnig Core i7 so the whole additional_variant_attribute becomes

"additional_variant_attribute":"bildskarm:HD 1920x1080,processor:1,8 GHz Fyrk\u00e4rnig Core i7,minne:16GB RAM,4g-cellular:Med 4G Cellular,lagring:512GB SSD"

##🚶‍♀️ Steps to reproduce

1. Create a variable product with attribute values that has comma in them ie "1,8 Ghz Four core i7"
2. Try to sync that product

## ✔️ Expected behavior
Sync it to facebook

## 🗃 Logs

`{"message":"Den additional_variant_attribute informationen under  \u00e4r ogiltig. Granska f\u00f6r mer information"}`

budzanowski:
@slaxxarn thank you for the bug report and the investigation. We will look into this.

slaxxarn:
> @slaxxarn thank you for the bug report and the investigation. We will look into this.

Great, thanks!

A new filter to handle this or just escaping comma signs would be enough

haszari:
Hi there @slaxxarn  - can you add some more details of exactly how to reproduce the issue?

I tried adding an attribute value with a comma to a product (`Color: "Blue, ocean, aquamarine, sky, duck-egg"`), and it synched correctly as far as I can tell.

<img width="1086" alt="Screen Shot 2021-04-29 at 12 03 28 PM" src="https://user-images.githubusercontent.com/4167300/116486949-ea2fab80-a8e2-11eb-92ec-220762d11a7b.png">
[...]

PR #1922, opened 2021-05-03
Escape commas inside additional attributes value.

budzanowski:

Commas inside attribute values are not supported and break the synchronization. The proposed solution is to replace the commas with a space char and add a filter that allows customization of behavior.

Fixes: #1909

1. Create attributes for variations that have commas ( more info inside the #1909 issue ):
image
2. Create Variations out of the attributes.
3. use Delete product(s) on Facebook from the Facebook box in the product edit page.
4. Click Update the product.
5. Products should be synced correctly ( commas replaced )

Testing the filter:

```PHP
function facebook_for_woocommerce_variant_attribute_comma_replacement_callback( $replacement ) {
	return '.';
}

add_filter( 'facebook_for_woocommerce_variant_attribute_comma_replacement', 'facebook_for_woocommerce_variant_attribute_comma_replacement_callback' );
```

### Changelog entry

 * Fix - Attribute values with commas break product sync.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Facebook for WooCommerce Changelog ***
 
 2021.nn.nn - version 2.5.0
- * Fix - Attribute values with commas break product sync.
+ *  Ensure variable product attribute values containing , sync correctly to Facebook additional_variant_attribute field.
 
 2021.04.29 - version 2.4.1
  * Fix - PHP<7.1 incompatible code for Google Taxonomy Setting in products.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Facebook for WooCommerce Changelog ***
 
+2021.nn.nn - version 2.5.0
+ * Fix - Attribute values with commas break product sync.
+
 2021.04.29 - version 2.4.1
  * Fix - PHP<7.1 incompatible code for Google Taxonomy Setting in products.
 
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -298,7 +298,7 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 			foreach ( $data['custom_data'] as $key => $val ) {
 				$attribute_value = str_replace(
 					',',
-					apply_filters( 'facebook_for_woocommerce_variant_attribute_comma_replacement', ' ' ),
+					apply_filters( 'facebook_for_woocommerce_variant_attribute_comma_replacement', ' ', $val ),
 					$val
 				);
 				$attributes[]    = $key . ':' . $attribute_value;
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -282,21 +282,26 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 	 *
 	 * @since 2.0.0
 	 *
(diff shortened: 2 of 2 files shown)


includes/Products/Sync/Background.php
```diff
@@ -282,21 +282,26 @@ private function prepare_product_variation_data( $product ) {
...
+		// Attributes other than size, color, pattern, or gender need to be included in the additional_variant_attributes field.
 		if ( isset( $data['custom_data'] ) && is_array( $data['custom_data'] ) ) {
 
 			$attributes = array();
 
 			foreach ( $data['custom_data'] as $key => $val ) {
-				$attributes[] = $key . ':' . $val;
+				$attribute_value = str_replace(
+					',',
+					apply_filters( 'facebook_for_woocommerce_variant_attribute_comma_replacement', ' ' ),
```
haszari, line 301: Suggest adding a docblock for all new hooks or public APIs, explaining the purpose, intended use, parameters & return value, similar to how [WP core documents hooks](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/). This also is a good place to link to the relevant Facebook API/platform docs.

Something like:

```php
/**
 * Filter: facebook_for_woocommerce_variant_attribute_comma_replacement
 *
 * The Facebook API expects a comma-separated list of attributes in `additional_variant_attribute` field.
 * https://developers.facebook.com/docs/marketing-api/catalog/reference/
 * This means that WooCommerce product attributes included in this field should avoid the comma (`,`) character.
 * Facebook for WooCommerce replaces any `,` with a space by default. 
 * This filter allows a site to provide a different replacement string.
 *
 * @since 2.5.0
 *
 * @param string $replacement The default replacement string (`,`).
 * @return string Return the desired replacement string.
 */
```

haszari (approved):
Nice fix - thanks for quickly attending to this little niggle, @budzanowski !

I had problems testing this, because I couldn't get any variable products to sync from my test site (unrelated to this branch - 🤷). So I haven't tested, but the code and reasoning look great to me 🎉 

I added a couple of comments to ensure we document this clearly, not a blocker. 

Suggest getting someone to test before merge, though I'm happy if you've tested thoroughly 🚀

includes/Products/Sync/Background.php
rcstr: I think also worth passing `$val`, so merchants can choose a different replacement based on it.
```suggestion
					apply_filters( 'facebook_for_woocommerce_variant_attribute_comma_replacement', ' ', $val ),
```

rcstr (approved):
The code looks good, I added a suggestion. Sync is running on my store (~5k products) and might take a while until Ill be able to validate this, will keep you posted.

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** Facebook for WooCommerce Changelog ***
 
+2021.nn.nn - version 2.5.0
+ * Fix - Attribute values with commas break product sync.
```
budzanowski, line 4:
```suggestion
Fix – Ensure variable product attribute values containing, sync correctly to Facebook additional_variant_attribute field.
```

rcstr:
Tested and worked as expected

### WC Store 
Screen Shot 2021-05-04 at 22 30 18

### FB Catalog
Screen Shot 2021-05-04 at 22 30 06

PR #1925, opened 2021-05-05
phpcs: Exclude WordPress class file naming rules

danielbitzer:
I think this plugin, like Woo's other flagship plugins (WC Admin, WC Blocks), should ignore WP's class file naming rules and follow PSR-4 instead.

N/A

### Changelog entry

N/A

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -16,7 +16,10 @@
 	<config name="testVersion" value="7.0-" />
 
 	<!-- Rules -->
-	<rule ref="WooCommerce-Core" />
+	<rule ref="WooCommerce-Core">
+		<exclude name="WordPress.Files.FileName.InvalidClassFileName"/>
+		<exclude name="WordPress.Files.FileName.NotHyphenatedLowercase"/>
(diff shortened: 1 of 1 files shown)


phpcs.xml
```diff
@@ -16,7 +16,10 @@
 	<config name="testVersion" value="7.0-" />
 
 	<!-- Rules -->
-	<rule ref="WooCommerce-Core" />
+	<rule ref="WooCommerce-Core">
+		<exclude name="WordPress.Files.FileName.InvalidClassFileName"/>
+		<exclude name="WordPress.Files.FileName.NotHyphenatedLowercase"/>
```
haszari, line 21: Makes sense to me, especially if this is consistent with other cutting-edge Woo plugins (admin, blocks). 

Would be extra nice to add a comment here documenting the decision & rationale.

haszari (approved):

phpcs.xml
danielbitzer: Good call. I've added a comment.

Issue #1871, opened 2021-04-10
Out of Memory issues since 2.3.2


galapogos01:
## 🔎 Isolate the bug

- [X] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [ ] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug

Hi guys,

I have a store with over 3500 products.

Since upgrading to 2.3.2, every time the action scheduler task wc_facebook_regenerate_feed runs it is failing due to out of memory issues:
`unexpected shutdown: PHP Fatal error Allowed memory size of 419430400 bytes exhausted (tried to allocate 20480 bytes) in /wp-includes/functions.php on line 624`

As you can see I have already had to raise the single thread memory limit to 400MB which is very high. I had to do this for prior versions to work at all, but it's been failing consistently since the version upgrade.

Removing the Google Product Categories from the config allows the cron instance of the task to run but the action scheduler still reports errors constantly, and Facebook complains that my products have less visibility due to not having Google categories.

Creating this as https://wordpress.org/support/topic/out-of-memory-issues-since-2-3-2/#post-14300819 led me here.

Thanks,
Jason

##🚶‍♀️ Steps to reproduce

1. Install the plugin on a large (3000+ product) site
2. Watch action scheduler task logs

## ✔️ Expected behavior
Scheduled tasks should not cause OOM errors.

## 🗃 Logs
Can provide specific logs on request

danielbitzer:
Hi @galapogos01, thanks for the report! 

Could you confirm this issue persists when using the latest (2.3.5) version of the plugin?

galapogos01:
Yes @danielbitzer . I think this is different to https://github.com/woocommerce/facebook-for-woocommerce/issues/1818 as 1818 relates to wp-admin and my ticket is the background AS task.

danielbitzer:
Ok I see. We are currently working on improving the feed and product sync areas of the plugin. 

@budzanowski since you've been looking into the feed is there anything we can do here. Sounds like that proposed option to disable the feed would be ideal.

galapogos01:
Confirming these are occurring worse in 2.4.0. Happy to provide any logs requested.

danielbitzer:
[...]

PR #1928, opened 2021-05-06
Set upload schedule on feed creation

### LINKED ISSUE #1903: Create feed file with upload schedule.

rcstr:

PR sets *upload schedule settings* to feed when this is created. These are the available params for the request https://developers.facebook.com/docs/marketing-api/catalog/guides/scheduled-feeds/

Closes #1903.

1. Checkout this branch
2. If we already have a `wc_facebook_feed_id` set, we'll need to delete this. If you have `wp-cli` you can run `wp option delete wc_facebook_feed_id`
3. Now from *wp-admin* go to **Marketing > Facebook > Feed Status** (it might take a few seconds to load)
Screen Shot 2021-05-06 at 02 18 04

4. Now go to **Marketing > Facebook > Feed Status** and go to your *Catalog*
Screen Shot 2021-05-06 at 02 19 52

5. From there, go to **Catalog > Data Sources**, a new feed should have ben added (You can compare the ID from *Feed Status* screen) with schedule information
Screen Shot 2021-05-06 at 02 21 10

### Changelog entry

> New - Configure update schedule so Facebook refreshes product sync feed regularly.

### DIFF
--- a/includes/Products/class-fb-feed-generator.php
+++ b/includes/Products/class-fb-feed-generator.php
@@ -586,5 +586,24 @@ class FB_Feed_Generator extends \WC_Product_CSV_Exporter {
 		return $secret;
 	}
 
+	/**
+	 * Function will return the default parameters for the feed's upload schedule
+	 *
+	 * @link https://developers.facebook.com/docs/marketing-api/reference/product-feed-schedule/
+	 */
+	public function get_feed_upload_schedule() {
+		$schedule = new \stdClass();
 
+		$schedule->interval = 'DAILY';
+		$schedule->url      = esc_url_raw( $this->get_feed_uri() );
+		$schedule->hour     = ( new \WC_DateTime( '+1 hour', new \DateTimeZone( 'UTC' ) ) )->date( 'H' );
+		$schedule->timezone = 'UTC';
+
+		/**
+		 * Filters the value of the default feed schedule settings.
+		 *
+		 * @param  \stdClass  $schedule  Object contains the settings to be passed to fbgraph to configure upload schedule.
+		 */
+		return apply_filters( 'facebook_for_woocommerce_feed_upload_schedule_settings', $schedule );
+	}
 }
--- a/includes/Products/class-fb-feed-generator.php
+++ b/includes/Products/class-fb-feed-generator.php
@@ -141,6 +141,17 @@ class FB_Feed_Generator extends \WC_Product_CSV_Exporter {
 	}
 
 	/**
+	 * Returns the feed uri.
+	 *
+	 * @return string
+	 */
+	public function get_feed_uri() {
+		$upload_dir = wp_upload_dir();
+
+		return sprintf( '%s/%s/%s', $upload_dir['baseurl'], 'facebook_for_woocommerce', $this->get_filename() );
+	}
+
+	/**
 	 * Get file path to export to.
 	 *
 	 * @return string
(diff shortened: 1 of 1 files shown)


includes/Products/class-fb-feed-generator.php
```diff
@@ -189,7 +200,7 @@ public function get_default_column_names() {
 	private function create_feed() {
 		$result = \WC_Facebookcommerce_Utils::$fbgraph->create_feed(
 			facebook_for_woocommerce()->get_integration()->get_product_catalog_id(),
-			array( 'name' => self::FEED_NAME )
+			array( 'name' => self::FEED_NAME, 'schedule' => $this->get_feed_upload_schedule() )
```
budzanowski, line 203: I think that we will end up with `update_schedule` but that is something we need to figure out later.

includes/Products/class-fb-feed-generator.php
```diff
@@ -575,5 +586,24 @@ public static function get_feed_secret() {
 		return $secret;
 	}
 
-
+	/**
+	 * Function will return the default parameters for the feed's upload schedule
+	 *
+	 * @link https://developers.facebook.com/docs/marketing-api/reference/product-feed-schedule/
```
haszari, line 592: Nice linking out to docs for the option 🏆

haszari:
This seems like a solid improvement - thanks @rcstr . However, should we fix this in the existing feed code and ship it in 2.5, or should we wait and implement this once we've got the feed basics working well?

@budzanowski keen for your thoughts here - should we ship this as an enhancement to existing feed in 2.5.0, or hold off?

As far as I understand, this is an enhancement that will work well for most stores (will exacerbate feed-generation issues with large stores, e.g. #1871). The enhancement is that when we tell Facebook about the feed file, we also tell Facebook how often to refresh the catalogue from the feed. Is that correct?

If so, I think we can make this clearer in the changelog. I tried to improve it but it's tricky to make it clear!

> New - Configure update schedule so Facebook refreshes product sync feed regularly.

haszari:
I'm keeping this out of 2.5.0 for now. Since we are generally revamping the feed engine, I don't think we should ship changes to how the existing feed is used/enabled, unless it's to fix a serious or high-impact issue. What do you think @budzanowski ?

includes/Products/class-fb-feed-generator.php
rcstr: Based on [Product Catalog Product Feed](https://developers.facebook.com/docs/marketing-api/reference/product-catalog/product_feeds/) documentation:

- `schedule`:  this will be a full replace operation where items not present in the feed file will be removed, new items will be created and existing items will be updated.
- `update_schedule`: The uploads would only update the items in the feed or create new ones. No items would be deleted. This is useful for sending price and availability updates for selected items in the feed.

`schedule` should works for us unless we plan to create two separate feeds for each case @haszari @budzanowski

haszari: Let's look at this next - when we have a reliable engine for generating the feed file :)

rcstr:
Tried to redo this PR and adapt for `feature/generate-feed-job-merged-implementations` (#1924) but didn't find where to implement 65dacab, looks like the logic is implemented on that branch yet. @budzanowski should I leave this as it is and merge to `implement/feed-generator`?

haszari:
Bumping this to 2.7; 2.6 is focused on safe bug fixes and adding tracking to understand feed usage.

I'm not clear on whether this exact code will be used, but we do want some version of this feature in 2.7. Namely, merchants should be able to configure and use the feed to sync products.

budzanowski:
Closing in favor of: https://github.com/woocommerce/facebook-for-woocommerce/pull/2005

PR #1932, opened 2021-05-07
add option for disabling existing feed

### LINKED ISSUE #1873: Add setting that allows enabling/disabling feed file generation.

haszari:
Fixes #1873

This PR, inspired by #1874, adds a new hidden option to allow sites to disable the current feed generation code.

- Option: `wc_facebook_legacy_feed_file_generation_enabled`
- Value: `yes | no` string
- Default: `yes`

#### Feature flag for new feed engine?
This PR is focused tightly on the current issue for larger stores and giving them a targeted option to disable. At the same time, we may want a feature flag to help us ship the new-improved feed engine (work in progress). 

It might be a good idea to consider the feature flag use case (or even implement in this PR) to ensure this remedial/temporary option doesn't impact our ability to ship improvements to larger stores! See code comment for details.

On the other hand, if we end up using a UI or guided process for enabling the new feed engine, the option can be reset/removed as part of running that. E.g. task list item says `you should enable a feed for improved product sync` > merchant clicks `Set it up`, accepts defaults, feed is configured, disable option is removed. 

Reviewers let me know what you think about this :) cc @budzanowski @danielbitzer 

1. Check out `master`. Confirm that the feed generation job is running:
  - Enable debug logging - `Facebook > Connection > Log events for debugging`.
  - Should see feed profiler logs in `facebook_for_woocommerce_profiling`. 
    - e.g. `2021-05-07T02:27:17+00:00 DEBUG generate_feed - Memory: 23,045.33 KB, time: 2.49s`
  - Should see feed job in `WooCommerce > Status > Scheduled Actions.
    - `wc_facebook_regenerate_feed` 

Screen Shot 2021-05-07 at 2 31 09 PM
Screen Shot 2021-05-07 at 2 45 51 PM

Optional - set up your site with lots of products to reproduce performance issues.

2. Set the option to opt-out of feed. e.g. using wp-cli `wp option set wc_facebook_legacy_feed_file_generation_enabled no`.
3. Confirm that the `wc_facebook_regenerate_feed` is no longer pending and the feed is not regenerated. Previous feed generation interval was every fifteen minutes.
4. Opt back in - set the option to `yes` or delete it. Confirm that the feed starts regenerating again (should happen within ~30 mins?).

### Changelog entry

> New - Add option to allow larger sites to opt-out of current feed generation (product sync)

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -40,6 +40,9 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var string the WordPress option name where the external merchant settings ID is stored */
 	const OPTION_EXTERNAL_MERCHANT_SETTINGS_ID = 'wc_facebook_external_merchant_settings_id';
 
+	/** @var string Option name for disabling feed. */
+ 	const OPTION_LEGACY_FEED_FILE_GENERATION_ENABLED = 'wc_facebook_legacy_feed_file_generation_enabled';
+
 	/** @var string the WordPress option name where the feed ID is stored */
 	const OPTION_FEED_ID = 'wc_facebook_feed_id';
 
@@ -3174,6 +3177,26 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		 */
 		return (bool) apply_filters( 'wc_facebook_is_product_sync_enabled', 'yes' === get_option( self::SETTING_ENABLE_PRODUCT_SYNC, 'yes' ), $this );
 	}
+
+	/**
+ 	 * Return true if (legacy) feed generation is enabled.
+ 	 *
+ 	 * Feed generation for product sync is enabled by default, and generally recommended.
+ 	 * Large stores, or stores running on shared hosting (low resources) may have issues
+ 	 * with feed generation. This option allows those stores to disable generation to
+ 	 * work around the issue.
+ 	 *
+ 	 * Note - this is temporary. In a future release, an improved feed system will be
+ 	 * implemented, which should work well for all stores. This option will not disable
+ 	 * the new improved implementation.
+ 	 *
+ 	 * @since 2.5.0
+ 	 *
+ 	 * @return bool
+ 	 */
+ 	public function is_legacy_feed_file_generation_enabled() {
+ 		return (bool) ( 'yes' === get_option( self::OPTION_LEGACY_FEED_FILE_GENERATION_ENABLED, 'yes' ) );
+ 	}
 
 
 	/**
--- a/includes/Products/Feed.php
+++ b/includes/Products/Feed.php
@@ -158,10 +158,14 @@ class Feed {
 	 */
 	public function schedule_feed_generation() {
 
-		$integration = facebook_for_woocommerce()->get_integration();
+		$integration   = facebook_for_woocommerce()->get_integration();
+		$configured_ok = $integration && $integration->is_configured();
 
-		// only schedule if configured
-		if ( ! $integration || ! $integration->is_configured() || ! $integration->is_product_sync_enabled() ) {
+		// Only schedule feed job if store has not opted out of product sync.
+		$store_allows_sync = $configured_ok && $integration->is_product_sync_enabled();
+		// Only schedule if has not opted out of feed generation (e.g. large stores).
+		$store_allows_feed = $configured_ok && $integration->is_legacy_feed_file_generation_enabled();
+		if ( ! $store_allows_sync || ! $store_allows_feed ) {
 			as_unschedule_all_actions( self::GENERATE_FEED_ACTION );
 			return;
 		}


facebook-commerce.php
```diff
@@ -3175,6 +3178,26 @@ public function is_product_sync_enabled() {
...
+ 	 * Note - this is temporary. In a future release, an improved feed system will be
+ 	 * implemented, which should work well for all stores. This option will not disable
+ 	 * the new improved implementation.
+ 	 *
+ 	 * @since 2.5.0
+ 	 *
+ 	 * @return bool
+ 	 */
+ 	public function is_legacy_feed_file_generation_enabled() {
+ 		return (bool) ( 'yes' === get_option( self::OPTION_LEGACY_FEED_FILE_GENERATION_ENABLED, 'yes' ) );
```
haszari, line 3198: Keen for feedback on how this option works. 

- Our immediate, urgent need is to disable the current feed generation (I'm calling this "legacy" in the code/option name).
- We also want to ship an improved feed engine soon - should the option affect that?

To limit the scope of this option, I've decided it should only affect the legacy feed system. When we merge the new improved version, we will need to remove the logic for the option too.

### Alternatives
- Add two options - enable/disable, and "mode". Like a feature flag, so we can ship the new feed engine in parallel and sites can opt-in to the new or old version. This would allow us to ship early and test improvements without them affecting customers. This is most flexible, but might be overkill.
- Add a single "mode" option which includes some combinations of enable/disable and engine mode, e.g. `disable | allow_default | allow_experimental`. This is a feature flag style approach, but with 'stable' (aka default) and 'experimental'; at some point the new engine code would move from `experimental` to `default` and apply to all stores.

includes/Products/Feed.php
```diff
@@ -158,10 +158,14 @@ public function regenerate_feed() {
...
-		// only schedule if configured
-		if ( ! $integration || ! $integration->is_configured() || ! $integration->is_product_sync_enabled() ) {
+		$integration   = facebook_for_woocommerce()->get_integration();
+		$configured_ok = $integration && $integration->is_configured();
+
+		// Only schedule feed job if store has not opted out of product sync.
+		$store_allows_sync = $configured_ok && $integration->is_product_sync_enabled();
+		// Only schedule if has not opted out of feed generation (e.g. large stores).
+		$store_allows_feed = $configured_ok && $integration->is_legacy_feed_file_generation_enabled();
+		if ( ! $store_allows_sync || ! $store_allows_feed ) {
```
budzanowski, line 168:
```suggestion
		if ( ! ( $store_allows_sync && $store_allows_feed ) ) {
```
I think that this way we are more explicit about the intention.

facebook-commerce.php
budzanowski: > We also want to ship an improved feed engine soon - should the option affect that?

No. The control structure for the new feed is in a totally different place.

> Add two options - enable/disable, and "mode". Like a feature flag, so we can ship the new feed engine in parallel and sites can opt-in to the new or old version. This would allow us to ship early and test improvements without them affecting customers. This is most flexible, but might be overkill.

This would confuse people. Besides if the new feed works why have the old one still around?

> Add a single "mode" option which includes some combinations of enable/disable and engine mode, e.g. disable | allow_default | allow_experimental. This is a feature flag style approach, but with 'stable' (aka default) and 'experimental'; at some point the new engine code would move from experimental to default and apply to all stores.

That could be an interesting idea but it should be introduced with an "experimental" build of the plugin and not by a flag.

budzanowski:
This needs a paragraph in the documentation.

budzanowski (approved):

danielbitzer:
This looks good to me. I think we should hold off on considering whether to implement a "new-feed feature flag" until the new feed is in a usable state.

includes/Products/Feed.php
haszari: The suggested change would be clearer IMO if the if condition was the positive case, but since we have to negate that undermines the clarity. I'm leaving as is - the idea is that if either option is disabled, we'll bail - and the code reads like that. If that doesn't make sense let me know, we can revisit :)

facebook-commerce.php
haszari: Sounds good - we can look at ways to selectively enable the new engine in future (in line with your and @danielbitzer feedback).

haszari:
>  This needs a paragraph in the documentation.

Good point - thanks @budzanowski. Added `documentation`  label for now, I'll handle updating docs as part of 2.5.0.

Note - I don't think this should be promoted and documented as a regular option - it's a temporary workaround for specific (large) stores. IMO the main thing is that it's documented for tech support and in troubleshooting section of doc.

haszari:
If anyone is looking for docs on how to use this option, here's a brief outline.

To disable the legacy feed engine, you'll need to set the option to `no` on your store. For more information about WordPress options, check out [the documentation](https://developer.wordpress.org/plugins/settings/options-api/).

You can do this with WP-CLI, or with a mu plugin. For example:

```shell
wp option set wc_facebook_legacy_feed_file_generation_enabled no
```

```php
<?php
/* mu-plugins/disable-legacy-facebook-feed.php */

function yourprefix_disable_facebook_legacy_feed() {
  update_option( 'wc_facebook_legacy_feed_file_generation_enabled', 'no' );
}
add_action( 'init', 'yourprefix_disable_facebook_legacy_feed' );
```

If you need to re-enable the feed again, simply delete the option.

PR #1933, opened 2021-05-07
Create a single class responsible for validating whether a product should be synced to FB

### LINKED ISSUE #1929: Create product sync validator
A class or similar that validates whether a product should be synced to Facebook.

This should support both the batch API update and the feed.

Part of #1904 epic.

danielbitzer:

This PR refactors a lot of existing code into a single class that is soley responsible for validating whether a product should be synced to FB. Previously this logic was spread between classes that contained a lot of other unrelated logic and was rather confusing. 

Since this was only a refactor we may want to postpone this to 2.6.0.

The only addition was to validate the product stock status. I noticed that i[n other parts of the code](https://github.com/woocommerce/facebook-for-woocommerce/blob/f3cf0d7417bcc492597aef23d392ff4f6c076136/includes/Products.php#L228-L241) if a product is out of stock it get deleted from FB so it makes sense to include in the validator I think.

Closes #1929.

TBA

### Changelog entry

> Dev - Create a single class responsible for validation whether a product should be synced to Facebook


includes/ProductSync/ProductValidator.php
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 * Check whether the product's status excludes it from sync.
+	 *
+	 * @param WC_Product $product A product object.
+	 *
+	 * @throws ProductExcludedException If product should not be synced.
+	 */
+	protected function validate_product_status( WC_Product $product ) {
+		$product = $product->get_parent_id() ? wc_get_product( $product->get_parent_id() ) : $product;
```
budzanowski, line 65: We have multiple wc_get_product which is quite costly for our needs. It should be cached ( or not, we would need to verify ) but still, there is an overhead.

budzanowski:
I like the structure of the refactor. Though I don't like the use of exceptions.

> An exception is an object that describes an error or unexpected behaviour of a PHP script.

from https://www.w3schools.com/php/php_exceptions.asp

There is nothing erroneous or unexpected about a product being excluded from synchronization.

includes/ProductSync/ProductValidator.php
danielbitzer: Any database calls would be object cached but we could get the parent product in the class constructor to cut the product object instantiation from 4 down to 1.

danielbitzer:
Exceptions can be useful for for validation purposes as well as for errors. But we don't have to do it like this, we could also just catch the exception and return a boolean from the `validate` method. I used to not like exceptions too but changed my mind after seeing them used well in an application.

The main benefit of using them here is the calling code can know _why_ the product is excluded because there is a message attached to each exception. Otherwise it's just `true` or `false` right? Although we don't need this now it's practical to retain this info. In the future we may want to display why a product didn't sync to a user or as a developer it helps narrow down why a product was excluded.

includes/ProductSync/ProductValidator.php
danielbitzer: See f3cf0d7417bcc492597aef23d392ff4f6c076136 for a version that only gets the parent product once. I think this is cleaner too.

includes/ProductSync/ProductValidator.php
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	/**
+	 * Check whether the product's status excludes it from sync.
+	 *
+	 * @throws ProductExcludedException If product should not be synced.
+	 */
+	protected function validate_product_status() {
+		$product = $this->product_parent ?: $this->product;
```
budzanowski, line 105: Using short ternaries is not allowed (WordPress.PHP.DisallowShortTernary.Found)
I wonder why this is was not annotated.

includes/ProductSync/ProductValidator.php
```diff
@@ -0,0 +1,203 @@
...
+		$this->validate_product_categories_and_tags();
+	}
+
+	/**
+	 * Check whether product sync is globally disabled.
+	 *
+	 * @throws ProductExcludedException If product should not be synced.
+	 */
+	protected function validate_sync_enabled_globally() {
+		if ( $this->integration->is_product_sync_enabled() ) {
```
budzanowski, line 94:
```suggestion
		if ( ! $this->integration->is_product_sync_enabled() ) {
```

includes/ProductSync/ProductValidator.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @throws ProductExcludedException If product should not be synced.
+	 */
+	public function validate_product_sync_field() {
+		$invalid_exception = new ProductExcludedException( 'Sync disabled in product field.' );
+
+		if ( $this->product->is_type( 'variable' ) ) {
+			foreach ( $this->product->get_children() as $child_id ) {
+				$child_product = wc_get_product( $child_id );
+				if ( $child_product && 'no' !== $child_product->get_meta( self::SYNC_ENABLED_META_KEY ) ) {
+					break;
```
budzanowski, line 173:
```suggestion
					return;
```
I think that the logic should check for at least one child with sync enabled. I think that this was the intention.
With the `break;`  the code is meaningless because we will always reach `throw`.

includes/ProductSync/ProductExcludedException.php
```diff
@@ -0,0 +1,12 @@
...
+namespace SkyVerge\WooCommerce\Facebook\ProductSync;
+
+use Exception;
+
+/**
+ * Class ProductExcludedException
+ *
+ * Exception for when a product is excluded from Facebook product sync.
+ */
+class ProductExcludedException extends Exception {}
```
haszari, line 12: Might be clearer to name this `ProductSyncExcludeException`.

haszari: Similarly, I see the exception message strings don't mention sync - I think it's clearer to mention sync everywhere, so it's clear what this is part of.

includes/Products.php
```diff
@@ -193,16 +186,20 @@ public static function disable_sync_for_products_with_terms( array $args ) {
...
 	 * @return bool
 	 */
 	public static function product_should_be_synced( \WC_Product $product ) {
-
-		return 'publish' === $product->get_status() && self::published_product_should_be_synced( $product );
+		try {
+			facebook_for_woocommerce()->get_product_sync_validator( $product )->validate();
+			return true;
+		} catch ( \Exception $e ) {
+			return false;
```
haszari, line 201: I guess could log here if verbose debug logging is enabled - make use of those exception messages. This would make things a bit more transparent too, so merchants can diagnose why some product isn't syncing.

haszari:
This is looking good (note - I haven't tested). Organising all the product sync filtering logic to one place is a huge improvement. 😍 

> Since this was only a refactor we may want to postpone this to 2.6.0.

I think this is going to make it easier to implement fixes to feed or product sync in general. We should prioritise getting it merged so help accelerate that effort.

Having said that .. we have limited time before 2.5.0 and this is a fairly extensive refactor (as you've noted). 

Ideally we'd have e2e or PHPUnit tests (WP style integration tests) to confirm the behaviour. If we delay to 2.6, perhaps we can implement some PHPUnit tests to confirm this doesn't change behaviour of the refactored code paths, or if that's too complex, just tests of the (new) validator interface. (Not sure what testing infra we have set up, this might be more effort than we want to spend right now.)

### Use of exceptions
I agree in principle that exceptions are best used for exceptional (unexpected) situations. As a hard requirement, exceptions shouldn't break request processing in normal use.

Exceptions are a good option for a standard interface for exposing richer error info to client (calling) code (as you've done here). Here, that makes the code / logic flow simpler (no need to check for failure repeatedly), and could even benefit merchants if we log the messages. 

### Next steps
I would love to get this shipped in 2.5.0. If we log the error messages, that's a nice little enhancement.

@budzanowski and @danielbitzer can you work out a plan for whether to ship this in 2.5 or 2.6, and how to test. We'll need the same effort/work to get it shipped no matter which version. As a minimum we could probably run some manual tests confirming that the same products sync before and after this PR; ideally we have automated tests in place that protect us in the future.

class-wc-facebookcommerce.php
```diff
@@ -924,6 +925,22 @@ public function get_profiling_logger() {
...
 		}
 
+		/**
+		 * Get the product sync validator class.
+		 *
+		 * @param WC_Product $product A product object to be validated.
+		 *
+		 * @return ProductSyncValidator
+		 */
+		public function get_product_sync_validator( WC_Product $product ) {
```
budzanowski, line 935: We can't operate this way.  `get_product_sync_validator` will not create a fresh instance when we will try to validate another product. We are stuck with the original `$product` that was passed to the constructor.
Functions like `add_product_list_table_columns_content` need a new instance each time - used for products table display. Or if we want to keep a single class then we need to pass `$product` to `validate()` and not use it in the constructor.

includes/Products.php
budzanowski: Please don't do that. You don't want to log here. There are legitimate reasons why products should not be synced that do not issue. Global sync disabled, product sync disabled, excluded categories. These may be intentional settings. 

If we want to use it we should be a bit more direct about this. I will create a PR that will use this information for the

budzanowski: Please check https://github.com/woocommerce/facebook-for-woocommerce/pull/1937. This is how we should utilize the information from this code.

budzanowski:
> Exceptions can be useful for for validation purposes as well as for errors. But we don't have to do it like this, we could also just catch the exception and return a boolean from the validate method. I used to not like exceptions too but changed my mind after seeing them used well in an application.

Of course, I will not oppose the use of exceptions - I am just voicing my personal preferences. There is nothing wrong with the products, also they are not valid or invalid in any way. The original name was more straight to the point about the intentions of the code. `product_should_be_synced` is a very clear indication of what was happening.

That being said the structure itself is very good. The separation of tests into separate functions makes it easier to grasp the meaning of each section. And as @haszari has mentioned it will make spotting errors and making fixes easier.

I have pointed out some problems that I have observed that will need more work. The original code was super convoluted so getting this right is very challenging.

> I would love to get this shipped in 2.5.0. If we log the error messages, that's a nice little enhancement.

As I have mentioned here: https://github.com/woocommerce/facebook-for-woocommerce/pull/1933#discussion_r629238086 we should not log. For example, if global sync gets disabled and the merchants enter the products view we will generate a log entry for each item in the table. That is not very useful. Instead, we should display this information directly. Please check #1937 for how I think we should do this.

includes/ProductSync/ProductValidator.php
danielbitzer: Oops my mistake. Fixed in c90ffca010ad6c8dd29e7e18a49fe01e9b029768.

includes/Products.php
danielbitzer: Adding logging here would be very noisy due to the frequency of daily feed generation so we might not want to do that. Although it could be useful for debugging/dev purposes. Something like #1937 is a nicer solution for users.

includes/ProductSync/ProductExcludedException.php
danielbitzer: Can do. These classes are in the `ProductSync` namespace but I'm not against includeing "sync". Actually `ExcludedFromSyncException` could be a helpful name for this class.

includes/ProductSync/ProductValidator.php
budzanowski: > Personally, I think we should globally allow short ternaries as they are very useful.
Although I agree about the usefulness, if we want to change this we should convince the WP community to this change. Keeping standards helps us interact better with the outside world. https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

danielbitzer:
@haszari my preference is to aim for 2.6.

includes/ProductSync/ProductValidator.php
```diff
@@ -0,0 +1,203 @@
...
+
+	/**
+	 * "allow simple or variable products (and their variations) with zero or empty price - exclude other product types with zero or empty price"
+	 * unsure why but that's what we're doing
+	 *
+	 * @throws ProductExcludedException If product should not be synced.
+	 */
+	protected function validate_product_price() {
+		// Permit simple and variable products to have an empty price
+		if ( in_array( $this->product_parent->get_type(), array( 'simple', 'variable' ), true ) ) {
```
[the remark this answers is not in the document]
danielbitzer: Thanks, fixed in ccf12eefce54ac15fcb53da91e4aad4945761a32

danielbitzer:
> That being said the structure itself is very good. 

> This is looking good (note - I haven't tested). Organising all the product sync filtering logic to one place is a huge improvement. 😍

Thanks guys! I agree this is going to help us a lot going forward in building a solid product sync solution.

> There is nothing wrong with the products, also they are not valid or invalid in any way. 

My reasoning behind the naming is that this class is _validating_ whether a product should be synced. It's synonymous with "checking", "confirming", "verifying" which fits the purpose of the class. I like to state a class's purpose, where possible, in it's name. E.g. "JobRegistry" where jobs are registered/initiated rather than just "Jobs". It makes the code more readable IMO.

> The original name was more straight to the point about the intentions of the code. `product_should_be_synced` is a very clear indication of what was happening.

Are you suggesting the class be called `ProductShouldBeSynced`? Naming things is hard but one concern is we are using namespaces and this class is contained in the `ProductSync` namespace. It's not ideal for every class in the `ProductSync` namespace to repeat "ProductSync" in it's name. On the other hand, I like being as descriptive as possible so the purpose of a class can be easily known from is name so I'm no against a longer name. The main thing I'm thinking is that using namespaces well lets us have shorter class names.

I've tried a few different namings for the validator and exception classes but ended up back at my original suggestions:
- `ProductSync\ProductValidator`
- `ProductSync\ProductExcludedException`

If the namespace isn't clear enough we could name them:
- `ProductShouldBeSyncedValidator`
- `ProductExcludedFromSyncException`

> Ideally we'd have e2e or PHPUnit tests

This code is definitely a good candidate for unit tests. There would be some setup effort needed for this.

> Exceptions are a good option for a standard interface for exposing richer error info to client (calling) code (as you've done here). Here, that makes the code / logic flow simpler (no need to check for failure repeatedly)

Totally agree! Not needing to check for failure repeatedly was a big reason behind my use of exceptions. Using `if` conditionals excessively can get very tedious IMO.

danielbitzer:
@haszari @budzanowski could you please take a look at my previous comment?

budzanowski:
> Are you suggesting the class be called ProductShouldBeSynced?

No, the method `validate()`.

We have `ProductValidator->validate()` and that is quite incomprehensive.

Even if we take namespaces for consideration it still is quite vague given no context. In your code, you are using the class instance fetcher function `get_product_sync_validator` and import the class with `as` and add Sync to the name to make ir more obvious.

That is just naming. We should not focus on it that much in this area. This is an internal interface. I know that they are people that will die over naming stuff - I don't plan to do that. This is good as it is IMO.

budzanowski (approved):

budzanowski:
> Are you suggesting the class be called ProductShouldBeSynced?

No, the method `validate()`.

We have `ProductValidator->validate()` and that is quite incomprehensive.

Even if we take namespaces for consideration it still is quite vague given no context. In your code, you are using the class instance fetcher function `get_product_sync_validator` and import the class with `as` and add Sync to the name to make ir more obvious.

That is just naming. We should not focus on it that much in this area. This is an internal interface. I know that they are people that will die over naming stuff - I don't plan to do that. This is good as it is IMO.

haszari:
One comment on the interface and making it more consistent. I see you have two kinds of methods:

- `validate_*` which throws if product should not be synced
- `is_excluded_* which returns false if product should not be synced

There's no `is_excluded` for the main use - this code (`product_should_be_synced()`) uses `validate()`. It might be more consistent to add an `is_excluded_from_sync` as well, and treat the `is_excluded_*` methods as the high-level api.

Also, it might be clearer to flip the logic in `is_excluded_*` so they return validation status - false if excluded. Then the name could be made more obvious, e.g. 

```php
if ( ! get_product_sync_validator( $product )->product_data_allows_sync(); ) { bail }
if ( ! get_product_sync_validator( $product )->product_category_tag_allows_sync(); ) { bail }
```

As noted by @budzanowski  - naming and API design is tricky, and not a blocker. Up to you if you want to revise the naming/logic direction, the above is just one idea. 

Aside from that, this PR looks fine to me, though we'll need to thoroughly test somehow (even if manual), especially if this is affecting existing sync behaviour (batch api and legacy feed). This can be follow up, just need to make sure we've independently confirmed the sync inclusion logic is working as we expect.

danielbitzer:
Thanks for the comments! I will have a think about naming. 

I think there may be an issue with this PR. How do we handle removing products when they need to be? For example:
- a product is synced
- the product's **FB sync field** is changed to "do not sync" 
- this validator class now tells us this class is not valid so we won't include it in the field or batch API
- but how does facebook know to remove the product?

haszari:
> How do we handle removing products when they need to be removed from FB?

Good question! This may be a wider issue – bigger than validating whether an individual product should be synced. There may be existing issues with this too; I haven't tested this much so I don't know what to expect.

danielbitzer:
We should possibly be thinking about (and naming) this class as being responsible for determining a product's **VISIBILTY** on Facebook not whether it **should sync**. And instead we say "all products sync to Facebook" but we hide products based on specific criteria. Basically we need to ensure that products can be correctly removed from FB when their properties change in Woo.

For example:

https://github.com/woocommerce/facebook-for-woocommerce/blob/604d7551701afa281a1acb8fbba8dce0ea9c7f4d/facebook-commerce.php#L1141

budzanowski:
> but how does facebook know to remove the product?

It depends on the feed configuration:

```The configuration for fetching a feed in a recurrant manner. If it's an update_schedule, the scheduled uploads would update existing items with the information delta provided or create new ones if not present. If its a replace schedule, this will be a full replace operation where items not present in the feed file will be removed, new items will be created and existing items will be updated.```

We would be using replace schedule with complete information for each product. This means that items missing from the feed would be removed.

If the product is in `do not sync` it should not be in the FB catalog. Removing it from the feed file will do the trick.

**Comment by danielbitzer** [comment]
Circling back to this, I'm thinking that we put aside the conversation about removing deleted products from the feed for now and leave this PR as only a refactor. In which case we can merge this for 2.6 before the rest of the feed work is implemented. @haszari @budzanowski thoughts?

**Comment by danielbitzer** [comment]
@haszari:

> There's no `is_excluded` for the main use - this code (`product_should_be_synced()`) uses `validate()`. It might be more consistent to add an `is_excluded_from_sync` as well, and treat the `is_excluded_*` methods as the high-level api.

If we do this then calling code won't be able to access the exception validation messages like they are being used in #1937.

I mainly added the `is_excluded` methods to ease integration with the legacy product sync code but I think moving forward calling the `validate()` method would be the recommended option.

Anyway, I've gone ahead and added a `..ProductSync\ProductValidator::passes_all_checks()` method that returns a boolean. It could be a time saver.

> Also, it might be clearer to flip the logic in `is_excluded_*` so they return validation status - false if excluded. Then the name could be made more obvious, e.g.

Good point, I agree it would be clearer. Updated in my recent couple of commits.

**End result class interface:**

```php

$product_sync_validator = facebook_for_woocommerce()->get_product_sync_validator( $product );

if ( $product_sync_validator->passes_all_checks() ) { /**  */ }
if ( $product_sync_validator->passes_product_terms_check() ) { /**  */ }
if ( $product_sync_validator->passes_product_sync_field_check() ) { /**  */ }

// or 

$product_sync_validator->validate();
```

budzanowski:
This was heavily reviewed. We will evaluate this in 2.6 RC. Merging.

PR #1937, opened 2021-05-10
Add/sync status information to metabox.

budzanowski:

Using code from #1933 take the no-sync reason information and add display it in case the product will not be synced.

Closes # .

1. Open a product for edit.
2. Disable sync. 
3. Observe metabox.
image
Or
image

### Changelog entry

New - Add no synchronization reason to the product edit screen in the Facebook meta box.

### DIFF
--- a/includes/ProductSync/ProductValidator.php
+++ b/includes/ProductSync/ProductValidator.php
@@ -91,7 +91,7 @@ class ProductValidator {
 	 * @throws ProductExcludedException If product should not be synced.
 	 */
 	protected function validate_sync_enabled_globally() {
-		if ( $this->integration->is_product_sync_enabled() ) {
+		if ( ! $this->integration->is_product_sync_enabled() ) {
 			throw new ProductExcludedException( 'Product sync is globally disabled.' );
 		}
 	}
@@ -170,7 +170,7 @@ class ProductValidator {
 			foreach ( $this->product->get_children() as $child_id ) {
 				$child_product = wc_get_product( $child_id );
 				if ( $child_product && 'no' !== $child_product->get_meta( self::SYNC_ENABLED_META_KEY ) ) {
-					break;
+					return;
 				}
 			}
 
--- a/includes/ProductSync/ProductValidator.php
+++ b/includes/ProductSync/ProductValidator.php
@@ -190,8 +190,9 @@ class ProductValidator {
 	 * @throws ProductExcludedException If product should not be synced.
 	 */
 	protected function validate_product_price() {
-		// Permit simple and variable products to have an empty price
-		if ( in_array( $this->product_parent->get_type(), array( 'simple', 'variable' ), true ) ) {
+		// Permit simple and variable products to have an empty price.
+		$product = $this->product->is_type( 'variation' ) ? wc_get_product( $this->product->get_parent_id() ) : $this->product;
+		if ( in_array( $product->get_type(), array( 'simple', 'variable' ), true ) ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


haszari:
This looks like a potentially useful improvement. @budzanowski can you log a "feature request" issue stating the problem?

danielbitzer:
Cool! If we do this I'll need to make all my exceptions translatable strings. Currently they're not translatable since I was intending them as a message for developers.

budzanowski:
>  If we do this I'll need to make all my exceptions translatable strings. 

@danielbitzer this PR does that so you don't have to :)

danielbitzer:
> this PR does that so you don't have to

🤦 I missed that. Cool!

haszari:
Tentatively adding this to 2.7 - looks like it will be a useful part of a revamped feed/sync engine. However there's no urgency to improve the UI here – the priority is reliable sync for large and small stores.

danielbitzer:
@budzanowski please note the `ProductSync\ProductValidator` class from #1933 had a few changes that conflict with this PR. I also opened #1997 to make the exception messages translatable which will hopefully make merging this PR simpler since you can just remove all the changes to the `ProductValidator` class.

budzanowski:
image

I have added support for variations.
This is now ready for review.

rcstr (approved):
works great. Tested excluding categories and tags, all good, same with variations.

Screen Shot 2021-06-09 at 17 42 03

budzanowski:
Test comment

includes/ProductSync/ProductValidator.php
```diff
@@ -256,4 +289,79 @@ protected function validate_product_price() {
...
+		$description = $this->facebook_product->get_fb_description();
+
+		/*
+		 * Requirements:
+		 * - No all caps descriptions.
+		 * - Max length 5000.
+		 * - Min length 30 ( tested and not required, will not enforce until this will become a hard requirement )
+		 */
+		if ( \WC_Facebookcommerce_Utils::is_all_caps( $description ) ) {
+			throw new ProductExcludedException( __( 'Product description is all capital letters. Please change the description to sentence case in order to allow synchronization of your product.', 'facebook-for-woocommerce' ) );
```
danielbitzer, line 314: I think it would be nice to introduce a different type of exception to handle these new validations because they are related to invalid data entry rather than to valid reasons like  "being out of stock" or "being part of an excluded category".

My suggestion would be something like this:
- `ProductInvalidException` for cases when the product fails validation for a reason that a property is invalid and "needs fixing by the user"  
- `ProductExcludedException` for cases when the product is excluded for an expected reason and doesn't need fixing 

This could be useful in the UI. For example if the product is "needs fixing" we could display a different color or icon to the user vs if the product is just excluded because it's out of stock.

budzanowski: > This is a nice addition 👍
>
> This facebook-commerce.php file is 4000 lines long and is one of the messy dumping grounds in this plugin. It would be nice if we could extract this meta box into a separate class. Perhaps in the 'Admin' namespace?
I'm wondering if we can change the heading of the meta box to something more descriptive like "Facebook Product Feed"?


Agree. Done in 06737dc

danielbitzer (approved):
Looks good and is working well 👍 

There are a few minor PHPCS issues in`Product_Sync_Meta_Box.php`...

solstudioim:
Hi team, thanks for bringing the metabox that displays product validation issues.

I came across an issue where if a variable product has disallowed characters like `:` the variable product won't get synced.

In the log, the following is visible:

`body: {“validation_status”:[{“retailer_id”:“vela-7-med-mad_147”,“errors”:[{“message”:“The additional_variant_attribute information under  is invalid. Review for more details”}]},{“retailer_id”:“vela-7-med-alg_223",“errors”:[{“message”:“The additional_variant_attribute information under  is invalid. Review for more details”}]}
`
Found in the setting of the user:

Markup 2021-11-26 at 08 21 08

Do you think it's worth adding this to the list of status in the metabox for product validation issues?

I checked in our doc, this isn't mentioned: https://woocommerce.com/document/facebook-for-woocommerce/#section-27

Thanks a lot!

PR #1939, opened 2021-05-11
Move to admin init.

### LINKED ISSUE #1632: Move Feed Actions away from init hook
### Description of observed issue
The latest version of the plugin (2.0.5) hooks `SkyVerge\WooCommerce\Facebook\Products\Feed::schedule_feed_generation()` on init [here](https://github.com/facebookincubator/facebook-for-woocommerce/blob/master/includes/Products/Feed.php#L57). 

This means that this method can be fired with any frontend page request (Example: single product page).

The method uses [as_next_scheduled_action](https://github.com/facebookincubator/facebook-for-woocommerce/blob/master/includes/Products/Feed.php#L179) to check if the `wc_facebook_regenerate_feed` action is scheduled.

The problem is that this method uses a query that is very slow on sites with large number of scheduled actions. On a site with 500,000 scheduled actions rows the query for this check added 4 seconds to the transaction time for a single product page.

Here is a NewRelic trace of the next scheduled action query that dds 4,460 ms to the page load: 

New Relic One 2020-10-13 15-56-45

### Suggested change

Here are suggestions for reducing the impact this check has on frontend pages:
- Move the hook to `admin_init` so only admin 
- Add a check for `! is_admin()` and exit.
- Run this check via cron? 

We will probably be implementing something one of these on this site, but it would be nice to see this added to the plugin.

budzanowski:

Using the suggestion from https://github.com/woocommerce/facebook-for-woocommerce/issues/1632#issuecomment-837679022

We are moving to admin_init to reduce the load.

Closes #1632 .

1. Plugin should operate normally and the feed should still be scheduled.

### Changelog entry

Fix - Move feed scheduling to admin_init from init for better performance.

### DIFF
--- a/includes/Products/Feed.php
+++ b/includes/Products/Feed.php
@@ -54,7 +54,7 @@ class Feed {
 	private function add_hooks() {
 
 		// schedule the recurring feed generation
-		add_action( 'init', array( $this, 'schedule_feed_generation' ) );
+		add_action( 'admin_init', array( $this, 'schedule_feed_generation' ) );
 
 		// regenerate the product feed
 		add_action( self::GENERATE_FEED_ACTION, array( $this, 'regenerate_feed' ) );


danielbitzer (approved):
This is a good temporary solution to reduce the performance impact of this plugin 👍

rcstr (approved):
Great improvement! 🚀 🚀 🚀

PR #1940, opened 2021-05-11
Delete the job from the options table after the job has finished.

### LINKED ISSUE #1943: Syncing leaves useless `wc_facebook_background_product_sync_job` rows in options table.
## 🔎 Isolate the bug
@budzanowski was able to confirm and isolate the error along with a proposed fix in PR #1940. 

## ✍️ Describe the bug
"After the Background Batch API job is completed the option that was used as the control structure for the process remains in the options table. We have no use for it."

This error resulted in a forum user having thousands of unneeded `wc_facebook_background_product_sync_job` rows in their database's options table: https://wordpress.org/support/topic/wc_facebook_background_product_sync_job/

Every background sync job [adds an option to wp_options table with job data, metadata and status info](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/utilities/class-sv-wp-background-job-handler.php#L402). These options are not cleaned up automatically, so they accumulate over time.

This is likely a performance / database size issue on many stores. The job data includes a list of products to update. So for larger stores with many products, the options will be larger. Because these options accumulate over time, longer-lived stores will have more redundant job rows in their database.

## 🚶‍♀️ Steps to reproduce
Surmised from PR: 
1. Open the options table in the DB and look for `wc_facebook_background_product_sync_job_%` where % is a number
2. Start the Sync, wait till the end,
3. Observe that the new/existing `wc_facebook_background_product_sync_job_%` options were not removed, leading to the accumulation of rows.

Since @budzanowski reproduced this bug and already filed a PR, I have not personally tested these steps.

## ✔️ Expected behavior
There should not be unused, leftover rows once the sync job is completed.

## 🗃 Logs
None at this time.

budzanowski:
Closes #1943

After the Background Batch API job is completed the option that was used as the control structure for the process remains in the options table. We have no use for it. IMO we should just delete it.

More information: 
- https://wordpress.org/support/topic/wc_facebook_background_product_sync_job/
- 3974081-zd-woothemes

1. Open options in the db and look for `wc_facebook_background_product_sync_job_%` where % is a number
2. Start the Sync, wait till the end,
3. Look again for `wc_facebook_background_product_sync_job_%` no new options should be added.
4. Options should not be in the DB.

### Changelog entry

 Fix - Delete the job from the options table after the job has finished.

### DIFF
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -108,7 +108,7 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 	 * @since 2.6.0
 	 *
 	 * @param \stdClass|object|string $job Job instance or ID.
-	 * @return \stdClass|object|false on failure
+	 * @return \stdClass|object|false on failure.
 	 */
 	public function complete_job( $job ) {
 
@@ -126,7 +126,6 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 		$this->delete_job( $job );
 		return $job;
 	}
-
 
 	/**
 	 * Processes multiple items.
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -102,6 +102,31 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 		return $job;
 	}
 
+	/**
+	 * Handles job completion.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @param \stdClass|object|string $job Job instance or ID.
+	 * @return \stdClass|object|false on failure
+	 */
+	public function complete_job( $job ) {
+
+		if ( is_string( $job ) ) {
+			$job = $this->get_job( $job );
+		}
+
+		if ( ! $job ) {
+			return false;
+		}
+
+		$job->status       = 'completed';
+		$job->completed_at = current_time( 'mysql' );
+
+		$this->delete_job( $job );
+		return $job;
+	}
+
 
 	/**
 	 * Processes multiple items.
--- a/includes/Utilities/Background_Handle_Virtual_Products_Variations.php
+++ b/includes/Utilities/Background_Handle_Virtual_Products_Variations.php
@@ -96,6 +96,30 @@ class Background_Handle_Virtual_Products_Variations extends Framework\SV_WP_Back
 		return $job;
 	}
 
+	/**
+	 * Handles job completion.
+	 *
+	 * @since 2.6.0
+	 *
+	 * @param \stdClass|object|string $job Job instance or ID.
+	 * @return \stdClass|object|false on failure.
+	 */
+	public function complete_job( $job ) {
+
+		if ( is_string( $job ) ) {
+			$job = $this->get_job( $job );
+		}
+
+		if ( ! $job ) {
+			return false;
+		}
+
+		$job->status       = 'completed';
+		$job->completed_at = current_time( 'mysql' );
+
+		$this->delete_job( $job );
+		return $job;
+	}
 
 	/**
 	 * Counts the number of virtual products or product variations with sync enabled and visible.
--- a/includes/Utilities/Background_Remove_Duplicate_Visibility_Meta.php
+++ b/includes/Utilities/Background_Remove_Duplicate_Visibility_Meta.php
@@ -80,7 +80,6 @@ class Background_Remove_Duplicate_Visibility_Meta extends Framework\SV_WP_Backgr
 			}
 		}
 
-		// job complete! :)
 		if ( $this->count_remaining_products() === 0 ) {
 
 			update_option( 'wc_facebook_background_remove_duplicate_visibility_meta_complete', 'yes' );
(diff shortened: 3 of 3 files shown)


danielbitzer (commented):
Cool, it definitely makes sense to me to be deleting these jobs after they run. I'm surprised they are kept for forever!

There are 2 other jobs extending the `\SkyVerge\WooCommerce\PluginFramework\v5_10_0\SV_WP_Background_Job_Handler` class. Should we be deleting the option rows for those jobs also once they are completed?

danielbitzer:
One alternative would be, like Action Scheduler, to delete the records X number of days after the job is completed. That way it's possible to view recent job activity.

budzanowski:
> One alternative would be, like Action Scheduler, to delete the records X number of days after the job is completed. That way it's possible to view recent job activity.

It could be a possibility but I don't see any usage for this. If we would like to store some sort of statistical information we should be more deliberate with our approach.

budzanowski:
> There are 2 other jobs extending the \SkyVerge\WooCommerce\PluginFramework\v5_10_0\SV_WP_Background_Job_Handler class. Should we be deleting the option rows for those jobs also once they are completed?

Yes definitely! I have just copied the code. Not the most elegant solution but I believe that we will move this onto AS framework soon.

includes/Products/Sync/Background.php
```diff
@@ -102,6 +102,30 @@ public function process_job( $job, $items_per_batch = null ) {
...
+		if ( is_string( $job ) ) {
+			$job = $this->get_job( $job );
+		}
+
+		if ( ! $job ) {
+			return false;
+		}
+
+		$job->status       = 'completed';
+		$job->completed_at = current_time( 'mysql' );
```
haszari, line 124: This method overrides [base `complete_job` from the SkyVerge framework](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/utilities/class-sv-wp-background-job-handler.php#L742), and changes the behaviour, by calling another base class lifecycle method - `delete_job`.

That seems risky to me - perhaps there's a cleaner supported way to manage the options that accumulate over time.

haszari (commented):
I took a wider look at the issue and the SkyVerge background jobs framework to understand what's happening and what we can do about it.

### The problem
Before diving into solutions I want to define the problem as I understand it.

Every background sync job adds an option to `wp_options` table with job data, metadata and status info. These options are not cleaned up, so they accumulate over time.

This is likely a performance / database size issue on many stores. The job data includes a list of products to update. So for larger stores with many products, the options will be larger. Because these options accumulate over time, longer-lived stores will have more redundant job rows in their database.

### SkyVerge plugin framework
The background job engine / framework we're using comes from [SkyVerge WooCommerce Plugin Framework](https://github.com/skyverge/wc-plugin-framework).

So my first question is - how does the framework suggest we solve this problem?

I didn't find any docs for the background sync API, other than [the source code](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/utilities/class-sv-wp-background-job-handler.php).

There is a [public `delete_job()` method](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/utilities/class-sv-wp-background-job-handler.php#L823). So I assume we're supposed to use that to clean up jobs we no longer need. 

Side note: looks like the framework is [up to 5.10.7](https://github.com/skyverge/wc-plugin-framework/releases/tag/5.10.7). I think we're on 5.10.0.

This issue might be worth reporting to SkyVerge repo.

### Possible fixes
These are the options as I see it. There are two approaches - delete immediately after job completion, or periodic clean up.

Of these, the periodic clean up is most critical IMO - we need it to fix the accumulated items. Deleting immediately after completion might be useful but we may not need it if we have a cleanup job.

Note also we should document how to manually clean these up, so merchants / tech support can manually remove all the accumulated rows.

##### Options: delete after completion
1. Override the base `complete_job`, call parent to do the normal processing, then call `delete_job`. Similar to what's implemented in this PR, but calling through to base class instead of paste-forking logic.
2. Use a hook to call `delete_job` after completion - there's a `{$this->identifier}_job_complete`.

##### Options: periodic clean up
3. Add a periodic task which searches for completed jobs and deletes them. This might be a slow query - will need to take care and do in appropriately sized batches.

## Next steps
Based on the above, I don't think we should merge this PR. I think we should implement a periodic cleanup job.

1. Implement periodic cleanup job and ship soon (2.5 / 2.6).
2. Document how to manually remove rows cleanly and safely.
3. (Optional) Log an issue or feedback with SkyVerge.

@budzanowski @danielbitzer - Keen for your thoughts on this approach, or other alternatives.

danielbitzer:
Good points @haszari!

I also looked at the Skyverge framework but didn't find any solution to clean old job records.

Arguably we could just build a periodic clean-up solution not merge this PR since that would clean up all options rows including future ones. This could be a good candidate for an Action Scheduler powered job.

In terms of priority, I don't think this issue is critical priority and we should be fine to wait until 2.6.

haszari:
Closing - the issue was fixed in #1962, this PR no longer needed. Feel free to reopen if this is still useful.

PR #1942, opened 2021-05-11
Reschedule the feed file generation to 24h.

### LINKED ISSUE #1941: Increase feed generation interval.
Currently, the feed file is generated every 15 minutes. This is much too often. Once a day sounds much more reasonable.  We don't really lose here because all the important events are still synchronized using Batch API on demand.

Task:
1. Stop current feed generation actions.
2. Reintroduce the actions but with the feed cadence set to 24h.

budzanowski:
By increasing the feed file generation interval we reduce the load that is put on the sites.

Closes #1941 .

1. In wp-content/plugins/facebook-for-woocommerce/class-wc-facebookcommerce.php change the VERSION const to `2.5.0`.
2. Refresh the admin panel.
3. Go to WooCommerce -> Status -> Scheduled Actions
4. Look for `wc_facebook_regenerate_feed`, it should be scheduled to run once a day.
image

### Changelog entry 

Fix - Change feed generation cadence to 24h.

### DIFF
--- a/includes/Lifecycle.php
+++ b/includes/Lifecycle.php
@@ -43,6 +43,7 @@ class Lifecycle extends Framework\Plugin\Lifecycle {
 			'2.0.3',
 			'2.0.4',
 			'2.4.0',
+			'2.5.0',
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


danielbitzer (approved):
Looks good, works as expected 👍

rcstr (approved):
works as expected

galapogos01:
When the feed generation task fails, it appears to reschedule itself for the next hour. So even though the task says daily, it's running (and failing) ~hourly (due to #1871).

image

Issue #1944, opened 2021-05-12
Add tracking events to key features to better understand product usage


haszari:
It would be useful to have a clearer idea of how certain features are used. For example, generating a feed file for product sync. To help us understand how this feature is used in stores, we could add telemetry events, such as: `feed_generation_started` `feed_generation_completed` `feed_requested_by_facebook` etc.

This could give us data on how often features are used, and also how they are used (e.g. config options, flows through admin UI).

PR #1945, opened 2021-05-12
add tracks proxy API and some example events for feed/background sync

haszari:
Fixes: #1944

Adds a basic tracks API to record telemetry events. Events will only be recorded for sites that opt in, and are running a supported version of WooCommerce (and Facebook for WooCommerce).

This is an experiment - the events, and the API are all up for discussion. 

1.
2.
3.

### Changelog entry

> Fix / Dev / New - Remove duplicate visibility meta entries from postmeta table

### DIFF
--- a/includes/Proxies/Tracks.php
+++ b/includes/Proxies/Tracks.php
@@ -1,7 +1,7 @@
 <?php
 declare( strict_types=1 );
 
-namespace Automattic\WooCommerce\Facebook\Proxies;
+namespace SkyVerge\WooCommerce\Facebook\Proxies;
 
 use WC_Site_Tracking;
 use WC_Tracks;
--- a/includes/Proxies/Tracks.php
+++ b/includes/Proxies/Tracks.php
@@ -1,0 +1,36 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Facebook\Proxies;
+
+use WC_Site_Tracking;
+use WC_Tracks;
+
+/**
+ * Wrapper (proxy) class for WC_Tracks from WooCommerce Core.
+ *
+ * Supplies a standard prefix to all events.
+ *
+ * @package Automattic\WooCommerce\Facebook\Proxies
+ */
+class Tracks {
+
+	/**
+	 * Record a tracks event.
+	 *
+	 * @param string $name       The event name to record.
+	 * @param array  $properties Array of properties to include with the event.
+	 */
+	public static function record_event( string $name, array $properties = [] ) {
+		if ( ! class_exists( WC_Tracks::class ) || ! WC_Site_Tracking::is_tracking_enabled() ) {
+			return;
+		}
+
+		$prefix = 'facebook_for_woocommerce_';
+
+		// In future we might want some standard default properties here.
+		// GLA has a more extensive API, not sure if we need that.
+
+		WC_Tracks::record_event( $prefix . $name, $properties );
+	}
+}
--- a/includes/fbbackground.php
+++ b/includes/fbbackground.php
@@ -8,7 +8,7 @@
  * @package FacebookCommerce
  */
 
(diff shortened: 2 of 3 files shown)


includes/Proxies/Tracks.php
```diff
@@ -0,0 +1,36 @@
...
+class Tracks {
+
+	/**
+	 * Record a tracks event.
+	 *
+	 * @param string $name       The event name to record.
+	 * @param array  $properties Array of properties to include with the event.
+	 */
+	public static function record_event( string $name, array $properties = [] ) {
+		if ( ! class_exists( WC_Tracks::class ) || ! WC_Site_Tracking::is_tracking_enabled() ) {
```
danielbitzer, line 25: Looks like we are doubling up by checking `WC_Site_Tracking::is_tracking_enabled()` here.

https://github.com/woocommerce/woocommerce/blob/dc6b44fc6f733e0526c5e16786ac9715aa019b74/includes/tracks/class-wc-tracks.php#L80-L82

includes/Proxies/Tracks.php
```diff
@@ -0,0 +1,36 @@
...
+			return;
+		}
+
+		$prefix = 'facebook_for_woocommerce_';
```
danielbitzer, line 29: Could use `\WC_Facebookcommerce::PLUGIN_ID` here but not a big deal.

includes/fbbackground.php
```diff
@@ -148,6 +150,8 @@ protected function task( $item ) {
 		 * performed, or, call parent::complete().
 		 */
 		protected function complete() {
+			WC_Tracks::record_event( 'product_sync_background_sync_completed' );
```
danielbitzer, line 153: Whoa what is this class? Is this another product sync background process? I thought correct background process was `\SkyVerge\WooCommerce\Facebook\Products\Sync\Background`. @budzanowski?

danielbitzer (changes requested):
I really like how you set this up following the proxy pattern! I think it would be great to make the `record_event` method non-static. That way we can mock calls to it in any future unit tests.

danielbitzer:
Sorry I realise review wasn't asked for but I had some thoughts when I saw this pop up.

haszari:
> Sorry I realise review wasn't asked for but I had some thoughts when I saw this pop up.

Reviews are always welcome!

includes/fbbackground.php
haszari: Uh oh!? I wondered about this too .. in any case, this event is just an example, if we move forward with tracking can remove or rename this event.

includes/Proxies/Tracks.php
haszari: Good point - I was inspired by the implementation in another plugin (pasted the proxy class over and renamed) 😁 

> I really like how you set this up following the proxy pattern! I think it would be great to make the `record_event` method non-static. That way we can mock calls to it in any future unit tests.

I don't see a huge benefit to mock this class in tests, but I think it's a good idea to do avoid `static` just for consistency. The main concern I have is that the utility is easy to use in client code, and doesn't affect the interface of client code. Should be as close to a one-liner drop in when you want to add an event somewhere in PHP.

danielbitzer: > I don't see a huge benefit to mock this class in tests

Mocking the class would be a neat way to test that a specific event was recorded without actually recording the event. Or just to be certain about avoiding executing tracking related code on a test site.

> Should be as close to a one-liner drop in when you want to add an event somewhere in PHP.

Definitely. Since we don't have a DI container I'd suggest something like: `facebook_for_woocommerce()->tracks_proxy()->record_event()`

haszari:
Closing this for now. We're [using tracker to get usage info on feed-based sync](https://github.com/woocommerce/facebook-for-woocommerce/pull/1972). Tracks events may be useful in future but aren't essential right now.

PR #1953, opened 2021-05-13
Add cron heartbeat to offload some tasks from `init` and `admin_init`

### LINKED ISSUE #1796: large actions table results in non-performant query
## 🔎 Isolate the bug

- [x ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.

## ✍️ Describe the bug
Reported in 3801136-zen that the query that runs on the init hook to `schedule_feed_generation` runs on every page load and ads 3-4 seconds of load time.

The site does have several subscriptions - the system status report shows a total of ~15k and the scheduled actions table seems to reflect that as well.

##🚶‍♀️ Steps to reproduce

## ✔️ Expected behavior
That the query needn't run on every page load

### LINKED ISSUE #1952: Avoid use of `init` or `admin_init` hooks for regular tasks, prefer `Heartbeat` system
Related issues: #1632 #1796

This plugin performs some actions, like scheduling tasks, on every request which unnecessarily degrades the whole site's performance. This is especially impactful on larger sites. 

## Examples:
- [x] [Scheduling product feed generation](https://github.com/woocommerce/facebook-for-woocommerce/blob/ea04112de070c4227802ce0a6c185534d9a05218/includes/Products/Feed.php#L57)
- [ ] [Schedules a recurring ACTION_FETCH_ORDERS action (I'm not sure what this is used for)](https://github.com/fbtransfer/facebook-for-woocommerce/blob/9b1a1dcb696fe4e1926b9f196b0375705fad4c41/includes/Commerce/Orders.php#L549)
- [x] [Refresh the local business configuration data with the latest from Facebook](https://github.com/woocommerce/facebook-for-woocommerce/blob/9b1a1dcb696fe4e1926b9f196b0375705fad4c41/includes/Handlers/Connection.php#L108) - This is cached to run every hour.

### **Update** - `Heartbeat` system implemented in #1953
The below "heartbeat" approach is now implemented. There are more places it could be used (see checklist above, or search for `init` or `admin_init`. Keeping this issue open to improve these over time. 

### WP CRON - Status checker

My suggestion is to implement a "heartbeat style" WP Cron job i.e. a cron job that runs every hour. This cron job would only be responsible for scheduling an action with Action Scheduler in order to minimize our impact on WP Cron requests.

Then we would move anything from `init` to this scheduled action's hook.

**Why WP Cron?** WP Cron events are storead in an autoloaded option so the performance impact of reading is extremely low. We should be able to confidently run the following code on every request.

```php
// Schedules the heartbeat cron scheduler
add_action( 'init', function() {
	if ( ! wp_next_scheduled( 'facebook_for_woocommerce_hourly_heartbeat_cron' ) ) {
		wp_schedule_event( time(), 'hourly', 'facebook_for_woocommerce_hourly_heartbeat_cron' );
	}
} );

// Schedules the heartbeat Action Scheduler action
add_action( 'facebook_for_woocommerce_hourly_heartbeat_cron', function() {
	as_schedule_single_action( time() + 60, 'facebook_for_woocommerce_hourly_heartbeat' );
} );
```

@woocommerce/tetris please let me know your thoughts on this approach and suggest any alternatives, thanks!

danielbitzer:

Implement an hourly cron heartbeat where setup tasks can be performed e.g. scheduling important actions. See #1952 for background and proposal.

Closes #1952
Closes #1796

1. Check out `master` and using Query Monitor check that the **regenerate feed action** is checked on every admin page load:

Screen Shot 2021-05-13 at 11 22 30 am

2. Check out this branch and that will no longer happen

Screen Shot 2021-05-13 at 11 22 58 am

3. Delete the `wc_facebook_generate_feed` action via the Action Scheduler UI
4. Install [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/) and reschedule the `facebook_for_woocommerce_hourly_heartbeat_cron` task to run now.
5. Head back to the Action Scheduler UI and check the `wc_facebook_generate_feed` was re-added.

### Changelog entry

> Performance - Move scheduling of certain actions to run every hour rather than on every request

### DIFF
--- a/includes/Heartbeat.php
+++ b/includes/Heartbeat.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace SkyVerge\WooCommerce\Facebook;
+
+use WC_Queue_Interface;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Heartbeat
+ *
+ * Responsible for scheduling cron heartbeat hooks. Currently there is a single hourly heartbeat:
+ * - `facebook_for_woocommerce_heartbeat_hourly`
+ *
+ * @since x.x.x
+ */
+class Heartbeat {
+
+	/**
+	 * Hook name for hourly heartbeat.
+	 */
+	const HOURLY = 'facebook_for_woocommerce_hourly_heartbeat';
+
+	/**
+	 * @var string
+	 */
+	protected $hourly_cron_name = 'facebook_for_woocommerce_hourly_heartbeat_cron';
+
+	/**
+	 * @var WC_Queue_Interface
+	 */
+	protected $queue;
+
+	/**
+	 * Heartbeat constructor.
+	 *
+	 * @param WC_Queue_Interface $queue WC Action Scheduler proxy.
+	 */
+	public function __construct( WC_Queue_Interface $queue ) {
+		$this->queue = $queue;
+	}
+
+	/**
+	 * Add hooks.
+	 */
+	public function init() {
+		add_action( 'init', array( $this, 'schedule_cron_events' ) );
+		add_action( $this->hourly_cron_name, array( $this, 'schedule_hourly_action' ) );
+	}
+
+	/**
+	 * Schedule heartbeat cron events.
+	 *
+	 * WP Cron events are stored in an auto-loaded option so the performance impact is much lower than checking and
+	 * scheduling an Action Scheduler action.
+	 */
+	public function schedule_cron_events() {
+		if ( ! wp_next_scheduled( $this->hourly_cron_name ) ) {
+			wp_schedule_event( time(), 'hourly', $this->hourly_cron_name );
+		}
+	}
+
+	/**
+	 * Schedule the hourly heartbeat action to run immediately.
+	 *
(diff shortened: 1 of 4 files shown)


includes/Heartbeat.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace SkyVerge\WooCommerce\Facebook;
```
haszari, line 3: Should this be in `Utilities` or some other folder? Probably a good idea to avoid adding stuff to `includes` root.

rcstr (approved):
The code looks great. Tested new `heartbeat` and action was scheduled hourly as expected 💯

includes/Heartbeat.php
danielbitzer: Sounds good, moved in 65fd9e2a6c74743c78478c1149f1cc80ddc0e5d4

haszari:
FYI I've updated the title and changelog entry to clarify that this uses the new `Heartbeat::HOURLY` to schedule the feed generation task (specifically).

PR #1962, opened 2021-05-17
Clean up background sync options

### LINKED ISSUE #1943: Syncing leaves useless `wc_facebook_background_product_sync_job` rows in options table.

### LINKED ISSUE #1948: Delete past completed jobs from the options table
Related to #1943 - "Syncing leaves useless `wc_facebook_background_product_sync_job` rows in options table".

The PR #1940 will cause any future jobs to be deleted on completion but there will still be many job records stored in the options table. We should delete these.

Unfortuantely we can't just run:
```php
$wpdb->select(
	"DELETE FROM {$wpdb->options}
	WHERE option_name LIKE 'wc_facebook_background_product_sync_job_%'
	OR option_name LIKE 'wc_facebook_background_handle_virtual_products_variations_job_%'
	OR option_name LIKE 'wc_facebook_background_remove_duplicate_visibility_meta_job_%'"
);
```
As this would also delete jobs in progress so we ideally need to reference the status of the job which is stored in the option row.

haszari:
Closes #1943
Closes #1948

~~**_Note_ this PR is based off #1953 `feature/add-cron-heartbeat` branch** because it uses the new hourly heartbeat event to schedule the cleanup task. I plan on merging this to master after #1953 is merged (`blocked` label).~~

This PR adds a cleanup task to remove `wp_options` rows for old, completed background sync jobs. 

As noted in the issue, the [SkyVerge plugin framework](https://github.com/skyverge/wc-plugin-framework) has a system for managing background jobs - see `SV_WP_Background_Job_Handler `. Each job is represented by a `wp_options` row, and they are not cleaned up.

Facebook for WooCommerce is using this to run a [background sync](https://github.com/woocommerce/facebook-for-woocommerce/blob/master/includes/Products/Sync/Background.php#L31) regularly. This is a periodic task, so is causing more and more options to accumulate over time (which could cause performance issues).

It's possible that this system is intended for one-off tasks. The other tasks that use it (`Background_Remove_Duplicate_Visibility_Meta`, `Background_Handle_Virtual_Products_Variations`) look like one-off jobs to me (based on comments), and I don't see any options rows for them in my dev site's database.

This PR only cleans up product sync jobs - `wc_facebook_background_product_sync_job_`. If other jobs are accumulating options or causing issues for merchants, we can clean those up too (in a separate PR/follow up).

1. Investigate to confirm your test site's database has redundant option rows - e.g. `SELECT * FROM `wp_options` WHERE option_name like 'wc_facebook_background%' `.
2. Switch to this branch. Wait 1+ days (e.g. leave overnight).
3. Confirm that the number of rows reduces over time (job cleans up incrementally, a few at a time). Check that the oldest ones are deleted first.
4. ~~View Facebook log in `WooCommerce > Status > Logs` and confirm there's a clear message logging how many rows were deleted.~~ (logging removed in 2cac5f656761bfd8bc76eb2261d640f951c0f3d7)

Also please confirm that this doesn't break or impact product sync, or delete options that it shouldn't.

### Changelog entry

> Fix - Clean up old product sync background job `wp_options` rows #1962


includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,69 @@
...
+	public function __construct() {
+		$this->add_hooks();
+	}
+
+	/**
+	 * Add hooks.
+	 */
+	public function add_hooks() {
+		// Register our cleanup routine to run on the hourly heartbeat.
+		add_action( Heartbeat::HOURLY, array( $this, 'clean_up_old_completed_options' ) );
```
haszari, line 35: Could add a `DAILY` heartbeat for things like this.

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,69 @@
...
+			"DELETE
+			FROM {$wpdb->options}
+			WHERE option_name LIKE 'wc_facebook_background_product_sync_job_%'
+			AND option_value LIKE '%%\"status\":\"completed\"%'
+			ORDER BY option_id ASC
+			LIMIT 10"
+		);
+
+		if ( $deleted_rows ) {
+			facebook_for_woocommerce()->log( "Deleted $deleted_rows old completed jobs from wp_options table." );
```
haszari, line 65: The log might not be necessary - thoughts?

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,69 @@
...
+		// - Order by lowest id, to delete older rows first.
+		// - Limit number of rows (periodic task will eventually remove all).
+		// Using `get_results` so we can limit number of items; `delete` doesn't allow this.
+		$deleted_rows = $wpdb->query(
+			"DELETE
+			FROM {$wpdb->options}
+			WHERE option_name LIKE 'wc_facebook_background_product_sync_job_%'
+			AND option_value LIKE '%%\"status\":\"completed\"%'
+			ORDER BY option_id ASC
+			LIMIT 10"
```
haszari, line 61: Setting a limit of 10 for testing. I think we should bump this to delete 100 at a time before merge. Keen for feedback on this limit, and performance of the query.

haszari:
@rcstr Can you review this and #1953 please? Thanks :)

includes/Utilities/BackgroundJobCleanup.php
budzanowski: Yep. Not needed.

includes/Utilities/BackgroundJobCleanup.php
budzanowski: 10 is OK. If you would not delete all in one go you would be slowly depleting with each heartbeat. No need to hurry, the site is not creating 10 new entries every hour.

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * Logic and database query are adapted from SV_WP_Background_Job_Handler::get_jobs().
+	 *
+	 * @see SV_WP_Background_Job_Handler
+	 * @see Products\Sync\Background
+	 */
+	public function clean_up_old_completed_options() {
+		global $wpdb;
+
+		// Query notes:
```
budzanowski, line 49: Please use multi line comment `/* */`

budzanowski (approved):

includes/Utilities/BackgroundJobCleanup.php
haszari: Any thoughts on bumping it up to 50 or 100? Or adding a filter?

I figure that for the longer-lived stores, they'll want these to be cleared up faster, and there's probably little performance difference between 10 and 100.

includes/Utilities/BackgroundJobCleanup.php
haszari: 49bc9eb081c30d316e9a6b70d1147f04870d1c9b

includes/Utilities/BackgroundJobCleanup.php
haszari: Removed it in 2cac5f656761bfd8bc76eb2261d640f951c0f3d7

includes/Utilities/BackgroundJobCleanup.php
budzanowski: This query looks relatively safe ( I have not to measure it ), bumping to 50/100 sounds OK.

rcstr: We can increase this number safely (up to 1k). I ran a few tests:

1. Generated new entries using ( ran a few times until I had ~10k records )
```php
add_action('admin_init', function() {
	global $wpdb;
	$values = '';
	$limit = 550;
	$time = time();

	for ($i = 1;$i<=$limit;$i++) {
		// Queued
		$values .= " ('wc_facebook_background_product_sync_{$time}1{$i}', '{\"requests\":{\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4209$i\":\"UPDATE\",\"p-4213$i\":\"UPDATE\"},\"id\":\"{$time}1{$i}\",\"created_at\":\"2021-05-04 17:17:05\",\"created_by\":1,\"status\":\"queued\"}', 'no'), \n";

		// Completed.
		$values .= " ('wc_facebook_background_product_sync_{$time}2{$i}', '{\"requests\":{\"p-42092$i\":\"UPDATE\",\"p-42092$i\":\"UPDATE\",\"p-420942$i\":\"UPDATE\",\"p-42092$i\":\"UPDATE\",\"p-42092$i\":\"UPDATE\",\"p-42092$i\":\"UPDATE\",\"p-42092$i\":\"UPDATE\",\"p-42132$i\":\"UPDATE\"},\"id\":\"{$time}2{$i}\",\"created_at\":\"2021-05-04 17:17:05\",\"created_by\":1,\"status\":\"completed\"}', 'no')";
		if ( $i < $limit ):$values .= ", \n";
		else:$values .= ";";endif;
	}

	$wpdb->query("
		INSERT INTO `wp_options` (`option_name`, `option_value`, `autoload`) VALUES $values");
});
```
2. Ran (**9914 total, Query took 0.0007 seconds.**)
```sql
SELECT * FROM wp_options WHERE option_name like 'wc_facebook_background%'
```
3. Deleted 10 records (**Query took 0.0699 seconds.**)
```sql
delete FROM wp_options WHERE option_name LIKE 'wc_facebook_background%' AND option_value LIKE '%\"status\":\"completed\"%' ORDER BY option_id ASC LIMIT 10

```
4. Tested with 100 (**Query took 0.0675 seconds**)
4. Increased limit to 500 (**Query took 0.0761 seconds.**)

The difference is very little, even at 500.

includes/Utilities/BackgroundJobCleanup.php
rcstr: If we increase the `limit` in the `delete` query I'd suggest we increase this to _every 12 hours_ or even _daily_.

rcstr (approved):
The feature works as expected, `completed` jobs were deleted from the database as expected. I left a few suggestion, but not blockers

includes/Utilities/BackgroundJobCleanup.php
haszari: Will need to extend `Heartbeat` to support that, but I agree it's worth doing.

includes/Utilities/BackgroundJobCleanup.php
haszari: Thanks for perf testing @rcstr ! I'm thinking we bump it up fairly conservatively, since most stores won't need a huge number. Stores that do have thousands of these should probably manually clean them up. Maybe we bump to 100 or 250 to keep it fairly aggressive (speed of cleanup) but still super safe.

includes/Utilities/BackgroundJobCleanup.php
haszari: Keen for @danielbitzer feedback on this :)

haszari: Added a `daily` heartbeat and switched this to using it. Now .. we need a day+ to test!! @danielbitzer can you review this? If review all good, I'm thinking we merge this (since it's low risk) and then we can test it more on master to ensure it's cleaning up.

includes/Utilities/BackgroundJobCleanup.php
haszari: Now bumped to 250.

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace SkyVerge\WooCommerce\Facebook\Utilities;
+
+use SkyVerge\WooCommerce\Facebook\Heartbeat;
```
danielbitzer, line 5:
```suggestion
use SkyVerge\WooCommerce\Facebook\Utilities\Heartbeat;
```

Heartbeat class has moved as you suggested.

class-wc-facebookcommerce.php
```diff
@@ -156,6 +160,7 @@ public function init() {
 				$this->product_sets_sync_handler = new \SkyVerge\WooCommerce\Facebook\ProductSets\Sync();
 				$this->commerce_handler          = new \SkyVerge\WooCommerce\Facebook\Commerce();
 				$this->fb_categories             = new \SkyVerge\WooCommerce\Facebook\Products\FBCategories();
+				$this->background_job_cleanup    = new BackgroundJobCleanup();
```
danielbitzer, line 163: I'd move this into the `JobsRegistry` class.

includes/Utilities/BackgroundJobCleanup.php
danielbitzer: I like the idea of adding doing this daily 💯

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace SkyVerge\WooCommerce\Facebook\Utilities;
```
danielbitzer, line 3: Why not move this into `namespace SkyVerge\WooCommerce\Facebook\Jobs`? We don't want the Utilities directory to become a dumping ground.

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @since x.x.x
+ */
+class BackgroundJobCleanup {
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->add_hooks();
```
danielbitzer, line 27: Personally I like to avoid coupling "real work" to class instantiation but not a blocker. A class as simple as this it's probably fine.

includes/Utilities/BackgroundJobCleanup.php
```diff
@@ -0,0 +1,67 @@
...
+ * Responsible for cleaning up old completed background sync jobs from SkyVerge background job system.
+ * Each job is represented by a row in wp_options table, and these can accumulate over time.
+ *
+ * Note - this is closely coupled to the SkyVerge background job system, and is essentially a patch to improve it.
+ *
+ * @see SV_WP_Background_Job_Handler
+ *
+ * @since x.x.x
+ */
+class BackgroundJobCleanup {
```
danielbitzer, line 21: This class name seems a bit ambiguous to me. What about something like `Facebook\Jobs\CleanupSkyvergeFrameworkJobOptions`? Since we have 3 different job systems some specificity can't hurt IMO.

danielbitzer (changes requested):
Looks pretty good! 

- love the daily heartbeat addition. We could honestly use that for the feed check too. Doing it hourly is perhaps overkill.
- I like the simplicity of the delete solution 
- I left some inline structural comments

includes/Utilities/BackgroundJobCleanup.php
haszari: Doh! Thanks for catching that.

class-wc-facebookcommerce.php
haszari: Yep that makes sense, I'm starting to see that as `JobManager` or `JobController`. I can move it there to get behind that.

includes/Utilities/BackgroundJobCleanup.php
haszari: Originally it was because this was potentially targeting jobs from other places. It is a bit of a script/utility, I'll look at moving it.

includes/Utilities/BackgroundJobCleanup.php
haszari: If it's better to add this to an `init` hook, or something - let me know, happy to do that.

haszari: .. I'm following the pattern used in `JobRegistry`. AKA `JobManager` (I think we should rename it, since it's not a generic registry).

haszari:
@danielbitzer I've pushed changes - all good suggestions. Just giving it another test with `hourly` now to confirm it's working.

class-wc-facebookcommerce.php
haszari: Checked to see if `registry` is an "official" design pattern - https://wiki.c2.com/?RegistryPattern

includes/Utilities/BackgroundJobCleanup.php
danielbitzer: Perhaps, what are you basing your definition of a registry class on? I'm thinking it "registers" job classes and then "registers/adds" the job hooks for each class.

class-wc-facebookcommerce.php
haszari: This explains it better - https://www.martinfowler.com/eaaCatalog/registry.html

includes/Utilities/BackgroundJobCleanup.php
danielbitzer: I have a slight dislike for "manager" named classes because I've seen them end up as catch-all for all related functionality and then become a mess.

haszari: A registry as I understand it is a generic container for well-known classes of things - so you might register logging classes, or middlewares for processing things, or image processing filters. Specifically (in our case), it doesn't know at compile time all the things that are in it. Any component that needs an X can get one from the registry. Any component that implements/provides an X registers it with the registry so others can get it.

https://www.martinfowler.com/eaaCatalog/registry.html

> I have a slight dislike for "manager" named classes because I've seen them end up as catch-all for all related functionality and then become a mess.

`Manager` and `Controller` both have that risk for sure. But I've now convinced myself that we shouldn't call this `JobRegistry` - let me know what you think. Unrelated to this PR so can follow up later :)

haszari:
> Just giving it another test with hourly now to confirm it's working.

Testing complete - my last few options are now gone. Reinstated daily, ready to merge :)

danielbitzer (approved):
🚢

includes/Utilities/BackgroundJobCleanup.php
haszari: Added a follow up issue for this! #1968

PR #1971, opened 2021-05-20
Check variation parent status before sync.

### LINKED ISSUE #1803: New product copies are immediately sent to Facebook
## 🔎 Isolate the bug

## ✍️ Describe the bug
When a synced product is copied to a new draft, a copy is immediately synced to Facebook. Quite often, this workflow is used to create completely new and different products. It doesn't mean that duplicate products are also needed on the Facebook side though.

##🚶‍♀️ Steps to reproduce

1. Copy a synced product to a new draft.
2. Change all the details about the product including SKUs.
3. Don't publish the product.
4. There will be a duplicate on Facebook side.

## ✔️ Expected behavior
Ideally, products should only be synced to Facebook after they're published on WooCommerce side so that users get a chance to disable Facebook sync before publishing.

budzanowski:

Duplicated variable products are duplicated with `publish` status. Only the parent has `draft` status. This caused the sync logic to trigger sync for duplicated variations.

This PR adds a parent status check when we determine if a variation is eligible for synchronization.

Closes #1803 .

Check reproduction instructions in #1803

### Changelog entry

Fix: New variation copies are immediately sent to Facebook Catalog.

### DIFF
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -230,6 +230,11 @@ class Products {
 			$terms_product = null;
 		}
 
+		// Check parent status.
+		if ( null !== $terms_product && 'publish' !== $terms_product->get_status() ) {
+			return false;
+		}
+
 		// allow simple or variable products (and their variations) with zero or empty price - exclude other product types with zero or empty price
 		if ( $should_sync && ( ! $terms_product || ( ! self::get_product_price( $product ) && ! in_array( $terms_product->get_type(), array( 'simple', 'variable' ) ) ) ) ) {
 			$should_sync = false;


danielbitzer (commented):
I'm don't think we can make this change. At least not without a lot of testing. I've seen this method (`published_product_should_be_synced`) used in certain places where this change could cause issues e.g. in `\WC_Facebookcommerce_Integration::fb_change_product_published_status` where the which this is hooked into `transition_post_status`. I think the intention of the code is to have a function that deliberately ignores the "published" status check.

I feel like there could be a less risky way to fix #1971

rcstr:
> I think the intention of the code is to have a function that deliberately ignores the "published" status check.

I have the same perspective of this function, it already expects `published` products. 

--- 

`WC_Facebookcommerce_Integration::product_should_be_synced` or `SkyVerge\WooCommerce\Facebook\Products::product_should_be_synced` seem to be a better place for this check

https://github.com/woocommerce/facebook-for-woocommerce/blob/e52d30a155d4440757d8de171ac9f467d5644edc/includes/Products.php#L203-L206

https://github.com/woocommerce/facebook-for-woocommerce/blob/604d7551701afa281a1acb8fbba8dce0ea9c7f4d/facebook-commerce.php#L1350-L1365

PR #1972, opened 2021-05-21
Add tracker props to understand usage of feed-based sync and other FB business config options

### LINKED ISSUE #1944: Add tracking events to key features to better understand product usage

haszari:
Fixes #1944 (not exact match - we can open follow up issues if we need more tracking)

This PR adds a new tracker properties for various settings configured in Facebook UI.

- `feed-file-requested`: true if the feed file has been requested by Facebook (since last snapshot).
- `feed-schedule-enabled` (`catalog_feed_scheduled.enabled`): may indicate that the store has configured a schedule for feed-based sync.
- `ig-shopping-enabled` (`ig_shopping.enabled`): may indicate that the store has configured Instagram Shopping (or Instagram Checkout)
- `ig-cta-enabled` (`ig_cta.enabled`): unsure what this relates to, likely related to Instagram checkout/shopping 
- `product-feed-config`: various feed config options, and metadata about any recent feed upload

How it works:

- When feed is requested, we add a transient.
- When the tracker snapshot is generated, we add a new prop (based on the transient value, if it exists). Then delete the transient so the next snapshot uses the latest info.
- When we request config (`fbe_business`) from Facebook, we cache various config settings in a transient, and later include these in tracker snapshot.
- Once a day (`Heartbeat::DAILY`) we call various Facebook Graph API endpoints to get feed config settings and info about the most recent upload (if available), and cache in transient for inclusion in snapshot.

Note: transient values can be viewed in `wp_options` table, or via WP-CLI.

I recommend testing on Jurassic Ninja or similar, so the site is accessible and you can leave it running to test the config reflects the latest info.

##### Test: Configure manual upload in Facebook, store upload request as transient
1. Set up plugin, connect to Facebook.
2. Wait for feed to be generated, or force it via Action Scheduler.
3. In Facebook Business, go to `Facebook > Commerce > Catalogue > Data sources` and select / create a data source (feed config).
4. Click `Settings` then `Upload now` and specify your feed URL. It should be something like this:
   - `https://your-test-site.com/?wc-api=wc_facebook_get_feed_data&secret=your-feed-secret`
   - You can get the secret from `wp_options` table or `wp option get` - option name `wc_facebook_feed_url_secret`
   - Your site will need to be accessible to Facebook - e.g. [use `ngrok`](https://ngrok.com) or similar if you're developing on local machine
5. Proceed with feed upload. Facebook should request your feed (check logs for `Facebook is requesting the product feed.` - if logs are enabled!).
6. After Facebook has requested the feed, the transient `wc_facebook_wctracker_feed_requested` should exist with value `1` (true). 

##### Test: Configure upload schedule(s) in Facebook, store upload request as transient
1. In Facebook Business, go to `Facebook > Commerce > Catalogue > Data sources` and select a data source (feed config).
4. Click `Settings` then set up a schedule in `Schedules`. Feel free to set up `update` or `update_schedule` or both.
7. Wait a day, or manually trigger `Heartbeat::DAILY` (please comment with steps - I don't know how to do this!). There should be a transient (`facebook_for_woocommerce_wctracker_fb_feed_config`) with various settings.
    - If you want a shortcut - uncomment `add_action( 'admin_init', array( $this, 'track_data_source_feed_tracker_info' ) );` 😄 

##### Test: Refresh Facebook config and store settings in transient
1. Delete transient `wc_facebook_business_configuration_refresh`. This forces a refresh of Facebook settings. 
2. Confirm `wc_facebook_wctracker_fbe_business_config` transient exists.

##### Test: Confirm tracker includes new props

I've [opened a PR in WooCommerce Core to add a CLI command to show the tracker snapshot](https://github.com/woocommerce/woocommerce/pull/30010). So an easy way to check is to check out that branch of WooCommerce and run `wp wc tracker snapshot --format=yaml`. Then check each field to make sure it's correct and useful to us.

**This is the key thing to test** - the above steps all ultimately add data to the tracker snapshot. Please experiment with different scenarios that you are interested in, and confirm the tracker snapshot is accurate.

Here's the relevant chunk of tracker info on my store. 

```php
    (
        [is-connected] => yes
        [product-sync-enabled] => yes
        [messenger-enabled] => no
        [feed-file-requested] => yes
        [ig-shopping-enabled] => no
        [ig-cta-enabled] => no
        [product-feed-config] => Array
            (
                [site-feed-id] => 
                [feed-count] => 1
                [active-feed] => Array
                    (
                        [created-time] => 2021-05-27 23:47:15
                        [product-count] => 0
                        [schedule] => Array
                            (
                                [interval] => HOURLY
                                [interval-count] => 1
                            )

                        [latest-upload] => Array
                            (
                                [end-time] => 2021-06-01 04:10:59
                                [error-count] => 1
                                [warning-count] => 0
                                [num-persisted-items] => 0
                                [url-matches-site-endpoint] => 1
                            )
                    )
            )
    )
```

Once we've confirmed this is working, and the data is correct, I'm keen to get a review from data team to make sure we're using the right types/encoding for dates/times etc.

## Checklist for this PR

- [x] Test and confirm the tracker props are accurate.
- [ ] Get a review and feedback from data team (data types, dates, will data answer our questions). 
- [ ] Team review and confirm props are sufficient for our needs (shipping improved feed engine).
- [ ] **_remove `blocked` label_**
- [ ] Follow up: is #1945 still useful, or should we close?
- [ ] Merge.
- [ ] Ship in 2.6.
- [ ] Document changes on https://woocommerce.com/usage-tracking/

### Changelog entry

> New - Add WooCommerce Tracker props for feed and Instagram related settings.


haszari:
This is ready for review now. Leaving `blocked` and `documentation` label as a reminder to document any changes to tracking, and to confirm these will provide useful data to us. 

I haven't managed to test `feed-schedule-enabled`, `ig-shopping-enabled`, or `ig-cta-enabled` yet:

- I believe Instagram is geo-gated (I'm not in USA).
- I have configured a feed schedule but `catalog_feed_scheduled.enabled` is still false. Maybe this is cached on the Facebook end.

includes/API/FBE/Configuration/Read/Response.php
```diff
@@ -40,5 +40,58 @@ public function get_messenger_configuration() {
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @return null|API\FBE\Configuration\CatalogFeedSchedule
+	 */
+	public function get_catalog_feed_configuration() {
+
+		$configuration = null;
+
+		if ( ! empty( $this->response_data->catalog_feed_scheduled ) && is_object( $this->response_data->catalog_feed_scheduled ) ) {
```
budzanowski, line 54: In my testing, the `catalog_feed_schedule` is `false` despite the fact that I have a valid and active feed configuration.

rcstr:
> I don't think that we can use `catalog_feed_schedule`, it has no actual information.

This should change once we merge #1928.

haszari:
> > I don't think that we can use catalog_feed_schedule, it has no actual information.

> This should change once we merge #1928.

@rcstr I'm not confident `catalog_feed_schedule` will give us any useful information. Especially if we want to track use of the feed before we ship any changes.

@budzanowski prototype PR #1977 has some other API calls related to feed that give us useful information to track - I'll update the PR to use data from those APIs. Example data that would help us understand feed usage:

- Feed nodes associated with store catalog (i.e. "data sources", feed config).
- Recent uploads / requests of feed content (date of last feed sync).

includes/API/FBE/Configuration/Read/Response.php
```diff
@@ -40,5 +40,58 @@ public function get_messenger_configuration() {
...
+	}
+
+	/**
+	 * Is Instagram Shopping enabled?
+	 *
+	 * @since x.x.x
+	 *
+	 * @return boolean
+	 */
+	public function get_ig_shopping_enabled() {
```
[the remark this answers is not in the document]
haszari: I agree, that's clearer. Fixed in d864cfe585c8a649995d604acb41a5be45dd514c

includes/Utilities/Tracker.php
```diff
@@ -21,6 +21,27 @@
...
+	 * @var string
+	 */
+	const TRANSIENT_WCTRACKER_FEED_REQUESTED = 'wc_facebook_wctracker_feed_requested';
+
+	/**
+	 * Transient key name; stores various FBE business settings.
+	 *
+	 * @var string
+	 */
+	const TRANSIENT_WCTRACKER_FBE_BUSINESS_CONFIG = 'wc_facebook_wctracker_fbe_business_config';
```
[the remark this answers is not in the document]
haszari: Consistency gets my vote! Thanks for spotting this 🍇  48b70f1768f310488ff577bf7fb7c1f840be6395

includes/API/FBE/Configuration/CatalogFeedSchedule.php
```diff
@@ -0,0 +1,58 @@
...
+namespace SkyVerge\WooCommerce\Facebook\API\FBE\Configuration;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * The catalog feed configuration settings.
+ *
+ * @since x.x.x
+ */
+class CatalogFeedSchedule {
```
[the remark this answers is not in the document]
haszari: Thanks for linking me to the API docs – I was curious what might show up in `catalog_feed_scheduled`. 

I'm going to look at adding tracking that will work for the current feed, based on @budzanowski PR #1977. If this API isn't giving us useful info, I think it's better to remove it (and the class representing that chunk of response). Will see how I go with the other APIs used in #1977.

haszari:
@budzanowski @rcstr  - FYI I'm still in progress on this, but it's looking much better, now that it's using the Graph API to get the feed config. Thanks for the tip Bartek - I've adapted (pasted) the code from #1977 on to this branch, decoupled from the new feed implementation. 

See below for an example of what we could add to tracker snapshot - I'm excited to get error rates, `num_persisted_products` etc to help us understand how the feed is being used, and to measure impact of forthcoming improvements! Let me know what you think, and what else might help illuminate feed usage.

```php
[product_feed_config] => Array
    (
        [feed_count] => 1
        [active_feed] => Array
            (
                [created_time] => 2021-05-07T02:26:26+0000
                [product_count] => 0
                [update_schedule] => Array
                    (
                        [interval] => DAILY
                        [interval_count] => 1
                    )
                [latest_upload] => Array
                    (
                        [end_time] => 2021-05-26T02:32:01+0000
                        [error_count] => 1
                        [warning_count] => 0
                        [num_persisted_items] => 0
                    )
            )
    )
```

includes/Feed/FeedConfigurationDetection.php
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * A class responsible detecting feed configuration.
+ */
+class FeedConfigurationDetection {
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_action( Heartbeat::DAILY, array( $this, 'get_data_source_feed_tracker_info' ) );
```
haszari, line 22: Note - this runs on daily heartbeat, so you will likely need to force the heartbeat to test.

haszari:
@rcstr @budzanowski  This is ready for another review. 

There's quite a lot happening now - the tracking data is recorded in a few different places. In particular the feed config is updated on `Heartbeat::DAILY`, so you'll need to manually trigger it somehow.

Please test a few scenarios to ensure the tracking data is a valid, useful representation of the config. My store / account seems broken at the moment, so I haven't yet confirmed the feed settings with a working feed. So if you have a store with a working feed, please test and confirm this is correct!

includes/Feed/FeedConfigurationDetection.php
```diff
@@ -0,0 +1,147 @@
...
+
+		$info['feed_count'] = count( $feed_nodes );
+
+		// Check if the catalog has any feed configured.
+		if ( empty( $feed_nodes ) ) {
+			facebook_for_woocommerce()->log( 'No feed nodes!' );
+			return false;
+		}
+
+		// Determine which is the most active feed (recently updated).
```
budzanowski, line 64: `most actvive` !== `recently updated`

If you have two feeds: one that was updated 1h ago and another one that was updated 23h ago it may also mean that the 23h ago will be updated +1h ( if this is a daily schedule ). Out of those two, you don't know which one would be the correct one.

The first thing that would need to happen is that you need to evaluate the feed indicated by the `$integration_feed_id` which now is not used. If that one is not set or has no corresponding config then we should proceed to evaluate the remaining items.

The items should only be evaluated if the 'url' in the `schedule` section matches the one used by our site. If the `url` does not match then this means that this feed config is not related to the site where the plugin is installed.

haszari:
Update – I've confirmed this with a test site with one feed file, and one product (nice easy example to start!!).

```
    [extensions] => Array
        (
                (
                    [is-connected] => yes
                    [product-sync-enabled] => yes
                    [messenger-enabled] => no
                    [feed-file-requested] => yes
                    [ig-shopping-enabled] => no
                    [ig-cta-enabled] => no
                    [product-feed-config] => Array
                        (
                            [feed_count] => 1
                            [active_feed] => Array
                                (
                                    [created_time] => 2021-05-27T23:47:15+0000
                                    [product_count] => 0
                                    [schedule] => Array
                                        (
                                            [interval] => HOURLY
                                            [interval_count] => 1
                                        )

                                    [latest_upload] => Array
                                        (
                                            [end_time] => 2021-06-01T02:09:14+0000
                                            [error_count] => 0
                                            [warning_count] => 6
                                            [num_persisted_items] => 1
                                            [url] => https://developing-ocelot.jurassic.ninja/?wc-api=wc_facebook_get_feed_data&secret=ece17f802b1a30d28b020ee265d48edd
                                        )

                                )

                        )

                )

        )
```

includes/Feed/FeedConfigurationDetection.php
haszari: Thanks for the tips! I've tweaked the selection of the "active" feed:

- If a feed config ID matches `$integration_feed_id`, it's selected as the "active" feed.

Note: the "active" feed in this case just means the one that we will use for tracker props. If the majority of sites have multiple feeds set up, we will not be able to draw many conclusions. This code assumes the normal case is a single feed config. Does that make sense? I just want to capture useful info to help us understand if the feed mechanism is being used on sites. In the case of multiple feeds configured, the most recent upload seems a good heuristic for selecting which config to track.

Also, for the most recent upload of the "active" feed, we now track `url-matches-site-endpoint`. If this is false, then something else is going on (e.g. site hostname has changed, secret has been regenerated, or the merchant has some complex config where their feed file is hosted elsewhere like a CDN or AWS bucket).

haszari:
Ready for another review and test - thanks all @rcstr @budzanowski. (And apologies for how complex this has become.)

The key thing to test here:

- Set up working or broken feed configs in Facebook.
- Trigger the tracking code and confirm the tracker props are accurate and useful to us (for shipping a new feed engine safely).

🙌

haszari:
@deltaWhiskey I've pinged you for a review as a woo tracker expert (feel free to re-assign and timebox). Specifically I'm interested in:

- Suggestions on how the data is encoded (dates, formats etc).
- Feedback on if these props will help answer our questions (e.g. "Is the feed being used to sync products" "Is it working").

I'll write a post with more detail and context soon.

budzanowski:
Data collected correctly.
```php
array (
  'is-connected' => 'yes',
  'product-sync-enabled' => 'yes',
  'messenger-enabled' => 'no',
  'feed-file-requested' => 'no',
  'ig-shopping-enabled' => 'no',
  'ig-cta-enabled' => 'no',
  'product-feed-config' => 
  array (
    'site-feed-id' => '176..............',
    'feed-count' => 1,
    'active-feed' => 
    array (
      'created-time' => '2021-04-26 09:16:18',
      'product-count' => 8,
      'schedule' => 
      array (
        'interval' => 'DAILY',
        'interval-count' => 1,
      ),
      'latest-upload' => 
      array (
        'end-time' => '2021-05-31 23:01:00',
        'error-count' => 1,
        'warning-count' => 0,
        'num-persisted-items' => 0,
        'url-matches-site-endpoint' => true,
      ),
    ),
  ),
)
```

includes/Feed/FeedConfigurationDetection.php
```diff
@@ -0,0 +1,170 @@
...
+			$active_feed['schedule']['interval-count'] = $active_feed_metadata['schedule']['interval_count'];
+		}
+		if ( array_key_exists( 'update_schedule', $active_feed_metadata ) ) {
+			$active_feed['update-schedule']['interval']       = $active_feed_metadata['update_schedule']['interval'];
+			$active_feed['update-schedule']['interval-count'] = $active_feed_metadata['update_schedule']['interval_count'];
+		}
+
+		$info['active-feed'] = $active_feed;
+
+		$latest_upload      = $active_feed_metadata['latest_upload'];
```
budzanowski, line 117: It looks like this may be problematic. I have created a new fresh feed config. It immediately started to fetch the data. I have killed the local server to simulate upload failure. I hoped for an empty 'latest_upload' array but still, I got some data:

image

Start and end times are the same indicating that nothing really has happened.

This means that even if this is broken on a merchant's site we will not catch that. This is true for this part and for the code at line 93 for detecting the lastest upload time.

budzanowski: I have dig through the documentation and I have found nothing that would directly indicate that a feed upload is valid or not. One approach is to request uploads instead of 'latest_upload' and parse data in there.
Still, we are not able to get clear direct information but it looks like there is a solid way to get what we want.
We need to request additional fields:
`xxx/?fields=uploads{start_time,end_time,num_detected_items,num_deleted_items,num_invalid_items,num_persisted_items}`

When the feed upload fails the fields:
`num_detected_items,num_deleted_items,num_invalid_items,num_persisted_items` all equal to 0. When the feed upload is a success then some of the fields are not equal to 0:
image

There is also an errors field but it would be very problematic to parse it for this information because it can have an for the whole feed or error field for each individual line item. It just has textual information without any sort of code.

The `latest_uploads` does not have these additional fields.  If we want to just check the `latest_upload` but with the additional data we can do that by limiting the `uploads` field:
`?fields=uploads.limit(1){start_time,end_time,num_detected_items,num_deleted_items,num_invalid_items,num_persisted_items,errors{id,suggested_rules}}`

We need to get that right since we expect that most of the sites do not have a usable feed.

haszari: > This means that even if this is broken on a merchant's site we will not catch that. This is true for this part and for the code at line 93 for detecting the lastest upload time.

> We need to request additional fields: `fields=uploads{start_time,end_time,num_detected_items,num_deleted_items,num_invalid_items,num_persisted_items}

Yes - this is what I'm thinking too! I'm using this kind of approach to populate `latest-upload`. 

https://github.com/woocommerce/facebook-for-woocommerce/blob/15fd70979f82e2e66b91b655413efc406c6bebc0/includes/Feed/FeedConfigurationDetection.php#L121-L125

I'm thinking that we'll use a combination of the following fields to get an idea of whether feed is working or not. For example, for your test data (feed not working):

```
active-feed.product-count = 8 
active-feed.latest-upload.num-persisted-items = 0
active-feed.latest-upload.error-count = 1
```

If the feed is working correctly, `error-count` will be zero, and `num-persisted-items` will be non-zero; ideally, equal to `product-count`.

Note: The goal of this PR is to track enough data to help us release a new feed engine with confidence. In future we may need to build on this to detect and migrate existing config - but that's a different goal. Do you think the data collected here will give enough insight for our launch plan? If there's something important missing, let's add it.

haszari:
# 🚢 🎉 🍾 

I've just seen this update on my test site, and it looks like a functioning feed!

Note the following values:

- `active-feed.product-count` = 1
- `active-feed.latest-upload.num-persisted-items` = 1
- `active-feed.latest-upload.error-count` = 0
- `active-feed.latest-upload.url-matches-site-endpoint` = 1

```php
    (
        [is-connected] => yes
        [product-sync-enabled] => yes
        [messenger-enabled] => no
        [feed-file-requested] => yes
        [ig-shopping-enabled] => no
        [ig-cta-enabled] => no
        [product-feed-config] => Array
            (
                [site-feed-id] => 
                [feed-count] => 1
                [active-feed] => Array
                    (
                        [created-time] => 2021-05-27 23:47:15
                        [product-count] => 0
                        [schedule] => Array
                            (
                                [interval] => HOURLY
                                [interval-count] => 1
                            )

                        [latest-upload] => Array
                            (
                                [end-time] => 2021-06-02 03:09:17
                                [error-count] => 0
                                [warning-count] => 6
                                [num-persisted-items] => 1
                                [url-matches-site-endpoint] => 1
                            )
                    )
            )
    )
```

Now I'm going to change the feed secret, which should break it for tomorrow's heartbeat :)

includes/Utilities/Tracker.php
```diff
@@ -45,22 +73,95 @@ public function add_tracker_data( array $data = array() ) {
...
 		$data['extensions']['facebook-for-woocommerce']['product-sync-enabled'] = wc_bool_to_string( $product_sync_enabled );
 		$messenger_enabled = facebook_for_woocommerce()->get_integration()->is_messenger_enabled();
 		$data['extensions']['facebook-for-woocommerce']['messenger-enabled'] = wc_bool_to_string( $messenger_enabled );
 
+		/**
+		 * Has the feed file been requested recently?
+		 *
+		 * @since x.x.x
+		 */
+		$feed_file_requested = get_transient( self::TRANSIENT_WCTRACKER_FEED_REQUESTED );
```
budzanowski, line 103: You need to protect the code from expired transients:

```bash
[01-Jun-2021 10:54:28 UTC] PHP Notice:  Trying to get property 'ig_cta_enabled' of non-object in /Users/bartoszbudzanowski/Local/facebook/app/public/wp-content/plugins/facebook-for-woocommerce/includes/Utilities/Tracker.php on line 114
```

budzanowski:
<p>One interesting piece of data that we should track is <code>wc_facebook_average_feed_generation_time</code> - it will tell us how long the site takes to create a feed file. I would go one step further. Since it looks like the average feed generation metric is not used - at least I was not able to found any usage - I would do as follows:</p>



<ul><li>remove the average feed generation time calculation code.</li><li>directly store the feed generation time in an option.</li><li>create an update script that during installation would wipe the old value.</li></ul>



<p>This gives us insight if the feed generation is not failing. The option would be populated only if the feed file generation is able to finish.</p>



<p>If we don't want to make those changes ( I believe that they are painless to the workings of the plugin ) we should at least track the old value of <meta charset="utf-8"><code>wc_facebook_average_feed_generation_time</code>. We risk storring old old value this way though.</p>

rcstr (approved):
Looking great! I was able to test successfully. As @budzanowski suggested, adding `wc_facebook_average_feed_generation_time` will be very insightful 

```bash
[www-data@7fc45b2d4b5d html]$ wp transient get facebook_for_woocommerce_wctracker_fb_feed_config
array (
  'site-feed-id' => '538350824218720',
  'feed-count' => 1,
  'active-feed' => 
  array (
    'created-time' => '2021-06-02 16:53:20',
    'product-count' => 694,
    'schedule' => 
    array (
      'interval' => 'DAILY',
      'interval-count' => 1,
    ),
    'latest-upload' => 
    array (
      'end-time' => '1970-01-01 00:00:00',
      'error-count' => 0,
      'warning-count' => 0,
      'num-persisted-items' => 0,
      'url-matches-site-endpoint' => 'yes',
    ),
  ),
)

```

includes/Feed/FeedConfigurationDetection.php
```diff
@@ -0,0 +1,172 @@
...
+
+namespace SkyVerge\WooCommerce\Facebook\Feed;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+use Error;
+use SkyVerge\WooCommerce\Facebook\Utilities\Heartbeat;
+use SkyVerge\WooCommerce\Facebook\Feed\FeedFileHandler;
```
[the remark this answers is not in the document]
haszari: The dangers of stealing someone else's (@budzanowski) code from a future feature branch! 😆

includes/Utilities/Tracker.php
haszari: I think I've fixed this here: d9d57beac9b814c0a2c33bfe1ac839cd296fb819

I couldn't reproduce the notice though - not sure why. I have `WP_DEBUG` set to true, and I deleted the transient, then actually forced `$config = false`. Let me know if there's some other debugging settings I need to enable.

haszari:
Thanks for the thoughtful reviews @rcstr @budzanowski !

> One interesting piece of data that we should track is wc_facebook_average_feed_generation_time - it will tell us how long the site takes to create a feed file. I would go one step further.

Good idea. What I've done is add a new transient & tracker prop for the last generation time (microtime diff, i.e. seconds). If there's an error I set it to -1 (though we could use a separate prop for that if better). We can use the same interface (track the same metric) when we add the new feed engine too.

I haven't removed the existing averaging code. I agree this is worth removing, but don't want to add complexity/risk to this PR - keep it as safe as possible. @budzanowski feel free to log an issue for that (though it may get cleaned up when we move forward with feed engine - maybe not worth it?).

@budzanowski If you're happy and don't see any issues feel free to merge this. @deltaWhiskey  will give us feedback soon, but we can follow that up separately if needed - let's get the 2.6 train ready to go 🚋

budzanowski:
It all looks very nice now.

php
```
array (
  'is-connected' => 'yes',
  'product-sync-enabled' => 'yes',
  'messenger-enabled' => 'no',
  'feed-generation-time' => 0.0,
  'feed-file-requested' => 'no',
  'ig-shopping-enabled' => 'no',
  'ig-cta-enabled' => 'no',
  'product-feed-config' => 
  array (
    'site-feed-id' => '176504340994598',
    'feed-count' => 2,
    'active-feed' => 
    array (
      'created-time' => '2021-04-26 09:16:18',
      'product-count' => 7,
      'schedule' => 
      array (
        'interval' => 'DAILY',
        'interval-count' => 1,
      ),
      'latest-upload' => 
      array (
        'end-time' => '2021-06-02 23:01:46',
        'error-count' => 0,
        'warning-count' => 8,
        'num-detected-items' => 7609,
        'num-persisted-items' => 7609,
        'url-matches-site-endpoint' => 'yes',
      ),
    ),
  ),
)
```

Good job @haszari .

We will be able to test this in the RC. Merging.

deltaWhiskey (commented):
Data types and dates look good. I don't know enough about the exact questions we'll be trying to answer with this data, but this looks like a reasonable way to collect it all and associate it with snapshots / sites.

It was a wise move to separate the collection of config info, so that it happens asynchronously with snapshot generation.

Using `ig` to represent Instagram in this PHP is reasonable, because it has the context of your nice comments. But using that abbreviation in the snapshots may confuse future analysts, as all they'll see is a list of field names. Spelling out `instagram` there would be more self-documenting.

I'm curious about why this deletes the transient during snapshot generation? When things are working normally, snapshots would only be generated once every 7 days, so this shouldn't cause problems. But, if there's a problem updating the Facebook settings for a week, the latest settings would be missing from the next snapshot. It's also possible for a snapshot to be force-sent ahead of schedule, [such as is done here](https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/includes/admin/helper/class-wc-helper.php#L801), or someone testing on the command line.

haszari:
> I'm curious about why this deletes the transient during snapshot generation? When things are working normally, snapshots would only be generated once every 7 days, so this shouldn't cause problems. But, if there's a problem updating the Facebook settings for a week, the latest settings would be missing from the next snapshot. It's also possible for a snapshot to be force-sent ahead of schedule, such as is done here, or someone testing on the command line.

Good catch - this pattern started with the `facebook_for_woocommerce_wctracker_feed_requested` transient/prop. It represents if the feed has been requested since the last snapshot. I agree that it's risky to delete the other transients, especially when sending snapshots more often or manually – I've added a follow up issue here: #1992 and will get this into 2.6. cc @budzanowski 

> Using ig to represent Instagram in this PHP is reasonable, because it has the context of your nice comments. But using that abbreviation in the snapshots may confuse future analysts, as all they'll see is a list of field names. Spelling out instagram there would be more self-documenting.

That might be worth improving. Note there is more context in the snapshot, as it's nested under Facebook. The full prop name would be `extensions_facebook-for-woocommerce_ig-shopping-enabled`. I might sneak that into #1992 :)

budzanowski:
I agree with @deltaWhiskey about the `ig` . I am changing it to `instagram` https://github.com/woocommerce/facebook-for-woocommerce/pull/1998 it will be included in the 2.6.0 RC.

PR #1976, opened 2021-05-25
remove unnecessary use of `is_debug_mode_enabled()`

rcstr:
In some cases `is_debug_mode_enabled()` is being called before calling `facebook_for_woocommerce()->log()` / `$this->get_plugin()->get_integration()->is_debug_mode_enabled()` / `facebook_for_woocommerce()->log_api_request` which also calls `is_debug_mode_enabled()`.
https://github.com/woocommerce/facebook-for-woocommerce/blob/1ba51dda56c39d8000a47753a66e9697e69cf854/class-wc-facebookcommerce.php#L408-L416

https://github.com/woocommerce/facebook-for-woocommerce/blob/1ba51dda56c39d8000a47753a66e9697e69cf854/class-wc-facebookcommerce.php#L427-L435

This PR removes the unnecessary check.

### Changelog entry

> Fix / Dev / New - Removes duplicated calls of `is_debug_mode_enabled()`

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -1036,10 +1036,8 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 			} catch ( Framework\SV_WC_API_Exception $exception ) {
 
 				$success = false;
-
-				if ( facebook_for_woocommerce()->get_integration()->is_debug_mode_enabled() ) {
-					facebook_for_woocommerce()->log( 'Could not send Pixel event: ' . $exception->getMessage() );
-				}
+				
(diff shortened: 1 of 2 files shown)


budzanowski:
Ah nice :)

budzanowski (approved):

haszari:
> Fix / Dev / New - Removes duplicated calls of is_debug_mode_enabled()

I don't think this needs a changelog entry.

I wonder if we should consider renaming `log` so it's clear it will only log if `is_debug_mode_enabled`?

haszari:
@rcstr Can you merge this? Or if there's some reason it needs work, please comment (and maybe remove from 2.6).

rcstr:
merged 🚢

PR #1982, opened 2021-05-27
Add placeholder replace paths.

budzanowski:

Since WooRelease 2.0.4 versioning placeholders are supported. If we want to use that feature we must add paths that will be evaluated for version replacement.

WooRelease docs: https://github.com/woocommerce/woorelease/blob/develop/README.md#phpdoc-version-placeholder-requirements

This features works only if WooRelease process is started with `--product_version` flag.

WooRelease 2.0.4 is obviously required. 

The easiest way to test this is:
1.  Create a branch from this branch.
2. Add code with a docblock that contains `@since x.x.x` or `@version x.x.x`.
3. Add a dummy fresh changelog.txt and readme.txt.
4. Push the changes.
5. Run a simulation release ( stop when the plugin starts the build - that part is not necessary ):

```bash
php woorelease.phar simulate --product_version=2.5.1 https://github.com/woocommerce/facebook-for-woocommerce/tree/your-branch
```
6. Inspect temporary folder for changes.

### Changelog entry

No changelog entry needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     "build": "./bin/build-zip.sh"
   },
   "woorelease": {
-		"version_replace_paths": [
-			"includes",
+    "version_replace_paths": [
+      "includes",
       "class-wc-facebookcommerce.php",
       "facebook-commerce-events-tracker.php",
       "facebook-commerce-messenger-chat.php",
--- a/package.json
+++ b/package.json
@@ -25,5 +25,17 @@
     "start": "wp-scripts start",
     "deploy:sake": "npm run build:assets && npx sake deploy --deployAssets=0",
     "build": "./bin/build-zip.sh"
-  }
+  },
+  "woorelease": {
+		"version_replace_paths": [
+			"includes",
+      "class-wc-facebookcommerce.php",
+      "facebook-commerce-events-tracker.php",
+      "facebook-commerce-messenger-chat.php",
+      "facebook-commerce-pixel-event.php",
+      "facebook-commerce.php",
+      "facebook-config-warmer.php",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -25,5 +25,17 @@
...
+  "woorelease": {
+    "version_replace_paths": [
+      "includes",
+      "class-wc-facebookcommerce.php",
+      "facebook-commerce-events-tracker.php",
+      "facebook-commerce-messenger-chat.php",
+      "facebook-commerce-pixel-event.php",
+      "facebook-commerce.php",
+      "facebook-config-warmer.php",
+      "facebook-for-woocommerce.php"
```
rcstr, line 38: I tested using [`./`](https://github.com/woocommerce/facebook-for-woocommerce/commit/83155d7ef05a16b60b99601bc8fd097d13863aba) and I got the same result. I don't mind being explicit either

```suggestion
      "./"
```

rcstr (commented):
Awesome,`version/since` was updated as expected. I played a bit with `version_replace_paths` (https://github.com/woocommerce/facebook-for-woocommerce/commit/108cebf89d17ceb90b38eeb103387a5a3f47a48f and https://github.com/woocommerce/facebook-for-woocommerce/commit/83155d7ef05a16b60b99601bc8fd097d13863aba) (also this branch) and left a suggestion.

package.json
budzanowski: I have noticed this but I just wanted to exclude the ./vendors folder from processing.

rcstr: that makes sense! 👍🏽

rcstr (approved):

PR #1989, opened 2021-06-03
Feed relaunch integration branch.

budzanowski:
### Feed relaunch integration branch.
Please add PRs numbers to the list of PRs merged into this branch.

Merged PRs:
 - #1924

### TODO
 - [] Add feed generation measurement. This was rebased with the latest master which has a bunch of tracking variables added to the WC Tracker. During the merge, I have removed the feed generation time storing for tracking. Since we are not processing all products at once but in batches, the feed file generation time has a different meaning now ( or no meaning at all ) we need to figure out if we are removing it from tracking entirely or we are replacing it with some other measurement. @haszari


haszari:
> Add feed generation measurement. This was rebased with the latest master which has a bunch of tracking variables added to the WC Tracker. During the merge, I have removed the feed generation time storing for tracking. 

That makes sense - we'll need to review the tracked info and implement equivalent/similar tracking for the new feed engine. I'll log an issue for this - #1991.

PR #1999, opened 2021-06-04
Add validation to product feed file generation.

budzanowski:
Using the new validator class exclude from the feed file the products that should not be synchronized.

Testing instructions:

0. Inspect `wp-content/uploads/facebook_for_woocommerce` folder and remove any product_catalog_hashnumber.csv files.
1. Create a site with products
2. Set some of the products to not sync ( excluded categories, no sync settings ), do that for simple products and variable products with variations ( variations also have no sync setting ).
3. From: Woocommerce -> Status -> Scheduled Actions -> Pending trigger `facebook_for_woocommerce_start_feed_generation`
4. Wait untill the In Progres section of Scheduled Actions has no more facebook_for_woocommerce/jobs/generate_feed/chain_batch` actions - this means that the generation job has finished.
5. Inspect wp-content/uploads/facebook_for_woocommerce and filnd product_catalog_....csv file. Inspect it and check if the products that were disabled in step 2 are not part of the generated file.

### Changelog entry
No changelog entry for this PR.

### DIFF
--- a/includes/Jobs/GenerateProductFeed.php
+++ b/includes/Jobs/GenerateProductFeed.php
@@ -93,6 +93,10 @@ class GenerateProductFeed extends AbstractChainedJob {
 		$processed_items = array();
 
 		foreach ( $products as $product ) {
+			// Check if product is enabled for synchronization.
+			if ( ! facebook_for_woocommerce()->get_product_sync_validator( $product )->passes_all_checks() ) {
+				continue;
+			}
 			$processed_items[] = $this->process_item( $product, $args );
 		}
 
@@ -105,6 +109,10 @@ class GenerateProductFeed extends AbstractChainedJob {
 	 * @param array $processed_items Array of product fields to write to the feed file.
 	 */
 	protected function write_processed_items_to_feed( $processed_items ) {
+		// Check if we have any items to write.
+		if ( empty( $processed_items ) ) {
+			return;
+		}
 		$this->feed_file_handler->write_to_temp_file(
 			$this->feed_data_exporter->format_items_for_feed( $processed_items )
 		);


includes/Jobs/GenerateProductFeed.php
```diff
@@ -105,6 +109,10 @@ protected function process_items( array $items, array $args ) {
 	 * @param array $processed_items Array of product fields to write to the feed file.
 	 */
 	protected function write_processed_items_to_feed( $processed_items ) {
+		// Check if we have any items to write.
+		if ( empty( $processed_items ) ) {
```
rcstr, line 113: What do you think if we log that `$processed_items` is empty?

includes/Jobs/GenerateProductFeed.php
```diff
@@ -93,6 +93,10 @@ protected function process_items( array $items, array $args ) {
 		$processed_items = array();
 
 		foreach ( $products as $product ) {
+			// Check if product is enabled for synchronization.
+			if ( ! facebook_for_woocommerce()->get_product_sync_validator( $product )->passes_all_checks() ) {
```
rcstr, line 97: I suggest we add a filter here so external developers are able to filter as well which products to exclude

rcstr (approved):
Tested on a site following the instructions and feed came out correctly. Didn't include products from the excluded categories, nor products manually set to _Do not sync_.

includes/Jobs/GenerateProductFeed.php
budzanowski: I don't think we should. This is a normal case in a lot of scenarios.

includes/Jobs/GenerateProductFeed.php
budzanowski: That is a good idea. I will create an issue for this. I will not add it now because this requires more changes than just this line. We display if the product will be synchronized in other places so all of those spots will need to be compatible with what they are showing to the user. Maybe just adding a filter at the validator class will be the simplest way but that requires investigation beyond this PR.

budzanowski: Issue created https://github.com/woocommerce/facebook-for-woocommerce/issues/2003

PR #2000, opened 2021-06-04
Move feed detection to be based of the new code.

budzanowski:
Detection code moved from https://github.com/woocommerce/facebook-for-woocommerce/pull/1977 since master now has the main class ( used for WC_Tracker props ) it was hard to cherry-pick so copy past was used.

Closes # .
# This is WIP do not merge or even test :P 

1.
2.
3.

### Changelog entry

> Fix / Dev / New - Remove duplicate visibility meta entries from postmeta table

### DIFF
--- a/includes/Feed/FeedConfigurationDetection.php
+++ b/includes/Feed/FeedConfigurationDetection.php
@@ -8,7 +8,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 use Error;
 use SkyVerge\WooCommerce\Facebook\Utilities\Heartbeat;
-use SkyVerge\WooCommerce\Facebook\Products\Feed;
 
 /**
  * A class responsible detecting feed configuration.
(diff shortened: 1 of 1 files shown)


haszari:
@budzanowski I opened an issue to clarify the goals here (because I'm a bit lost - this is a complex PR!) – #2007

In that issue I've split out 3 goals, I think this PR is focused on two of them – determining the settings on a given site, and guiding merchant to fix config. 

We might need to think more about the ideal, correct feed settings first, and then design (write) the notices and guidance based on that. If we have a solid understanding of what valid settings look like, then we can better encode the failure cases, and define how we'll guide merchants in those scenarios. 

A related approach might be to split out a separate PR for detecting settings and evaluating them, with no notices/guidance (to be handled in a follow up PR). In the current code, the notices are tightly coupled to the settings evaluation code.

budzanowski:
> We might need to think more about the ideal, correct feed settings first, and then design (write) the notices and guidance based on that. If we have a solid understanding of what valid settings look like, then we can better encode the failure cases, and define how we'll guide merchants in those scenarios.

We have discussed that some time ago. For now, we don't really do any suggestions in the code. We only point to the documentation and ask to match the settings we what we want. Code does not guide the merchants. I have planned only one message facing merchants originally. I have added the second one only when I have noticed that the situation there is quite unique and having something else displayed makes sense. As we go we could add more detailed scenarios but I don't think that we should focus on that now.

> In the current code, the notices are tightly coupled to the settings evaluation code.

I disagree about 'tightly' coupled :) I have specially created two functions that decouple the notices from the algorithm. The core logic is located in `has_valid_feed_config`. This function is consumed by `evaluate_feed_config` which responsible for dispatching the notices and scheduling checks. This decouples the notices from the algorithm. If we would want to split then the only thing that we would do is to move the `evaluate_feed_config` into a separate PR. This would be splitting hairs with the additional downside of having this PR not really testable easily.

PR #2001, opened 2021-06-07
Feed file generation performance statistics.

### LINKED ISSUE #1991: [Feed engine] Ensure new feed engine includes appropriate tracking
Follow up to #1972.

- #1972 added tracker props to help us understand how the existing feed is being used on sites.
- #1924 re-implements an improved feed engine that scales better. 
- #1989 is a feature branch focused on re-enabling feed-based sync using the new feed engine.

The goal of this issue is to ensure that we have added relevant tracker props to the new feed engine (equivalent to what we're currently tracking in 2.6+). This will allow us to measure the impact of the new feed engine when it is launched, and monitor for any issues. Ideally we can compare the old and new system by measuring similar usage metrics.

budzanowski:

Since we have changed the feed generation process we no longer can measure how long it takes to generate the feed file directly. The generation process is now split into batches. This PR instruments the tracking code with the following data collection:

 - `feed-batch-generation-average-time` - how long it takes to process one batch
 - `feed-generation-start-time` - timestamp of when the feed generation started
 - `feed-generation-end-time` - timestamp of when the feed generation ended
 - `feed-generation-batch-size` - how big are the processed batches
 - `feed-generation-batch-count` - how many batches have been processed during the feed generation

This allows us to do some inferring:

```
Total processing time ( something like CPU time ) = `feed-batch-generation-average-time` * `feed-generation-batch-count`
```
This value is equivalent in a mathematical sense to the previous `feed-generation-time`.

```
Total feed generation time = `feed-generation-end-time` - `feed-generation-start-time` 
```
This tells us how much time the process takes in its entirety from start till finish.

```
Processed products count = `feed-generation-batch-size` * `feed-generation-batch-count`
```
The number of items processed (the last batch items count may be smaller than the batch size but we don't really need perfect accuracy )

Closes #1991 .

1. Go to WooCommerce: Scheduled Actions: Pending
2. Trigger `facebook_for_woocommerce_start_feed_generation`
3. Wait until generation finishes ( no more `facebook_for_woocommerce/jobs/generate_feed/chain_batch
` actions in pending or in-progress sections.
4. Check the tracked data ( more info on how to do this can be found at https://github.com/woocommerce/facebook-for-woocommerce/pull/1972 )
5. Evaluate the entries described in the PR description.

### Changelog entry

> New - Extend WooCommerce Tracker props with feed performance statistics.

### DIFF
--- a/includes/Jobs/GenerateProductFeed.php
+++ b/includes/Jobs/GenerateProductFeed.php
@@ -56,6 +56,38 @@ class GenerateProductFeed extends AbstractChainedJob {
 		$this->feed_file_handler->write_to_temp_file(
 			$this->feed_data_exporter->generate_header()
 		);
+		// Reset the statistics counters.
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_batch_generation_average_time( 0 );
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_batch_count( 0 );
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_generation_start( time() );
+	}
+
+	/**
+	 * Handle processing a chain batch.
+	 *
+	 * @hooked {plugin_name}/jobs/{job_name}/chain_batch
+	 *
+	 * @param int   $batch_number The batch number for the new batch.
+	 * @param array $args         The args for the job.
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
+	 */
+	public function handle_batch_action( int $batch_number, array $args ) {
+		$start_time = microtime( true );
+		parent::handle_batch_action( $batch_number, $args );
+		$generation_time       = microtime( true ) - $start_time;
+		$previous_average_time = facebook_for_woocommerce()->get_tracker()->get_feed_batch_generation_average_time();
+
+		/*
+		 * This algorithm continuously updates the average time for a batch generation.
+		 * Takes the previous average with the weight equal to the number of batches for which it was calculated
+		 * combines it with the current value and calculates new average.
+		 * This is mathematically equal to summing all batch execution time and dividing by batch count.
+		 *
+		 */
+		$average_time = ( ( $batch_number - 1 ) * $previous_average_time + $generation_time ) / $batch_number;
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_batch_generation_average_time( $average_time );
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_batch_count( $batch_number );
 	}
 
 	/**
@@ -63,6 +95,7 @@ class GenerateProductFeed extends AbstractChainedJob {
 	 */
 	protected function handle_end() {
 		$this->feed_file_handler->replace_feed_file_with_temp_file();
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_generation_end( time() );
 	}
 
 	/**
@@ -192,7 +225,7 @@ class GenerateProductFeed extends AbstractChainedJob {
 	 *
 	 * @return int
 	 */
-	protected function get_batch_size(): int {
+	public function get_batch_size(): int {
 		return 15;
 	}
 
--- a/includes/Utilities/Tracker.php
+++ b/includes/Utilities/Tracker.php
@@ -29,11 +29,33 @@ class Tracker {
 	const TRANSIENT_WCTRACKER_LIFE_TIME = 2 * WEEK_IN_SECONDS;
 
 	/**
-	 * Transient key name; how long it took to generate the most recent feed file, or zero if it failed.
+	 * Transient key to store how long it took to generate one feed file products batch.
 	 *
 	 * @var string
 	 */
-	const TRANSIENT_WCTRACKER_FEED_GENERATION_TIME = 'facebook_for_woocommerce_wctracker_feed_generation_time';
+	const TRANSIENT_WCTRACKER_FEED_BATCH_GENERATION_AVERAGE_TIME = 'facebook_for_woocommerce_wctracker_feed_batch_generation_average_time';
+
+	/**
+	 * Transient key to store feed generation start timestamp.
+	 *
+	 * @var string
+	 */
+	const TRANSIENT_WCTRACKER_FEED_GENERATION_START_TIME = 'facebook_for_woocommerce_wctracker_feed_generation_start_time';
+
+
+	/**
+	 * Transient key to store feed generation end timestamp.
+	 *
+	 * @var string
+	 */
+	const TRANSIENT_WCTRACKER_FEED_GENERATION_END_TIME = 'facebook_for_woocommerce_wctracker_feed_generation_end_time';
+
+	/**
+	 * Transient key to store feed generation batch count.
+	 *
+	 * @var string
+	 */
+	const TRANSIENT_WCTRACKER_FEED_GENERATION_BATCH_COUNT = 'facebook_for_woocommerce_wctracker_feed_generation_batch_count';
 
 	/**
 	 * Transient key name; true if feed has been requested by Facebook.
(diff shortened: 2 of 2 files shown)


budzanowski:
My site stats:
```PHP
  'feed-batch-generation-average-time' => 0.41999416470067003,
  'feed-generation-start-time' => 1623039653,
  'feed-generation-end-time' => 1623040361,
  'feed-generation-batch-size' => 15,
  'feed-generation-batch-count' => 522,
```

includes/Jobs/GenerateProductFeed.php
```diff
@@ -56,13 +56,46 @@ protected function handle_start() {
...
+	 * @param int   $batch_number The batch number for the new batch.
+	 * @param array $args         The args for the job.
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
+	 */
+	public function handle_batch_action( int $batch_number, array $args ) {
+		$start_time = microtime( true );
+		parent::handle_batch_action( $batch_number, $args );
+		$generation_time       = microtime( true ) - $start_time;
+		$previous_average_time = facebook_for_woocommerce()->get_tracker()->get_feed_batch_generation_average_time();
```
haszari, line 79: I think it's best to avoid updating transients that will be used for tracking in the middle of the batch. There is a risk that the snapshot will contain partial data.

I'd suggest `GenerateProductFeed` uses its own (private) transient to accumulate the data, and then update the tracker on completion.

haszari (commented):
This is looking good @budzanowski but I think we need some changes so the new and old metrics are comparable. We can add new metrics for the new feed engine (if needed), but the essential goal here is that we have the same info being tracked with new and old engine.

#### Essential tracker props
✅  `feed-file-requested` is looking good, I see it's [already in the feature branch PR](https://github.com/woocommerce/facebook-for-woocommerce/pull/1989/files#diff-186b9e2bedad93582d17d38c41a65579eb550cc1fdad63747e1f0836750c927dR44).

With the current feed code, `feed-generation-time` is a total time in seconds. Although we're now calculating in batches, I think we should set up a comparable metric in the new feed code. Note also this prop should be zero if feed generation is disabled or has failed.

Here's what I'm thinking:

- In `GenerateProductFeed`, measure how long each batch job takes and accumulate (add) these in a (private) transient. Could accumulate in transient directly (e.g. read, add, save at end of batch), or store an array in transient and sum when feed generation is complete.
- When feed generation completes (or fails), update the tracker prop transient, so that it always represents the complete last feed job (and is comparable with current value).
- Ensure the value is in seconds (not sure if this is the case currently).

#### Additional props
Batch size doesn't seem essential to track, though it might be useful if sites can override - i.e. in conjunction with #2002.

I don't think start time, end time and batch count are particularly useful, I'd remove them for now unless they're needed. Let me know if you have a plan for how we'd use this data.

haszari:
Do we need to take care in case feed batches are run in parallel or out of order?

budzanowski:
> Do we need to take care in case feed batches are run in parallel or out of order?

No. The Batches are synchronous. They are chained together:
https://github.com/woocommerce/action-scheduler-job-framework/blob/trunk/src/AbstractChainedJob.php#L137-L145

budzanowski:
> I don't think start time, end time, and batch count are particularly useful, I'd remove them for now unless they're needed. Let me know if you have a plan for how we'd use this data.

They have two purposes:
1) For statistics they will show how long does the process really take. Maybe for a very big store, the total processing time will be >24h. This will be interesting to know.
2) For setting up generation time. The system will know that the next generation should start at a particular time. Upcoming PR.

> With the current feed code, feed-generation-time is a total time in seconds. Although we're now calculating in batches, I think we should set up a comparable metric in the new feed code. Note also this prop should be zero if feed generation is disabled or has failed.

> Here's what I'm thinking:

> In GenerateProductFeed, measure how long each batch job takes and accumulate (add) these in a (private) transient. Could accumulate in transient directly (e.g. read, add, save at end of batch), or store an array in transient and sum when feed generation is complete.
When feed generation completes (or fails), update the tracker prop transient, so that it always represents the complete last feed job (and is comparable with current value).
Ensure the value is in seconds (not sure if this is the case currently).

I will do the changes.

budzanowski:
Please note that the main PR description no longer reflects the code after the changes done to address the review. We are currently only tracking:
`feed-generation-time` - how long it takes to generate the feed file in total
`feed-generation-batch-size` - how big are the batches

The feed start and end times are also corrected but have been moved out of the tracker and are not part of the tracked data.
They will be used for scheduling purposes.

includes/Jobs/GenerateProductFeed.php
```diff
@@ -219,4 +269,43 @@ protected function get_batch_size(): int {
...
+		update_option( self::OPTION_FEED_GENERATION_TIME, $time );
+	}
+
+	/**
+	 * Get feed generation execution length.
+	 *
+	 * @since x.x.x
+	 */
+	public function get_feed_file_generation_time() {
+		get_option( self::OPTION_FEED_GENERATION_TIME, 0 );
```
rcstr, line 288: missing return 
```suggestion
		return get_option( self::OPTION_FEED_GENERATION_TIME, 0 );
```

includes/Jobs/GenerateProductFeed.php
```diff
@@ -56,13 +77,42 @@ protected function handle_start() {
...
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
+	 */
+	public function handle_batch_action( int $batch_number, array $args ) {
+		$start_time = microtime( true );
+		parent::handle_batch_action( $batch_number, $args );
+		$generation_time = microtime( true ) - $start_time;
+
+		// Add processing time to previously accumulated execution times.
+		$accumulated_time  = get_option( self::OPTION_FEED_GENERATION_TIME, 0 );
```
rcstr, line 102: We can use the _getter_ here instead
```suggestion
		$accumulated_time  = $this->get_feed_file_generation_time();
```

includes/Jobs/GenerateProductFeed.php
```diff
@@ -219,4 +269,43 @@ protected function get_batch_size(): int {
...
+	public function track_feed_file_generation_time( $time ) {
+		update_option( self::OPTION_FEED_GENERATION_TIME, $time );
+	}
+
+	/**
+	 * Get feed generation execution length.
+	 *
+	 * @since x.x.x
+	 */
+	public function get_feed_file_generation_time() {
```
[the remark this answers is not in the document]
rcstr: > These methods should all be private - they are internal to GenerateProductFeed, if other code used them it could impact the measurement.

I agree, at least the setters should be `private`

includes/Jobs/GenerateProductFeed.php
```diff
@@ -56,13 +77,42 @@ protected function handle_start() {
...
+		$this->track_feed_file_generation_time( $accumulated_time );
 	}
 
 	/**
 	 * Called after the finishing the job.
 	 */
 	protected function handle_end() {
 		$this->feed_file_handler->replace_feed_file_with_temp_file();
+		$this->track_feed_file_generation_end( time() );
+		facebook_for_woocommerce()->get_tracker()->track_feed_file_generation_time(
```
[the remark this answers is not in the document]
budzanowski: The AS framework ensures that 'handle_end' is only triggered if the processing has finished.
If the chain is broken by killing the chain action then the next chain link will not be triggered - hence processing will not reach this point.
If there is an error the chain will also get broken and again we will not update the feed generation time and it will remain 0 - indicating that the failure has happened.

budzanowski:
> Just to clarify the most important part of this PR as I see it: feed-generation-time represents a complete feed generation time, with no errors; if there are errors, the tracked value should be zero.

> When testing this we'll need to confirm the value is accurate no matter what happens, for example user cancels batch jobs somehow, or individual batches hit server resource limits (or anything else that could go wrong).

Please check https://github.com/woocommerce/facebook-for-woocommerce/pull/2001#discussion_r648080022

budzanowski:
All comments and suggestions addressed.

PR #2002, opened 2021-06-07
Add/filter to change batch size

budzanowski:

Add a simple filter that allows modification of the processed batch size during the feed file generation job.
This will allow some of the sites ( if necessary ) to tweak the speed of the feed generation process.

1. Add a filter function that modifies the size of the batch.
2. Trigger the action that starts the feed file generation.
3. Use PHP debug to check that the batch size is now modified.

### Changelog entry

> New - Add filter that allows modification of the feed batch processing size.

### DIFF
--- a/includes/Jobs/GenerateProductFeed.php
+++ b/includes/Jobs/GenerateProductFeed.php
@@ -197,6 +197,10 @@ class GenerateProductFeed extends AbstractChainedJob {
 		 * Feed batch size filter.
 		 *
 		 * This filter allows modification of how many items will be processed in one batch during the feed file generation.
+		 * Increasing the number of items per batch can potentially speed up processing on some of the sites.
+		 * Bigger number means reducing the number of required batches, but at the same time increase the memory requirements for the process.
+		 * Smaller number of items per batch means that the memory requirements are smaller for the process and the stability of the system is better.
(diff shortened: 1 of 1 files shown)


includes/Jobs/GenerateProductFeed.php
```diff
@@ -193,7 +193,18 @@ public function get_plugin_name(): string {
 	 * @return int
 	 */
 	protected function get_batch_size(): int {
-		return 15;
+		/**
+		 * Feed batch size filter.
+		 *
+		 * This filter allows modification of how many items will be processed in one batch during the feed file generation.
+		 * Some, especially big sites with big products catalog, may want to increase this value in order to process the catalog faster.
+		 * This requires careful approach, bumping the value too high may lead to out of memory issues.
```
haszari, line 201: Could make this sentence clearer by stating the tradeoffs. Here's a rough idea, I think you understand the tradeoffs better than me 😁 

- more products per batch:
   - more memory required per batch
   - fewer batches required overall
- fewer products per batch:
  - lower overhead / resources required per batch, increase stability
  - more batches needed, feed generation make take longer overall

haszari (approved):
Great idea! Do you have any other scaling parameters in mind?

Pre-approving, I added some feedback on the docs/naming but this is 👍

PR #2011, opened 2021-06-09
Fix visibility parameter value for items if not visible.

### LINKED ISSUE #2009: Product feed visibility incorrect value
It looks like the feed is sending staging value in the visibility fieled:
image

Only `published` and hidden `values` are supported:
image

budzanowski:

The hidden item's visibility value was wrong. I don't really know why it was `staging`  when documentation clearly states that the values should be `hidden` or `published.

Closes #2009 .

1. Generate the feed with hidden items
2. Make sure that the feed file gets uploaded
3. Wait for feed report
4. Check for the issue mentioned in #2009

### Changelog entry

*  Fix - Feed visibilitie fild value for hidden items.

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -148,7 +148,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	const FB_SHOP_PRODUCT_VISIBLE = 'published';
 
 	/** @var string the API flag to set a product as not visible in the Facebook shop */
-	const FB_SHOP_PRODUCT_HIDDEN = 'staging';
+	const FB_SHOP_PRODUCT_HIDDEN = 'hidden';
 
 	/** @var string @deprecated  */
 	const FB_CART_URL = 'fb_cart_url';
--- a/includes/Products.php
+++ b/includes/Products.php
@@ -280,7 +280,7 @@ class Products {
 	 * @since 1.10.0
 	 *
 	 * @param \WC_Product $product product object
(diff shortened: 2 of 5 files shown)


facebook-commerce.php
```diff
@@ -148,7 +148,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	const FB_SHOP_PRODUCT_VISIBLE = 'published';
 
 	/** @var string the API flag to set a product as not visible in the Facebook shop */
-	const FB_SHOP_PRODUCT_HIDDEN = 'staging';
+	const FB_SHOP_PRODUCT_HIDDEN = 'hidden';
```
haszari, line 151: 👌

haszari (approved):
Good find! I assume this is causing actual problems – i.e. `staging` items are showing up as feed errors or are visible when they should be hidden.

@budzanowski do you know what the impact of the `staging` bug is specifically? Also, do we have this bug on the old feed code, or is it only in the new engine?

rcstr (approved):
LGTM. updated a few references to _staging_ in comments and unit tests https://github.com/woocommerce/facebook-for-woocommerce/pull/2011/commits/080abf7224b4b1c71d86324465af336d0b4921c2

facebook-commerce.php
budzanowski: Old code has the same problem.

budzanowski:
Thanks @rcstr

PR #2014, opened 2021-06-11
Smart feed generation time.

budzanowski:

We want to make sure that the feed file generation finishes before Facebook requests the file. This PR modifies feed scheduling in such a way that when Facebook requests the feed file, the feed file should be fresh and ready :)

1. Set this PR up on a site that has a valid connection and feed schedule in FB settings.
2. Run one full generation of the feed file. Roughly check how much time it took. From start till finish.
3. Cancel the next occurrence of `facebook_for_woocommerce_start_feed_generation` Action Scheduler action.
4. Refresh admin interface. A new `facebook_for_woocommerce_start_feed_generation` will be created. Inspect Schedule date.
5. Confirm that the next feed generation time is scheduled roughly 1.5 * feed generation ( wall ) time before the next feed schedule according to the FB settings.

This PR already takes into consideration that there are other valid feed schedules than `DAILY` even if other parts of the code do not support this.

My example scenario:

image
The next scheduled feed fetch will happen on Wednesday at 13:16 ( UTC + 2 ).

image
The next feed generation is planned for Wednesday at 10:59 ( UTC + 0 ).

This means that the feed file generation will be started roughly 15 minutes before the next feed fetch. My feed generates in roughly 10 minutes.

### Changelog entry

> New - Smart feed generation scheduling.

### DIFF
--- a/includes/Feed/FeedConfigurationDetection.php
+++ b/includes/Feed/FeedConfigurationDetection.php
@@ -8,12 +8,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 use Error;
 use SkyVerge\WooCommerce\Facebook\Utilities\Heartbeat;
-use SkyVerge\WooCommerce\Facebook\Products\Feed;
 
 /**
  * A class responsible detecting feed configuration.
  */
 class FeedConfigurationDetection {
+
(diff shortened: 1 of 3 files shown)


budzanowski:
@woocommerce/tetris We could consider besides multiplying the value by 1.5 also adding a constant factor to cancel AS time jitter for small feeds that. Something like 10 minutes.

PR #2019, opened 2021-06-14
Fix #2018 - Correct Content IDs in Purchase event

### LINKED ISSUE #2018: Purchase Event Content IDs mismatch from earlier events for variable products
## 🔎 Isolate the bug

## ✍️ Describe the bug
I have a variable product with SKU of `GTBD-MP` and ID 314 and it has 2 variants, with SKUs `GTBD-MP-1` and `GTBD-MP-2`, and IDs 315 and 316 respectively. I am using AJAX to add items to cart. When I add `GTBD-MP-1` to the cart, it correctly sends an `AddToCart` event with a content ID of `GTBD-MP-1_315`. This occurs for other events as well such as `InitiateCheckout`. However, when it gets to the `Purchase` event, it is using a Content ID of `GTBD-MP_314`. As this does not match any of the prior events Facebook is showing me an error in the Events console as per below screenshot. This looks like it is using the parent product instead of the variation.

image

Here is what the `AddToCart` event looks like:
image

Here is what the `Purchase` event looks like (The Content ID does not match the `AddToCart` event):
image

## 🚶‍♀️ Steps to reproduce

1. Ensure Facebook for Woocommerce is installed and working correctly (i.e. events are flowing through)
1. Create a variable product. Give it a SKU of "X" and let's say it has an ID of 1.
1. Add a variation under this product. Give it a SKU of "Y" and let's say it has an ID of 2.
1. Go through a purchase and track the parameters sent with the events on Facebook
1. All events up to purhcase correctly show the Content ID of "Y_2" but the Purchase event has a Content ID of "X_1"

## ✔️ Expected behavior
In all events, Facebook should receive a content ID of "Y_2"

dgurram:

Fixes mismatch of Content IDs between events. Use `get_product()` on order item which applies the logic - if `variation_id` is set then prefer that over `product_id` in order to get the relevant product. That ends up producing the same Content IDs as earlier events in the purchase. It works the same as before for simple products as `variation_id` is not set on the order item.

Closes #2018.

As per Steps to Reproduce in #2018.

### Changelog entry

> Fix - Correct Content IDs in Purchase event for variable products

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -889,8 +889,13 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 			foreach ( $order->get_items() as $item ) {
 
-				if ( $product = isset( $item['product_id'] ) ? wc_get_product( $item['product_id'] ) : null ) {
-
+				$product = null;
+				if ( isset( $item['variation_id'] ) ) {
+					$product = wc_get_product( $item['variation_id'] );
+				} else if ( isset( $item['product_id'] ) ) {
+					$product = wc_get_product( $item['product_id'] );
+				}
+				if ( $product ) {
 					$product_ids[]   = \WC_Facebookcommerce_Utils::get_fb_content_ids( $product );
 					$product_names[] = $product->get_name();
 
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -889,12 +889,8 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 			foreach ( $order->get_items() as $item ) {
 
-				$product = null;
-				if ( isset( $item['variation_id'] ) ) {
-					$product = wc_get_product( $item['variation_id'] );
-				} else if ( isset( $item['product_id'] ) ) {
-					$product = wc_get_product( $item['product_id'] );
-				}
+				$product = $item->get_product();
+				
 				if ( $product ) {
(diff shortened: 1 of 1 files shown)


haszari:
Thanks for logging and fixing this @dgurram ! @rcstr can you review this sometime this week? If it's working well and safe perhaps we can include in next release.

facebook-commerce-events-tracker.php
```diff
@@ -889,8 +889,13 @@ public function inject_purchase_event( $order_id ) {
...
 
-				if ( $product = isset( $item['product_id'] ) ? wc_get_product( $item['product_id'] ) : null ) {
-
+				$product = null;
+				if ( isset( $item['variation_id'] ) ) {
+					$product = wc_get_product( $item['variation_id'] );
+				} else if ( isset( $item['product_id'] ) ) {
+					$product = wc_get_product( $item['product_id'] );
+				}
+				if ( $product ) {
```
rcstr, line 898: What do you think if we use `$item->get_product()` which will do the same check?

Code reference: https://github.com/woocommerce/woocommerce/blob/trunk/includes/class-wc-order-item-product.php#L326-L331

```suggestion
				$product = $item->get_product();
				
				if ( $product ) {
```

rcstr (commented):
thank you for opening this PR @dgurram. I left a suggestion, let me know your thoughts 👍🏽

facebook-commerce-events-tracker.php
dgurram: Great suggestion - that's far better. I didn't know that existed.

rcstr (approved):
thank you @dgurram. Changes work great

Issue #2029, opened 2021-06-21
Remove scheduled background sync task


rcstr:
### Description
`SkyVerge\WooCommerce\Facebook\Products\Sync\Background` is used to schedule the background sync, with #2028 we will switch to feed file sync. So we need to unschedule the task from customer's sites and clean up the data (see `SkyVerge\WooCommerce\Facebook\Jobs\CleanupSkyvergeFrameworkJobOptions`)

haszari:
Closing and removing from 2.6.1. 

- We're going with a targeted feature flag to allow disable of full batch API background sync.
- Any other improvements or changes to product sync overall can be looked at as part of ongoing work, we may or may not need what's requested in this issue.

Issue #2025, opened 2021-06-17
Use Feed sync instead of Background Sync on `handle_connect` event


rcstr:
The background batch-api sync uses the SkyVerge job framework, which updates an option (`wp_options` table) regularly as the job is processed. This job status option includes all product IDs that need to be synched.

In the case of the a full catalog sync, on a large store, this can cause performance issues and is generally a bad idea. The option is very large and is updated often, which may cause issues if database changes are logged (as they are on some hosts). 

The goal of this issue is to disable the use of the background sync for full catalog sync. There are at least two situations that do this: `Sync products` button in admin, and the initial full sync triggered after successful connection to Facebook (i.e. after onboarding).

See also related issues #2023 #2022 #2013

haszari:
Removing this from 2.6.1 - we're going to provide a targeted feature flag to allow automated opt-out of full background sync.

In future we may look at switching the full sync mechanism from batch API to feed, but we need to prioritise that alongside all other product sync improvements (and may not be appropriate if we make batch API sync scale better). With that in mind, I'm closing this issue.

Issue #2023, opened 2021-06-15
Use core Options API instead of direct DB writes


aidvu:
## 🔎 Isolate the bug

- [X] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [X] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
The SkyVerge job framework is not using the [Options API](https://codex.wordpress.org/Options_API). It is writing directly to the DB, which disallows hosting providers to add some limitations that would break the hosting, or the site itself.

e.g. A site with millions of products is going to first write ~5MBs to the options table, then it's going to update it by adding/removing small chunks of data, but still writing ~5MBs every time. With some simple math, this is going to be over a terabyte after the first 200k rows.

I'd suggest either chunking these updates or completely changing how the sync works. It's better to have 1 million rows of specific easily queryable data, than one big chunk you have to `json_encode/decode` every time you want to make a change.

## 🚶‍♀️ Steps to reproduce

1. Generate a site with million of products
1. Run the full product sync

## ✔️ Expected behavior
Nothing breaks, or goes OOM, or writes terabytes of binary logs.

## 🗃 Logs

<details>
	
</details>

haszari:
Tentatively added this to 2.7 – not sure if there's a quick safe way to improve this, but it's definitely a pain point. Thanks for reporting @aidvu.

Some notes about impact:

- The [`Products/Sync/Background`](https://github.com/woocommerce/facebook-for-woocommerce/blob/master/includes/Products/Sync/Background.php) class is used for on-demand full product sync using Batch API (as I understand it). This can be triggered from `Marketing > Facebook > Product sync > Sync products` button. 
- It's implemented on top of the SkyVerge job framework, which stores the job state in an option. 
- That state grows with the number of products.

Can reproduce this issue by setting up a site with many products, then triggering a full sync. This will generate large options rows for the job state, and as the number of products increases, the option gets larger.

aidvu:
The problem isn't only the size of the row. It's consequent (small) updates which write the whole thing over and over with minor changes. 

As for a fix, maybe extend the class that updates the data, and override the methods to use the options api. I haven't really looked if it's possible.

haszari:
[...]

PR #2033, opened 2021-06-22
Add hook to allow disable of full batch api sync (e.g. for large stores) to avoid potential performance issues

### LINKED ISSUE #2022: Disabling product sync globally
## 🔎 Isolate the bug

- [X] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug

While trying to disable the product sync entirely, was told by @haszari to use `wc_facebook_legacy_feed_file_generation_enabled`. This doesn't disable explicitly running the sync via the `wp-admin` interface and clicking `Sync products`.

In the end explicitly disabled the action in `includes/AJAX.php` by returning early in `sync_products`.

Maybe I misunderstood the flag, but I think there should be an easy way to either disable this, or a filter to prevent sync from happening. For sites with a lot of products the sync writes hundreds of GBs to the binary logs, and it's fairly hard to disable.

## 🚶‍♀️ Steps to reproduce

1. Run `wp option set wc_facebook_legacy_feed_file_generation_enabled no` in `wp-cli`
2. Go the the Products tab in the UI and click `Sync products`

## ✔️ Expected behavior
Message is shown that the syncing is disabled.

## 🗃 Logs

<details>
	
</details>

haszari:
Fixes #2022, #2035

See also related #2023 #2025 #2029 

This PR provides a hook to allow developers (or hosting platforms) to disable specific full-catalog sync operations (via Batch API). See below `Why?` for more background and context.

Specifically these two flows are affected:

- Onboarding & initial connection. The initial full catalog sync is only triggered if the filter allows it.
- On-demand sync using button in `Marketing > Facebook > Product sync`. If full batch API sync is disabled via filter, an error will be shown.
- AJAX full sync endpoint (used for button above) returns an error if site has large number of products.

The default behaviour is that the full batch-api background sync is allowed (consistent with previous releases).

Some example use-cases for the hook.

- On shared hosting or other resource-limited environments, to disable full batch API sync for large-scale sites (5000+ products, or completely disable).
- On hosts that have detailed database logging, to disable batch API sync more aggressively (e.g. 1000+ products).

Example use of hook:

```php
function wpcomsh_disallow_fb_for_woo_full_batch_api_sync( $allow_full_sync, $product_count ) {
	// Disable only for sites with a large number of products.
	$max_products_for_safe_full_sync = 10;
	if ( $product_count > $max_products_for_safe_full_sync ) {
		return false;
	}

	return $allow_full_sync;
}
add_filter( 'facebook_for_woocommerce_allow_full_batch_api_sync', 'wpcomsh_disallow_fb_for_woo_full_batch_api_sync', 10, 2 );
```

Also in this PR the `get_product_count()` function has been optimised to use `wp_count_posts`. We're passing the total number of products to the hook. The previous query was loading all the IDs into memory, which is unnecessary and could impact performance for larger stores.

##### Why?
A full catalog sync using the batch API and the [`Products\Sync\Background`](https://github.com/woocommerce/facebook-for-woocommerce/blob/master/includes/Products/Sync/Background.php) task can cause issues.

The `Background` class uses the [`SkyVerge` background job framework](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/utilities/class-sv-wp-background-job-handler.php). This adds an option to the database representing the job status and progress, which includes all product IDs that are included in the sync job. 

There are two risks with this:

- The option is large, and grows with the number of products that need to be synched => large database size.
- The option is re-written constantly => large database traffic/throughput. This can generate significant database logs (depending on config of course).

Long term, we may need to tweak or replace the `Products\Sync\Background` so we have a scalable, robust way to sync full catalog. This is a medium-term patch to prevent this sync mechanism causing issues for merchants and web hosts, and to give more control over whether full batch sync happens.

##### How does this impact stores, and the product sync feature?
- Disabling the initial sync means that new stores (over the num products threshold) will not see their products in Facebook catalog by default. Products will incrementally sync as they are added or edited; or potentially sites can use feed-based sync.
- The manual full sync button will show an error (see below). Should be low impact, as it's not intended to be used regularly; is intended as a workaround for sync issues.
- Note: this will not affect any stores by default - a filter function needs to be added to prevent any syncs from occurring.

Screen Shot 2021-06-28 at 1 06 02 PM

##### Does this solve the issue?
It may be still possible to trigger a similar problem via code or API (or UI!), by marking a large number of products as needing sync (e.g. make a trivial edit). This PR doesn't fix the root of the problem (`Products\Sync\Background` scalability). One potential workaround: #2034

Add a filter for `facebook_for_woocommerce_allow_full_batch_api_sync` to disable full sync, e.g. for sites with > 5000 products.

Two main flows to test:

1. Full sync on-demand via admin button.
2. Full sync triggered automatically after onboarding.
3. Any other way of triggering a full batch API sync.

Test to ensure the filter disables full sync appropriately.

### Changelog entry

> Dev: Add `facebook_for_woocommerce_allow_full_batch_api_sync` filter to allow external developers to disable or enable full catalog sync (batch API)

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -809,7 +809,19 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	 * @since 2.6.1
 	 */
 	public function allow_full_batch_api_sync() {
-		return ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC );
+		/**
+		 * Allow full batch api sync to be enabled or disabled.
+		 *
+		 * @param bool $allow Is full batch allowed based on threshold (@see WC_Facebookcommerce_Integration::MAX_PRODUCTS_FOR_FULL_SYNC)
+		 * @param int $product_count Number of products in store
+		 *
+		 * @since 2.6.1
+		 */
+		return apply_filters(
+			'facebook_for_woocommerce_allow_full_batch_api_sync',
+			( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC ),
+			$this->get_product_count()
+		);
 	}
 
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -800,13 +800,13 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		return $product_counts->publish;
 	}
 
-
 	/**
 	 * Should full batch-API sync be allowed?
 	 *
 	 * Used to disable various full sync UI/APIs to avoid performance impact.
 	 *
 	 * @return boolean True if full sync is safe.
+	 * @since 2.6.1
 	 */
 	public function allow_full_batch_api_sync() {
 		return ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC );
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -113,12 +113,6 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var string custom taxonomy FB product set ID */
 	const FB_PRODUCT_SET_ID = 'fb_product_set_id';
 
-	/**
-	 * @var int Maximum number of products for full sync.
-	 * Used to disable full batch-API sync flows which may cause performance issues.
-	 **/
-	const MAX_PRODUCTS_FOR_FULL_SYNC = 5000;
-
 	/** @var string|null the configured product catalog ID */
 	public $product_catalog_id;
 
@@ -804,14 +798,13 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/**
 	 * Should full batch-API sync be allowed?
 	 *
-	 * Used to disable various full sync UI/APIs to avoid performance impact.
+	 * May be used to disable various full sync UI/APIs to avoid performance impact.
 	 *
 	 * @return boolean True if full batch sync is safe.
 	 * @since 2.6.1
 	 */
 	public function allow_full_batch_api_sync() {
-		// By default, disable batch API sync if store has a large number of products.
-		$default_allow_sync = ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC );
+		$default_allow_sync = true;
 
 		/**
 		 * Allow full batch api sync to be enabled or disabled.
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -809,17 +809,19 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	 * @since 2.6.1
 	 */
 	public function allow_full_batch_api_sync() {
+		$default = $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC;
+
 		/**
 		 * Allow full batch api sync to be enabled or disabled.
 		 *
-		 * @param bool $allow Is full batch allowed based on threshold (@see WC_Facebookcommerce_Integration::MAX_PRODUCTS_FOR_FULL_SYNC)
+		 * @param bool $allow Is full batch allowed?
 		 * @param int $product_count Number of products in store
 		 *
 		 * @since 2.6.1
 		 */
 		return apply_filters(
 			'facebook_for_woocommerce_allow_full_batch_api_sync',
-			( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC ),
+			$default,
 			$this->get_product_count()
 		);
 	}
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -116,7 +116,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	 * @var int Maximum number of products for full sync.
 	 * Used to disable full batch-API sync flows which may cause performance issues.
 	 **/
-	const MAX_PRODUCTS_FOR_FULL_SYNC = 500;
+	const MAX_PRODUCTS_FOR_FULL_SYNC = 5000;
 
 	/** @var string|null the configured product catalog ID */
 	public $product_catalog_id;
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -785,26 +785,14 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		return $product_item_ids;
 	}
 
-
 	/**
-	 * Gets the total of published products.
+	 * Gets the total number of published products.
 	 *
 	 * @return int
 	 */
 	public function get_product_count() {
-
-		$args = array(
-			'post_type'      => 'product',
-			'post_status'    => 'publish',
-			'posts_per_page' => -1,
-			'fields'         => 'ids',
-		);
-
-		$products = new WP_Query( $args );
-
-		wp_reset_postdata();
-
-		return $products->found_posts;
+		$product_counts = wp_count_posts( 'product' );
+		return $product_counts->publish;
 	}
 
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -112,6 +112,11 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var string custom taxonomy FB product set ID */
 	const FB_PRODUCT_SET_ID = 'fb_product_set_id';
 
+	/**
+	 * @var int Maximum number of products for full sync.
+	 * Used to disable full batch-API sync flows which may cause performance issues.
+	 **/
+	const MAX_PRODUCTS_FOR_FULL_SYNC = 500;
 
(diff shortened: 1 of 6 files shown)


facebook-commerce.php
```diff
@@ -112,6 +112,11 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	/** @var string custom taxonomy FB product set ID */
 	const FB_PRODUCT_SET_ID = 'fb_product_set_id';
 
+	/**
+	 * @var int Maximum number of products for full sync.
+	 * Used to disable full batch-API sync flows which may cause performance issues.
+	 **/
+	const MAX_PRODUCTS_FOR_FULL_SYNC = 500;
```
haszari, line 119: TBC - we need to determine what the appropriate threshold is. It may need to be smaller, but ideally we want this as large as we can safely get away with, to reduce the number of impacted stores.

facebook-commerce.php
```diff
@@ -785,26 +790,26 @@ private function find_variation_product_item_ids( $product_group_id ) {
...
-		$args = array(
-			'post_type'      => 'product',
-			'post_status'    => 'publish',
-			'posts_per_page' => -1,
-			'fields'         => 'ids',
-		);
-
-		$products = new WP_Query( $args );
-
-		wp_reset_postdata();
```
haszari, line 805: I've reimplemented this function so it's more memory efficient. Previously the query would load all product IDs; `wp_count_posts()` uses a `COUNT()` SQL statement and doesn't need to load any actual data.

This is not essential to the success of this PR but seems a useful improvement while we're here!

includes/Handlers/Connection.php
```diff
@@ -301,7 +301,10 @@ public function handle_connect() {
 			$this->update_system_user_id( $system_user_id );
 			$this->update_installation_data();
 
-			facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_all_products();
+			// Only trigger initial full batch-api sync if the store doesn't have large product count.
+			if ( $facebook_for_woocommerce()->get_integration()->allow_full_batch_api_sync() ) {
+				facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_all_products();
+			}
```
rcstr, line 307: We can define `wc_facebook_enable_product_sync` as `false` here to avoid BG sync running in the background
```suggestion
			// Only trigger initial full batch-api sync if the store doesn't have large product count.
			if ( $facebook_for_woocommerce()->get_integration()->allow_full_batch_api_sync() ) {
				facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_all_products();
			} else  {
				update_option( \WC_Facebookcommerce_Integration::SETTING_ENABLE_PRODUCT_SYNC, 'no' );
			}
```

rcstr (commented):
Changes look good, haven't fully tested yet. I think this will be a good place to define `wc_facebook_enable_product_sync` to be sure sync doesn't run

facebook-commerce.php
haszari: I'm leaning towards 5000 or maybe 2500 here - the site that triggered the issue had 19k products.

includes/Handlers/Connection.php
haszari: Is this needed to prevent a full sync? I don't think we want to completely disable product sync, just prevent a full catalog sync using `Background` job class (which has issue with the huge `wp_option`). 

If we do need to disable the feature like this, we run the risk of merchant re-enabling it in the UI.

facebook-commerce.php
rcstr: I have 2.6.0 running with 2k-5k products and products have been synced, these sites are running on a shared account on Siteground.

rcstr (approved):

facebook-commerce.php
haszari: I'm thinking we expose a filter for max products for full sync, then hosts or sites can tweak this threshold to suit.

includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -174,14 +174,17 @@ class="button button-large button-primary"
...
+		$show_sync_button = $integration->allow_full_batch_api_sync() &&
+			facebook_for_woocommerce()->get_connection_handler()->is_connected();
 		?>
 
 		<h2>
 
 			<?php esc_html_e( 'Product sync', 'facebook-for-woocommerce' ); ?>
 
-			<?php if ( facebook_for_woocommerce()->get_connection_handler()->is_connected() ) : ?>
+			<?php if ( $show_sync_button ) : ?>
```
haszari, line 187: Hmm, maybe we should not hide the button, so it's clear to merchant why it's disabled! 🤔

rcstr:
@haszari - I have applied the proposed changes in https://github.com/woocommerce/facebook-for-woocommerce/pull/2036#pullrequestreview-691303356

1. Added `facebook_for_woocommerce_allow_full_batch_api_sync` filter cd58075
2. Updated _Sync products_ button to be shown if site is connected, but show an error 70731bb

facebook-commerce.php
```diff
@@ -785,26 +790,38 @@ private function find_variation_product_item_ids( $product_group_id ) {
...
+	 * Used to disable various full sync UI/APIs to avoid performance impact.
+	 *
+	 * @return boolean True if full sync is safe.
+	 * @since 2.6.1
+	 */
+	public function allow_full_batch_api_sync() {
+		/**
+		 * Allow full batch api sync to be enabled or disabled.
+		 *
+		 * @param bool $allow Is full batch allowed based on threshold (@see WC_Facebookcommerce_Integration::MAX_PRODUCTS_FOR_FULL_SYNC)
```
haszari, line 815: Can probably just define this as the default value for the filter. The fact that it's based on the threshold is secondary / an implementation detail.

One way to make it clear would be to move out to its own line & variable, e.g.

```php
$default_allow_full_batch_sync = ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC );
```

facebook-commerce.php
haszari: Let's bump this up to 5k - that's approx a quarter of the 19k from the problematic site, and host/site can use the filter if they need a smaller threshold.

haszari: (And let's test it - how big is the option with 5000 products?)

includes/AJAX.php
```diff
@@ -214,6 +214,11 @@ public function admin_complete_order() {
 	 * @since 2.0.0
 	 */
 	public function sync_products() {
+		// Inhibit on-demand full batch-api sync if the store has large product count.
```
haszari, line 217: Let's tweak these comments, as this could be disabled based on various factors now that it's an opt-out hook.

E.g. `Allow opt-out of full batch-API sync, for example if store has a large number of products`.

includes/Handlers/Connection.php
```diff
@@ -301,7 +301,10 @@ public function handle_connect() {
 			$this->update_system_user_id( $system_user_id );
 			$this->update_installation_data();
 
-			facebook_for_woocommerce()->get_products_sync_handler()->create_or_update_all_products();
+			// Only trigger initial full batch-api sync if the store doesn't have large product count.
```
haszari, line 304: Likewise for this comment 👯

haszari (commented):
Looks great to me - I think we can close the other PRs and do some more testing & get feedback on this approach & then release 🎉 

Keen to get data on what the impact of the threshold is on the database option size - can you run some tests and add the info to this PR? E.g. for 500 products, 1000, 2500, 5000. 

Also please refresh the PR description to match the latest changes (or remind me – I can do tomorrow).

facebook-commerce.php
rcstr: Updated in 211dd13

haszari (commented):
This is looking splendid - thanks @rcstr . I've added some more comments and also updated the description. Thanks for proactively sorting out the changelog!

I'm still testing (in conjunction with a PR using the filter), but 🤞 this PR is ready to go.

haszari:
Update: I've reworked this so it just provides the filter. By default, no behaviour will change.

Explaining the rationale for this:

- We've only had reports of this from a specific hosting platform; we don't know that the issue is affecting merchants on other platforms.
- Full sync, especially after onboarding, is an important feature. If we silently break it for some (large) sites that could be confusing and cause more support requests.
- The hook allows all merchants, platforms and developers to adjust the behaviour as needed.

haszari:
Update: tweaked the error message for the AJAX full sync - now just says the sync was disabled via filter (may not be due to product count).

haszari:
I've tested this with and without a custom filter and confirmed it allows disabling of the two relevant full sync operations. I've added some more logging and tweaked the messages so it's clear that this was disabled via a customization.

Merging - and proceeding to release 2.6.1!

PR #2037, opened 2021-06-24
Disable full batch API sync on Atomic env

rcstr:
Fixes # 

Disables full batch API sync when the site is hosted on Atomic environment. We validate Atomic environment with the constant `IS_ATOMIC`

1. Edit your `wp-config.php` or create an `mu-plugin` file, and add:
```php
define( 'IS_ATOMIC', true );
``` 
2. Go to _Marketing > Facebook > Product sync_
3. Validate *Sync products* button is not rendered'
Screen Shot 2021-06-23 at 19 09 48

### Changelog entry

> Fix - Disable full batch API sync from Atomic sites

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -809,7 +809,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 	 * @return boolean True if full sync is safe.
 	 */
 	public function allow_full_batch_api_sync() {
-		return ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC );
+		return ( ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC ) && ! \WC_Facebookcommerce_Utils::is_Atomic_env() );
 	}
 
 
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -785,26 +785,14 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		return $product_item_ids;
 	}
 
-
 	/**
-	 * Gets the total of published products.
+	 * Gets the total number of published products.
 	 *
 	 * @return int
 	 */
 	public function get_product_count() {
-
-		$args = array(
-			'post_type'      => 'product',
-			'post_status'    => 'publish',
-			'posts_per_page' => -1,
-			'fields'         => 'ids',
-		);
-
-		$products = new WP_Query( $args );
-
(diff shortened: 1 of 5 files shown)


facebook-commerce.php
```diff
@@ -785,26 +790,26 @@ private function find_variation_product_item_ids( $product_group_id ) {
...
-		return $products->found_posts;
+	/**
+	 * Should full batch-API sync be allowed?
+	 *
+	 * Used to disable various full sync UI/APIs to avoid performance impact.
+	 *
+	 * @return boolean True if full sync is safe.
+	 */
+	public function allow_full_batch_api_sync() {
+		return ( ( $this->get_product_count() < self::MAX_PRODUCTS_FOR_FULL_SYNC ) && ! \WC_Facebookcommerce_Utils::is_Atomic_env() );
```
haszari, line 812: Ideally we'd expose a mechanism that any hosting platform (or site) could use, rather than hard-code this specifically for atomic platform. We can probably do that by exposing an option/filter, and working with Atomic platform team to deploy to all sites.

includes/Admin/Settings_Screens/Product_Sync.php
```diff
@@ -174,14 +174,17 @@ class="button button-large button-primary"
...
+		$show_sync_button = $integration->allow_full_batch_api_sync() &&
+			facebook_for_woocommerce()->get_connection_handler()->is_connected();
 		?>
 
 		<h2>
 
 			<?php esc_html_e( 'Product sync', 'facebook-for-woocommerce' ); ?>
 
-			<?php if ( facebook_for_woocommerce()->get_connection_handler()->is_connected() ) : ?>
+			<?php if ( $show_sync_button ) : ?>
```
haszari, line 187: I'm not sure if we should hide the button - might be confusing. Probably can leave it there and show the error message from `sync_products()`, so admin understands why it's disabled.

haszari (commented):
Looks good overall – I think we should combine all changes into one PR, and tweak so the same opt-out mechanism is available to all.

Ideally I'd prefer that atomic manages the code to opt-out, but happy to include specific support in the plugin if that's where we end up.

haszari:
Marking as blocked at least until we've discussed. These are the considerations:

- Hide the button vs. show button and error message if sync is disabled.
- Implement this specifically for Atomic platform vs. expose a mechanism for anyone to use + work with Atomic to activate.

haszari:
Closing in favour of #2033

PR #2046, opened 2021-07-07
Fix the file path when doing the is_readable check.

matt-h:
Fixes #1642

- [x] Do the changed files pass `phpcs` checks? Please remove `phpcs:ignore` comments in changed files and fix any issues, or delete if not practical.

### Changes proposed in this Pull Request:

This concatenates the path to be the same as used in the `load_class` method. This prevents the `is_readable` check from attempting to check in the root path of the operating system for the files and instead looks in the correct location of the plugin. Looking for `/includes/fbwpml.php` and `/includes/Integrations/Bookings.php`

In our production environment these failed checks were adding 3 seconds to the PHP execution time with PHP 7.4.21.

### How to test the changes in this Pull Request:

1. It is a simple fix.

### Changelog entry

> Fix - Fix the file path when doing the is_readable check for integrations.

PR #2071, opened 2021-09-03
Replace outdated `Integrations` class loader with simple require statements

### LINKED ISSUE #1642: is_readable(): open_basedir restriction in effect
## 🔎 Isolate the bug

- [X ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.

## ✍️ Describe the bug

PHP Warning:  is_readable(): open_basedir restriction in effect. File(/includes/fbwpml.php) is not within the allowed path(s): (D:\www\customDomain\) 

##🚶‍♀️ Steps to reproduce

Windows Server with PHP 7.3 having open_basedir setup will create the issue with logging lots of errors in debug.log

## ✔️ Expected behavior

while checking the file using is_readable() provide absolute path instead of relative path.

## WordPress (please complete the following information):
Core version: 5.5.1
Gutenberg version (if installed): --
WooCommerce version: 4.6.1
Facebook for WooCommerce version: 2.0.5
Site language: en-GB
Any other plugins installed:

## 🗃 Logs

PHP Warning:  is_readable(): open_basedir restriction in effect. File(/includes/fbwpml.php) is not within the allowed path(s): (D:\www\customDomain\) in D:\www\customDomain\public_html\wp-content\plugins\facebook-for-woocommerce\includes\Integrations\Integrations.php on line 61

PHP Warning:  is_readable(): open_basedir restriction in effect. File(/includes/Integrations/Bookings.php) is not within the allowed path(s): (D:\www\customDomain\) in D:\www\customDomain\public_html\wp-content\plugins\facebook-for-woocommerce\includes\Integrations\Integrations.php on line 61

danielbitzer:
Fixes #1642

The `Integrations` class uses a fairly outdated and discouraged method of class loading. I.e. returning instantiated class objects from file include statements and checking that a source-controller class file `is_readable()` before requiring it is unnecessary. It also has issues like #1642. The `Integrations` class doesn't actually do anything apart from loading Bookings and WPML classes so I suggest we remove it and use a simple require statement or better yet use psr-4 autoloading.

I'm fairly sure this PR will also re-enable the "Bookings integration" which has likely not been working since we switched to using composer autoloading. The reason behind this is that `class_exists` will return `true` is the class can be autoloaded which would have stopped the class actually being loaded by this `Integrations` class.

Personally, I added `var_dump` statements to `fbwpml.php` and `Integrations/Bookings.php` to check that they are now being loaded.

### Changelog entry

> Fix - Incorrect `is_readable()` usage when loading Integration classes


jconroy:
When this is being reviewed can someone look at https://github.com/woocommerce/facebook-for-woocommerce/pull/2046 at the same time

class-wc-facebookcommerce.php
```diff
@@ -164,7 +162,10 @@ public function init() {
 					$this->ajax = new \SkyVerge\WooCommerce\Facebook\AJAX();
 				}
 
-				$this->integrations = new \SkyVerge\WooCommerce\Facebook\Integrations\Integrations( $this );
+				// Load integrations
```
budzanowski, line 165: Full stop at the end of the comment.

budzanowski (approved):
One lousy nitpick left. This looks great!

budzanowski:
I have closed two other PRs #2046 and #2043 that have been addressing the same issue.

danielbitzer:
@budzanowski thanks for the review and for closing those other PRs!

dalepgrant:
@danielbitzer Thanks for the fix (and @budzanowski thanks too) 🙏

PR #2085, opened 2021-09-24
Update logs formatting for connect server logs.

budzanowski:
The intent of this PR  is ( part of the effort ) to have more comprehensive logs from connect server. The plugin only displays the logs that are passed from the connect server so this PR focuses only on the formatting of said logs. All the important changes are happening inside the connect server PR https://github.com/Automattic/connect.woocommerce.com/pull/137 which is needed to properly test this one.

One additional change ( I should be a separate PR I know ) is the addition of `CONNECTION_AUTHENTICATION_URL` const and the fetcher for this const. It is merely for making debugging easier. The added filter allows redirecting the authentication to the local connect server instance.

 - Enable the plugin using this PR
 - follow the instruction inside the https://github.com/Automattic/connect.woocommerce.com/pull/137 

Example output in the facebook_for_woocommerce_connect logs in Status->Logs:

```
09-23-2021 @ 07:20:24 - Failed to connect to Facebook. Reason: {
    "message": "Application does not have permission for this action",
    "type": "OAuthException",
    "code": 10,
    "error_subcode": 2859015,
    "is_transient": false,
    "error_user_title": "Your account is restricted right now",
    "error_user_msg": "You have been temporarily blocked from performing this action.",
    "fbtrace_id": "AgIrd2kMFugwcCqVAyH-HDX"
}
```

The actual message depends on the testing scenario.

### Changelog entry

> New - Better logging format for connections issues.

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -280,7 +280,7 @@ class Connection {
 
 			$is_error                 = ! empty( $_GET['err'] ) ? true : false;
 			$error_code               = ! empty( $_GET['err_code'] ) ? stripslashes( sanitize_text_field( $_GET['err_code'] ) ) : '';
-			$merchant_access_token    = ! empty( $_GET['merchant_access_token'] ) ?  sanitize_text_field( $_GET['merchant_access_token'] ) : '';
+			$merchant_access_token    = ! empty( $_GET['merchant_access_token'] ) ? sanitize_text_field( $_GET['merchant_access_token'] ) : '';
 			$system_user_access_token = ! empty( $_GET['system_user_access_token'] ) ? sanitize_text_field( $_GET['system_user_access_token'] ) : '';
 			$system_user_id           = ! empty( $_GET['system_user_id'] ) ? sanitize_text_field( $_GET['system_user_id'] ) : '';
 
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -334,7 +334,7 @@ class Connection {
 		} catch ( Connect_WC_API_Exception $exception ) {
 			$message = $this->prepare_connect_server_message_for_user_display( $exception->getMessage() );
 
-			facebook_for_woocommerce()->log( sprintf( 'Failed to connect to Facebook. Reason: %s', $message), 'facebook_for_woocommerce_connect' );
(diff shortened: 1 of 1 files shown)


includes/Handlers/Connection.php
```diff
@@ -545,7 +571,7 @@ public function get_connect_url( $connect_commerce = false ) {
 	 * include the final site URL, which is where the merchant will redirect to with the data that needs to be stored.
 	 * So the final URL looks like this without encoding:
 	 *
-	 * https://www.facebook.com/commerce_manager/onboarding/?app_id={id}&redirect_url=https://connect.woocommerce.com/auth/facebook/?site_url=https://example.com/?wc-api=wc_facebook_connect_commerce&nonce=1234
+	 * https://www.facebook.com/commerce_manager/onboarding/?app_id={id}&redirect_url=https://connect.woocommerce.local/auth/facebook/?site_url=https://example.com/?wc-api=wc_facebook_connect_commerce&nonce=1234
```
layoutd, line 574: What's the reason for this change?

includes/Handlers/Connection.php
```diff
@@ -848,6 +874,24 @@ public function get_app_store_login_url() {
...
 
+	/**
+	 * Gets connect server authentication url.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string URL
+	 */
+	public function get_connection_authentication_url() {
+
```
layoutd, line 885:
```suggestion
```

includes/Handlers/Connection.php
```diff
@@ -339,6 +343,28 @@ public function handle_connect() {
...
+	 */
+	public function prepare_connect_server_message_for_user_display( $message ) {
+			/*
+			 * In some scenarios the connect server message is a JSON encoded object.
+			 * This happens when we have detailed information from Facebook API endpoint about the error.
+			 * We want to print it pretty for the customers.
+			 */
+			$decoded_message = json_decode( $message );
+			if ( json_last_error() === JSON_ERROR_NONE ) {
+				// If error is the fist key we want to use just the body to simplify the message.
```
layoutd, line 361:
```suggestion
				// If error is the first key we want to use just the body to simplify the message.
```

layoutd (approved):
LGTM and works with the accompanying connect.wc changes.

Two small changes and one clarification in case something snuck in there.

includes/Handlers/Connection.php
budzanowski: Leftover from testing.

PR #2098, opened 2021-10-11
Feed code cleanup.

budzanowski:
## Changes proposed in this Pull Request:

Cleans up the feed generation, making it ready for further refactoring. This PR basically does 3 things.

### 1. Feed code not used in any way is removed.

- `WC_Facebookcommerce_Integration::handle_generate_product_catalog_feed` old feed handling entry points. Not used anymore.
-  `WC_Facebook_Product_Feed::schedule_feed_generation` not used, we are using  `Feed::schedule_feed_generation` for this.
- Feed generation time calculation functions: `set_feed_generation_time_with_decay`, `set_average_feed_generation_time`, `get_estimated_feed_generation_time`, `estimate_generation_time`, `get_average_feed_generation_time`, `get_feed_generation_estimate_sample_size` and `get_feed_generation_buffer_time`.

There are other things that are not used all over the place. I just wanted to declutter `includes/fbproductfeed.php`. 

### 2. Restructure the feed generation code in `includes/fbproductfeed.php`

The problem with the current feed generator ( out of memory issues ) is that it tries to process all of the products at once. This stops working with even fairly small catalogues, but how fast it does depends on the particular server configuration. What we want to do is to be able to process the products in bunches.
The processing happens in three basic steps. First is the preparation of an empty temporary feed file. The second step is processing products and writing output into the temp feed file. The third and last step is renaming the temp file into the final feed file. Those three steps have been written as one method. Now they have been extracted into distinct steps:

- `prepare_temporary_feed_file`
- `write_products_feed_to_temp_file`
- `rename_temporary_feed_file_to_final_feed_file`

This will allow us to trigger `write_products_feed_to_temp_file` as many times as we want with a reduced set of products ensuring that memory is not exhausted. This process will be orchestrated by `woocommerce/action-scheduler-job-framework` and it will be implemented in #2099 

### 3. Update the `woocommerce/action-scheduler-job-framework` to `2.0.0` and replace single item processing with the processing of the whole packet.

This code is not connected yet so the changes are not accurate. I just wanted to push a correct set of methods matching `2.0.0`.

This PR should not change anything. The feed generation should still work as it did ( failing for a big catalogue )

### Changelog entry

-  Dev - Cleanup Feed generation code.


includes/fbproductfeed.php
```diff
@@ -518,96 +326,133 @@ private function create_files_to_protect_product_feed_directory() {
...
+				);
 
-				WC_Facebookcommerce_Utils::log( json_encode( $e->getMessage() ) );
+				if ( ! empty( $temp_feed_file ) ) {
+					fwrite( $temp_feed_file, $product_data_as_feed_row );
+				}
+			}
 
-				$written = false;
+			wp_reset_postdata();
```
nima-karimi, line 430: What does this do here? Could you maybe add an inline comment describing why it's needed?

nima-karimi (approved):
LGTM 👍 I left one minor comment.

Issue #2008, opened 2021-06-09
Facebook and Instagram Prices


famuser:
## 🔎 Isolate the bug
X

- [x] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
 The prices sometimes are shown in the facebook and instagram shopping without tax. It varies from products time to time and would go back to with tax randomly.

## 🚶‍♀️ Steps to reproduce
 updated website - issues still there.

1. The first step
1. The next step… 

## ✔️ Expected behavior
 prices should display correctly with tax

## 🗃 Logs
 

<details>
	 www.freckledfrog.com.au

Version 2.5.1 for Facebook for woocommerce, Version 5.3.0 for woocommerce, version 5.7.2 wordpress, version 7.4 php
</details>

haszari:
Thanks for reporting this @famuser ! Can you give us more details of how we can reproduce the bug?

famuser:
> Thanks for reporting this @famuser ! Can you give us more details of how we can reproduce the bug?

We have had your facebook plugin for many years now and our shop feature has been on our account for a while too. However we have not changed anything differently that would impact any changes to the prices excluding tax, but we have recently started experiencing this issue where the price would suddenly change to ex tax and it would be for some products but not all. Just over a week ago 80% of the prices changed to ex tax I had to go into woo commerce and refresh all the products that changed and it went back to normal price. But now there are about 15% of our products that are once again slowly showing up as ex tax on the instagram/facebook shop.

famuser:
> Thanks for reporting this @famuser ! Can you give us more details of how we can reproduce the bug?

Any update on this one, please?

haszari:
> > Thanks for reporting this @famuser ! Can you give us more details of how we can reproduce the bug?

> Any update on this one, please? @haszari

Thanks for providing more info about how this is impacting your store. To find out what's happening we need steps to reproduce the problem reliably on a basic store setup running just WooCommerce and Facebook for WooCommerce.

famuser:
@haszari 
![1623807647175](https://user-images.githubusercontent.com/85598004/122145793-7f353680-ce88-11eb-9b02-d7a8b9a4387d.JPEG)

its so random it changes every day the ones are ive circled are prices ex tax
[...]

PR #2099, opened 2021-10-11
OOM issues resistant feed generator.

### LINKED ISSUE #1818: PHP Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 5242880 bytes) in /wordpress/core/5.7/wp-includes/class-wp-query.php on line 2986
## 🔎 Isolate the bug

## ✍️ Describe the bug

wp-admin fails 

PHP Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 5242880 bytes) in /wordpress/core/5.7/wp-includes/class-wp-query.php on line 2986  

##🚶‍♀️ Steps to reproduce

It happened in #3844188-zen  even with only these plugins:

akismet
coblocks
health-check
jetpack
woocommerce
full-site-editing

## ✔️ Expected behavior

wp-admin loads without error

## 🗃 Logs

SSR uploaded in https://github.com/woocommerce/woocommerce-facebook/issues/8

### LINKED ISSUE #1841: Improve performance of full product sync with large number of products
We've had reports (#1839) of full product sync (triggered by 2.3.3 upgrade - #1808) causing issues, especially on sites with large product catalogue.

The goal of this issue is to improve the performance, scalability and robustness of full product sync. For example, splitting larger updates into smaller batches, and using Action Scheduler to manage these jobs efficiently.

### LINKED ISSUE #1871: Out of Memory issues since 2.3.2

budzanowski:

This PR implements a feed generator that uses `woocommerce/action-scheduler-job-framework` as a backend. This allows us to process the products in packets small enough to prevent OOM issues. The changes are based on the code from https://github.com/woocommerce/facebook-for-woocommerce/pull/2098

I wanted to go back to work we did in https://github.com/woocommerce/facebook-for-woocommerce/pull/1989 but given the age and the scope of changes, I doubt that we could ever release it. I have decided to go with a much leaner approach. 

This PR and #2008 answer the question: What is the simplest way to have the current feed generator without OOM issues, no other improvements necessary. It turns out that the changes are rather simple if we utilize `action-scheduler-job-framework`

Since this is an experimental feature ( I feel confident about it but let's take it slow ). I have added an option that allows this new type of feed generation:

image

### wc_get_products vs wc_get_product
The generated feed file is a bit different. The description field is different. It turns out that `wc_get_product` that is used by the old feed does not fetch all of the meta fields so `WC_Facebook_Product` was not detecting the correct description. We now use `wc_get_products` and we have all of the necessary information. After short debugging, it looks like this behaviour is correct because it matches the output of the `Batch API` update procedure. The usage of `wc_get_products` has triggered a necessary change in `WC_Facebook_Product` - it can be now also initialized from `WC_Product` and not just id. 

Fixes #1818
Fixes #1841
Fixes #1871

1. Flip the feature switch.
2. Go to Scheduled Actions and run `wc_facebook_regenerate_feed`
3. Wait until the process is completed ( no actions lilke: `facebook_for_woocommerce/jobs/generate_feed/chain_start` `facebook_for_woocommerce/jobs/generate_feed/chain_batch` in pending or in progress.
4. Check uploads/facebook-for-woocommerce folder for the feed CSV file. It should have all of your products.

### Changelog entry

- New - Memory improved feed generation feature.

### DIFF
--- a/includes/Admin/Settings_Screens/Connection.php
+++ b/includes/Admin/Settings_Screens/Connection.php
@@ -334,10 +334,10 @@ class Connection extends Admin\Abstract_Settings_Screen {
 
 			array(
 				'id'       => \WC_Facebookcommerce_Integration::SETTING_ENABLE_NEW_STYLE_FEED_GENERATOR,
-				'title'    => __( 'Enable new style feed generation', 'facebook-for-woocommerce' ),
+				'title'    => __( 'Experimental! Enable new style feed generation', 'facebook-for-woocommerce' ),
 				'type'     => 'checkbox',
 				'desc'     => __( 'Use new, memory improved, feed generation process.', 'facebook-for-woocommerce' ),
-				'desc_tip' => __( 'Experimental feature. Only enable this if you are experiencing problems with feed generation.', 'facebook-for-woocommerce' ),
+				'desc_tip' => __( 'Experimental feature. Only enable this if you are experiencing problems with feed generation. This is an experimental feature in testing phase.', 'facebook-for-woocommerce' ),
 				'default'  => 'no',
 			),
 


nima-karimi (approved):
LGTM 💯

PR #2104, opened 2021-10-25
Track batched feed generation time.

budzanowski:

Add tracking of batched feed generation time.

Because the new feed generator generates the feed-in batches we need to modify how we are collecting elapsed time. This PR adds an accumulator value that sums all of the batches generation time for one feed generation:
`TRANSIENT_WCTRACKER_FEED_GENERATION_BATCH_TIME`
This is later used to populate ( when the generation process ends ) the
`TRANSIENT_WCTRACKER_FEED_GENERATION_TIME` This is the transient that is used ( also for the old feed generator ) by the tracking code to colled the feed generation time.
Collecting just this value is not enough for the new tracking.  In addition to calculating how much time it took to generate the feed batches we also need to collect how long the whole process took, start to finish, including the time between the batches. 
To achieve this we are using `TRANSIENT_WCTRACKER_FEED_GENERATION_BATCH_WALL_TIME` and `TRANSIENT_WCTRACKER_FEED_GENERATION_BATCH_WALL_START_TIME`. They are later used by the tracker code to populate the:
```PHP
['extensions']['facebook-for-woocommerce']['feed-generation-wall-time']
```
property.

1. Trigger the feed generation with: `wc_facebook_regenerate_feed` action - should be in pending.
2. Using wp wc tracker snapshot --format=json check if the new field `feed-generation-wall-time` is picked up in the snapshot - it should be the same ( minus fractional part ) as `feed-generation-time`.
3. Change the setting and enable 'Experimental! Enable new style...`
image
4. Trigger the feed generation again.
5. Check again the snapshot data to see if the value has changed. Depending on the number of products ( the more the better ) the `feed-generation-wall-time` will be bigger and bigger compared to `feed-generation-time`.

### Changelog entry

Dev - Collect feed generation wall time in tracking.

### DIFF
--- a/includes/Jobs/GenerateProductFeed.php
+++ b/includes/Jobs/GenerateProductFeed.php
@@ -26,6 +26,7 @@ class GenerateProductFeed extends AbstractChainedJob {
 		$feed_handler = new \WC_Facebook_Product_Feed();
 		$feed_handler->create_files_to_protect_product_feed_directory();
 		$feed_handler->prepare_temporary_feed_file();
+		facebook_for_woocommerce()->get_tracker()->reset_batch_generation_time();
 	}
 
 	/**
@@ -34,6 +35,7 @@ class GenerateProductFeed extends AbstractChainedJob {
 	protected function handle_end() {
 		$feed_handler = new \WC_Facebook_Product_Feed();
 		$feed_handler->rename_temporary_feed_file_to_final_feed_file();
+		facebook_for_woocommerce()->get_tracker()->save_batch_generation_time();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


nima-karimi (commented):
I started reviewing this but for some reason, I can't get the new feed generator to work on my machine. It keeps scheduling jobs with empty arguments. As a result, I can't test this PR properly yet. I get the same value for `feed-generation-wall-time` and `feed-generation-time` with the new experimental feed generator, which means something's not working as expected (at least on my machine).

I'll give this another go but please let me if this is a known issue. I've tried reinstalling composer but it didn't resolve the issue for me.

nima-karimi (approved):
Looking great 💯 I tested it with different sets of products and got the expected results (more products = more difference between the two tracked times).

PR #2109, opened 2021-11-01
AddToCart value should be product price

mikaelz:
AddToCart value should be product price instead cart total amount


rawdreeg:
Thanks for submitting this @mikaelz ... the issue has been resoved by #2215. I will close this PR.

PR #2118, opened 2021-11-09
Fix typo in message retrival.

budzanowski:
The code was fetching non-existing value for the message.

### Changelog entry

> Fix - Fetch correct message for api errors.


danielbitzer (approved):
Looks good 👍

PR #2119, opened 2021-11-09
Prevent attempts to delete non existing sets.

budzanowski:
When a new set gets created using categories that do not represent any products the code was detecting this empty set and interpreted it as an existing set that no longer has products. Because a set that has no products is pointless the code was trying to delete it. But in this case, the set was never created so the code was doing a delete request with an empty set id that was triggering the following error response:

```
11-09-2021 @ 13:35:11 - Response
code: 400
message: Bad Request
body: {"error":{"message":"Unsupported delete request. Please read the Graph API documentation at https://developers.facebook.com/docs/graph-api","type":"GraphMethodException","code":100,"error_subcode":33,"fbtrace_id":"APuNGdfgsdfgsfifbM0RyliCu"}}
```

1. Go to Products > FB sets
2. Create a FB set using a category that is not used by any products
3. Check logs ( remember to enable logging in the plugin settings ) the error message should not be there.

### Changelog entry

> Fix - No empty set delete graph API calls for new empty sets.


KICCI36 (changes requested):
> `[58781085]() #
Duplicate of #`

PR #2121, opened 2021-11-12
Account for parent product description when validating the variation products.

### LINKED ISSUE #2120: Product description not synced to Facebook in Variable products when attribute uses all caps
## 🔎 Isolate the bug

- [X] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug

If the attribute title uses all caps (SIZE: M), this causes the sync to fail as Facebook doesn't allow only uppercase in any field. Facebook seems to see the variation attributes as the only description on the variation. 

>Product description is all capital letters. Please change the description to lowercase in order to allow your product to be synchronized.

Once a description is added in the variation Description or Facebook Description, the variation can be synced. It seems like the product description and short description are ignored. 

## 🚶‍♀️ Steps to reproduce

1. Create a variable product
1. Use ALL CAPS in the attribute name and value (SIZE, XS | S | M | L )
2. Do not complete Description or Facebook Description in any variant
3. Publish the product
3. See an error in Facebook Product Sync that Product description is all capital letters.
4. Add a description to a variation and update the product. The variation has synced. https://d.pr/i/NlqwTM

Additional steps:

1. Change the attribute name to regular case (SIZE to Size)
2. Update the product
3. See that all variations have synced 

## ✔️ Expected behavior

I expected that the main product description/short description would also be used for variable products like it is for simple products. 

4448916-zen

budzanowski:
Fixes #2120 

1. Use testing instruction from #2120 
2. Product should sync
### Changelog entry

> Fix - Take parent product description into account when validatin the variable product.

### DIFF
--- a/includes/ProductSync/ProductValidator.php
+++ b/includes/ProductSync/ProductValidator.php
@@ -70,23 +70,33 @@ class ProductValidator {
 	protected $product_parent;
 
 	/**
+	 * The product parent object if the product has a parent.
+	 *
+	 * @var WC_Facebook_Product
+	 */
+	protected $fb_product_parent;
+
+	/**
 	 * ProductValidator constructor.
 	 *
 	 * @param WC_Facebookcommerce_Integration $integration The FB integration instance.
 	 * @param WC_Product                      $product     The product to validate. Accepts both variations and variable products.
 	 */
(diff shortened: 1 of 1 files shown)


layoutd (approved):
👍🏻  LGTM

PR #2183, opened 2022-04-06
Fix 2166: Use Automattic\WooCommerce\Admin\Features\Features::is_enabled.

### LINKED ISSUE #2166: WooCommerce\Admin\Loader::is_feature_enabled is deprecated
## 🔎 Isolate the bug

## ✍️ Describe the bug
Facebook for WooCommerce still makes use of the Loader::is_feature_enabled within WooCommerce Admin, this function has been deprecated ([for some time now](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Loader.php#L217)). Please use `Automattic\WooCommerce\Admin\Features\Features::is_enabled( $feature )` instead.
There seems to be two places, see this search:
https://github.com/woocommerce/facebook-for-woocommerce/search?q=Loader%3A%3Ais_feature_enabled

## ✔️ Expected behavior
Use `Automattic\WooCommerce\Admin\Features\Features::is_enabled( $feature )` instead.

rawdreeg:

Facebook for WooCommerce still makes use of the Loader::is_feature_enabled within WooCommerce Admin, this function has been [deprecated](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Loader.php#L217).

Closes #2166.

This PR introduces the usage of `Automattic\WooCommerce\Admin\Features\Features::is_enabled` instead of the deprecated `WooCommerce\Admin\Loader::is_feature_enabled`.

1. This method is used to correctly put the Facebook admin menu in the Marketing tab and warn the user that the Fb menu has moved. The Facebook menu should still be in the marketing tab after this change with no issues or warnings.

Leave the "Changelog entry" header in place completely empty, without any summary if no changelog entry is needed.
If you remove the "Changelog entry" header, the title of Pull Request will be used as the changelog entry.
-->
### Changelog entry

> Tweak - Use `Automattic\WooCommerce\Admin\Features\Features::is_enabled` instead of the deprecated `WooCommerce\Admin\Loader::is_feature_enabled`

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -17,6 +17,7 @@ use SkyVerge\WooCommerce\Facebook\Utilities\Background_Remove_Duplicate_Visibili
 use SkyVerge\WooCommerce\PluginFramework\v5_10_0 as Framework;
 use SkyVerge\WooCommerce\Facebook\ProductSync\ProductValidator as ProductSyncValidator;
 use SkyVerge\WooCommerce\Facebook\Utilities\Heartbeat;
+use Automattic\WooCommerce\Admin\Features\Features as WooAdminFeatures;
 
 if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
The code changes look good, although I was unable to test it fully.

I think the original issue is a bit misleading mentioning that it has been deprecated for a while. Maybe there was a delay with the merge to the monorepo, but this change hasn't been included in WC 6.4 yet (maybe landing in 6.5).

I wanted to also test Facebook showing up in the navigation, but since `class_exists( WooAdminFeatures::class )` always returns false `use_woo_nav` is never true. That would have seemed like a higher priority to fix.

Either way we are making it future proof (if there are no further changes), so I'll go ahead and approve this PR.

rawdreeg:
Thanks, @mikkamp , 

> Wanted to clarify that the deprecation warning isn't added yet in 6.4, but Features::is_enabled is already available.

The issue report made it sound a lot more imminent :-D. but yes, you're right this will make this future proof which a plus in my book :-D

PR #2185, opened 2022-04-08
Fix Facebook doesn't show up in the new navigation

### LINKED ISSUE #2184: Facebook doesn't show up in the new navigation
## 🔎 Isolate the bug

If we go to WooCommerce > Setttings > Advanced > Features and enable the Navigation option, Facebook doesn't show under Extensions. Instead it shows `FB Product Sets` which is part of the old menu.

image

## 🚶‍♀️ Steps to reproduce

1. Activate the Facebook plugin
2. Go to WooCommerce > Setttings > Advanced > Features and enable the Navigation option
3. View the new navigation menu and confirm Facebook doesn't show under Extensions

## ✔️ Expected behaviour

I tested by temporarily moving the line `$this->use_woo_nav = ...` to within the `add_menu_item` function and then it does appear.

image

So at first glance it looks like the code to detect the feature is called too early, not sure which init hook would be best to use instead tough.

rawdreeg:

If we go to WooCommerce > Settings > Advanced > Features and enable the Navigation option, Facebook doesn't show under Extensions. Instead, it shows FB Product Sets which is part of the old menu. This is because `$this->use_woo_nav ` is set too early. 

I am moving this initialization in the `add_menu_item` function, which runs on the `admin_menu` action hook.

Closes #2184.

### Screenshots:

Screenshot 2022-04-08 at 12 34 44

1. Activate the Facebook plugin
2. Go to WooCommerce > Settings > Advanced > Features and enable the Navigation option
3. View the new navigation menu and confirm Facebook is displayed under Extensions

### Changelog entry

> Fix - Issue with Facebook not displayed in the new WC navigation

### DIFF
--- a/includes/Admin/Settings.php
+++ b/includes/Admin/Settings.php
@@ -60,7 +60,7 @@ class Settings {
 
 		add_action( 'wp_loaded', array( $this, 'save' ) );
 
-		$this->use_woo_nav = class_exists( WooAdminFeatures::class ) && class_exists( WooAdminMenu::class ) && WooAdminFeatures::is_enabled( 'navigation' );
+		$this->use_woo_nav = false;
 	}
 
 
@@ -73,6 +73,7 @@ class Settings {
 
 		$root_menu_item       = 'woocommerce';
 		$is_marketing_enabled = false;
+		$this->use_woo_nav = class_exists( WooAdminFeatures::class ) && class_exists( WooAdminMenu::class ) && WooAdminFeatures::is_enabled( 'navigation' );
 
 		if ( Framework\SV_WC_Plugin_Compatibility::is_enhanced_admin_available() ) {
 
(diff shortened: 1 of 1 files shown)


includes/Admin/Settings.php
```diff
@@ -73,6 +72,7 @@ public function add_menu_item() {
 
 		$root_menu_item       = 'woocommerce';
 		$is_marketing_enabled = false;
+		$this->use_woo_nav = class_exists( WooAdminFeatures::class ) && class_exists( WooAdminMenu::class ) && WooAdminFeatures::is_enabled( 'navigation' );
```
mikkamp, line 75: Minor: even though `phpcs:ignore` is set for this file, we probably want to align the equals signs anyways.

mikkamp (approved):
Thanks for the change, it's working as expected. I see Facebook appear in the new nav menu now. And the extra nav on the Facebook page disappears when the new nav is enabled.

Just left a small note about alignment, since eventually we want to fix all the phpcs issues.

includes/Admin/Settings.php
rawdreeg: Thanks for looking at this. I totally missed this. I have corrected the alignment on the variable assignment.

PR #2194, opened 2022-04-26
Fix product local non-latin attribute names sync into Facebook

### LINKED ISSUE #2130: Imported variation description is not properly escaped
## 🔎 Isolate the bug

- [ ] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.

## ✍️ Describe the bug

Variation descriptions get escaped incorrectly when the site is using `utf-8` encoding and uses Hebrew characters.

image

## 🚶‍♀️ Steps to reproduce

@gabrielbonifacio Please correct me if I'm wrong

1. Setup a store with `utf-8` encoding
1. Create a product variation with a description containing  Hebrew characters, like `מס`
2. Sync the products with FB

## ✔️ Expected behavior

Synced products should contain the provided description, like `מס`. **Not** the HTML entities `%d5%9e`

## 🗃 Logs

<details>
	
</details>

message-dimke:
When you have local product attribute names to be all non-latin characters and you sync that to Facebook, you get corrupted attribute names.

Properly prepare attribute names before the sync by taking original attribute names from parent product instead of using adjusted meta_key values from the database as attribute name.

Closes #2130.

### Screenshots:

Before:

_4__Facebook

After:

_4__Facebook

1. Create variable product.
2. Add Custom product attribute in a language which has no latin characters in its alphabet (on the screen below I added two of such attributes).

Notification_Center

3. Create Variations with those attributes.

Cursor_and_Edit_product_“V-Neck_T-Shirt_UTF8_Compatible_12345678910111213141516”_‹_WordPress-Facebook_—_WordPress

4. Sync product to Facebook.

Edit_product_“V-Neck_T-Shirt_UTF8_Compatible_12345678910111213141516”_‹_WordPress-Facebook_—_WordPress

5. Visit product page on Facebook to see the results. You should see attribute names are readable in their corresponding languages instead of a mix of numbers and percent characters (like on the images under Screenshots section of this PR)

### Changelog entry

> Fix - Non-latin custom product attribute names sync

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -814,9 +814,8 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 							if ( ! isset( $product_data['custom_data'] ) ) {
 								$product_data['custom_data'] = array();
 							}
-
+							$new_name                                 = wc_attribute_label( $new_name, $product );
 							$product_data['custom_data'][ $new_name ] = urldecode( $option_values[0] );
-
 							break;
 					}//end switch
 				} else {


rawdreeg (approved):
LGTM!

✅  I can confirm attribute names are displayed correctly.

Thanks for the fix, @message-dimke

PR #2195, opened 2022-04-26
Fix product custom attribute value with semicolon breaks sync.

### LINKED ISSUE #2063: Disallowed characters in Custom Product Attribute causes error and sync failure
## 🔎 Isolate the bug

## ✍️ Describe the bug

If a disallowed character such as `:` is sent in the `additional_variant_attribute` properties, the associated variation will fail to sync, and an error will be thrown:

`The additional_variant_attribute information under  is invalid.`

## 🚶‍♀️ Steps to reproduce

In my observation, the extension excludes these kinds of characters from the `additional_variant_attribute` in most cases, which makes reproducing difficult. See p74xnj-uJ-p2#comment-1536

1. Create a variable product with an attribute that has the following values: `XS: 80/86 cm, S: 92/98 cm, M: 104/110 cm, L: 116/122 cm`
2. Create variations from all attributes
3. See if the attribute's values were sent in the `additional_variant_attribute` properties, such as:

> "additional_variant_attribute":"cor:Bege,tamanho:L: 116\/122 cm"

## ✔️ Expected behavior

The `additional_variant_attribute` properties only sends valid characters

message-dimke:
When variable product has custom attribute with semicolon in its values, product sync for Facebook fails.

Introduce a filter for custom attribute value to remove semicolon and allow to redefine such a behaviour.

Closes #2063.

1. Create variable product with at least one custom attribute which has value with semicolon in it.

Edit_product_“V-Neck_T-Shirt_UTF8_Compatible”_‹_WordPress-Facebook_—_WordPress

2. Sync product to Facebook and see product was not synced, but WooCommerce > Status > Logs has Facebook-for-woocommerce log with records about sync failure.

Cursor_and_WooCommerce_status_‹_WordPress-Facebook_—_WordPress

3. Pull the new code and repeat the procedure. You should now see the product appear to Facebook with no semicolon in attribute value and no errors in corresponding logs appear.

_4__Facebook

### Changelog entry

> Fix - remove semicolon from custom attribute value

### DIFF
--- a/includes/Products/Sync/Background.php
+++ b/includes/Products/Sync/Background.php
@@ -312,9 +312,30 @@ class Background extends Framework\SV_WP_Background_Job_Handler {
 				 * @param string $value Attribute value.
 				 * @return string Return the desired replacement string.
 				 */
-				$attribute_value = str_replace(
+				$val = str_replace(
 					',',
 					apply_filters( 'facebook_for_woocommerce_variant_attribute_comma_replacement', ' ', $val ),
+					$val
+				);
+				/**
+				 * Filter: facebook_for_woocommerce_variant_attribute_semicolon_replacement
+				 *
+				 * The Facebook API expects a comma-separated list of attributes in `additional_variant_attribute` field.
+				 * When encoded into JSON, attribute value with semicolons will break the format.
+				 * URL encoded semicolon version will remain so (%3A) at Facebook side after the sync.
+				 * This means that WooCommerce product attributes values should avoid the comma (`:`) character.
+				 * Facebook for WooCommerce replaces any `:` with a space by default.
+				 * This filter allows a site to provide a different replacement string.
+				 *
+				 * @since x.x.x
+				 *
+				 * @param string $replacement The default replacement string (`:`).
+				 * @param string $value Attribute value.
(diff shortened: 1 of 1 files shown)


rawdreeg (changes requested):
Thanks for the fix @message-dimke.

The logic works well and fixes the error.  However, the symbol replaced is a colon (:) and not a semicolon (;). Maybe we should rename the filter to  `facebook_for_woocommerce_variant_attribute_colon_replacement`-- although that [reads a bit funny](https://en.wikipedia.org/wiki/Colectomy) 😆 .

message-dimke:
Hey, @rawdreeg. I removed the filter for colon since colon can be removed by comma filter as well so there is no need in an extra filter for that character. Instead I forced removing , and : characters from the value before the function return statement since it possible to overwrite , with , or : with :, but we still must make sure those characters are replaced before Facebook sync.

rawdreeg (approved):
🚀  LGTM!

johndcoy:
The filter should include `attribute_name` as well. It currently only filters `attribute_value`

johndcoy:
@rawdreeg @message-dimke The filter should include `attribute_name` as well. It currently only filters `attribute_value`

PR #2204, opened 2022-05-12
Fix sales start time offset issue.

### LINKED ISSUE #1755: Product Time Synchronisation problem in WooCommerce and Facebook Catalog
## 🔎 Isolate the bug

## ✍️ Describe the bug

##🚶‍♀️ Steps to reproduce

1.For a variable product, In WooCommerce, put a start date 21/01/2021 , you will find inside the Facebook catalogue the day before 20/01/2021

https://1drv.ms/u/s!Ak_IFdTmvN05gi4U6pDaZX2Z-Y0m?e=bRMPW2
https://1drv.ms/u/s!Ak_IFdTmvN05gjAAhuPc-zQc-AH9?e=iJdd8d

rawdreeg:

WooCommerce normalizes the timestamp value for the sales_start_date and sales_post_date values before saving them in the post_meta table. By simply getting the _sale_price_dates_from post meta, we receive a timestamp with the timezone subtracted.  As a result, the sales start time sync to FB is incorrect (usually by plus or minus a day) for specific time zones. I tested with UTC+10h30.

To fix this issue, I am building the $sale_start and $sale_end strings with the offset timestamp instead of the raw value from the DB.

Closes #1755.

1. Change your store/site timezone to be further from UTC. E.g., UTC+10:30
2. Create/edit a product in WooCommerce, schedule a sales period for a variation item or a product

Screenshot 2022-05-12 at 17 20 54

Save and update.
3. Before this change you would have noticed after the changes have been synced with FB, the start date would be a day off. After this change, verify the product in your FB catalog, and the "Sale price start date" values should match the date set in WooCommerce.

Screenshot 2022-05-12 at 17 29 19

### Changelog entry

> Fix - Wrong sale price start date getting synced to  FB Catalog.

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -376,13 +376,13 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 			intval( round( $this->get_price_plus_tax( $sale_price ) * 100 ) );
 
 			$sale_start =
-			( $date     = get_post_meta( $this->id, '_sale_price_dates_from', true ) )
-			? date_i18n( 'Y-m-d', $date ) . self::MIN_TIME
+			( $date     = $this->woo_product->get_date_on_sale_from() )
+			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MIN_TIME
 			: self::MIN_DATE_1 . self::MIN_TIME;
 
 			$sale_end =
-			( $date   = get_post_meta( $this->id, '_sale_price_dates_to', true ) )
-			? date_i18n( 'Y-m-d', $date ) . self::MAX_TIME
+			( $date   = $this->woo_product->get_date_on_sale_to() )
+			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MAX_TIME
 			: self::MAX_DATE . self::MAX_TIME;
 
 			// check if sale is expired and sale time range is valid
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -377,12 +377,12 @@ if ( ! class_exists( 'WC_Facebook_Product' ) ) :
 
 			$sale_start =
 			( $date     = $this->woo_product->get_date_on_sale_from() )
-			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MIN_TIME
+			? date_i18n( WC_DateTime::ATOM, $date->getOffsetTimestamp() )
 			: self::MIN_DATE_1 . self::MIN_TIME;
 
 			$sale_end =
 			( $date   = $this->woo_product->get_date_on_sale_to() )
-			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MAX_TIME
+			? date_i18n( WC_DateTime::ATOM, $date->getOffsetTimestamp() )
 			: self::MAX_DATE . self::MAX_TIME;
 
 			// check if sale is expired and sale time range is valid


includes/fbproduct.php
```diff
@@ -376,13 +376,13 @@ public function add_sale_price( $product_data, $for_items_batch = false ) {
 			intval( round( $this->get_price_plus_tax( $sale_price ) * 100 ) );
 
 			$sale_start =
-			( $date     = get_post_meta( $this->id, '_sale_price_dates_from', true ) )
-			? date_i18n( 'Y-m-d', $date ) . self::MIN_TIME
+			( $date     = $this->woo_product->get_date_on_sale_from() )
+			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MIN_TIME
```
message-dimke, line 380: IMHO this change resets the local timezone and Facebook sale may start sooner or later than the original one depending on the timezone the store has. 

#### Example

I have my store in UTC+10:00 and I set Sale start `2022-05-15 00:00 UTC+10:00`. This is saved into database as `1652536800` timestamp, which is `2022-05-14 14:00:00 UTC+00:00`.

When posting to Facebook I get `1652536800` from the database and want to take a date `Y-m-d` out of it. And I get `2022-05-14`. Then I concatenate `T00:00+00:00` to it and send to Facebook. What I just did - I lost `14:00` hours. Jason had his Facebook Store set to `-08:00`, Facebook got our date `2022-05-14T00:00+00:00` and applied `-08:00` to it which resulted into `2022-05-13 16:00`. Which is kind of two days before we wanted it, right? 


#### Problem
The problem is not a timestamp we feed into `date_i18n`, but the concatenation ` . self::MIN/MAX_TIME`.

#### Solution
I would do it 

```suggestion
			? date_i18n( DATE_ATOM, $date->getOffsetTimestamp() )
```

or 

```suggestion
			? date_i18n( 'Y-m-d\TH:i:s+00:00', $date->getTimestamp() )
```

includes/fbproduct.php
```diff
@@ -376,13 +376,13 @@ public function add_sale_price( $product_data, $for_items_batch = false ) {
...
 			: self::MIN_DATE_1 . self::MIN_TIME;
 
 			$sale_end =
-			( $date   = get_post_meta( $this->id, '_sale_price_dates_to', true ) )
-			? date_i18n( 'Y-m-d', $date ) . self::MAX_TIME
+			( $date   = $this->woo_product->get_date_on_sale_to() )
+			? date_i18n( 'Y-m-d', $date->getOffsetTimestamp() ) . self::MAX_TIME
```
message-dimke, line 385:
```suggestion
			? date_i18n( DATE_ATOM, $date->getOffsetTimestamp() )
```

or 

```suggestion
			? date_i18n( 'Y-m-d\TH:i:s+00:00', $date->getTimestamp() )
```

message-dimke (changes requested):
Hey, @rawdreeg !

I added few suggestions to the code. Please, tell me what do you think?

includes/fbproduct.php
rawdreeg: Thanks, @message-dimke, you're assessment is 100% correct. I have applied your suggestion which works well. 

Thanks for taking the time to look at this. And I learned about the DATE_ATOM constant which is a bonus :-)

rawdreeg:
This ready for around review @message-dimke

includes/fbproduct.php
rawdreeg: You're spot on @message-dimke. Thanks for taking the time to look at this. 

PS: I learned about the DATE_ATOM constant.

Thanks,

R

message-dimke (approved):
Tested with timezone +10:00

General_Settings_‹_WordPress-Facebook_—_WordPress

Added Sale dates 

Edit_product_“Belt”_‹_WordPress-Facebook_—_WordPress

Updated the product

Checked Facebook

_3__Dima_commerce_account_name

Seems good to me.

Thanks!

PR #2215, opened 2022-05-20
Fix - Wrong Value Field in AddToCart Events

### LINKED ISSUE #1847: Wrong `Value` Field in AddToCart Events

rawdreeg:

When triggering multiple AddToCart events from a website, the value field contains a cumulative sum of all the previous AddToCart events instead of the value of the current AddToCart only.

This has a significant impact on multiple tools in the Facebook Business Manager platform.
For example, when monitoring campaign metrics the total value for AddToCart events is much higher than the actual value in WooCommerce carts.
This is because each new AddToCart event contains the cumulative sum of prices of all products in the cart:

- First AddToCart event: 1 product, value=X
- Second AddToCart event: 2 products in cart, value=X+Y
- Third AddToCart event: 3 products in cart, value=X+Y+Z

This PR fixes this.

Closes #1847.

_Replace this with a good description of your changes & reasoning._

1. Check out this branch
2. Navigate to a product page, add an item to your cart and observe that the value amount in the AddToCart pixel event code matches the value of the item added to the cart.
3. Repeat this with another product and confirm the value matches the value of the item added to the cart and not the cart total.
4. Go to WP Admin -> WooCommerce -> Settings -> Products. Ensure the option “Redirect to the cart page after successful addition” is unchecked. and "Enable AJAX add to cart buttons on archives" is selected.
5. Go to the shop page and add an item to the cart, then inspect the source code (right click -> inspect) and locate the div with class "wc-facebook-pixel-event-placeholder". Confirm it contains the pixel script tag and the values are correct:

Screenshot 2022-05-23 at 15 17 15
 

### Changelog entry

> Fix - Wrong Value Field in AddToCart Events

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -591,7 +591,7 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 					'content_name' => $this->get_cart_content_names(),
 					'content_type' => 'product',
 					'contents'     => $this->get_cart_contents(),
-					'value'        => $this->get_cart_total(),
+					'value'        => (float) $product->get_price() * $quantity,
 					'currency'     => get_woocommerce_currency(),
 				),
 				'user_data'   => $this->pixel->get_user_info(),
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -587,10 +587,17 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 			$event_data = array(
 				'event_name'  => 'AddToCart',
 				'custom_data' => array(
-					'content_ids'  => $this->get_cart_content_ids(),
-					'content_name' => $this->get_cart_content_names(),
+					'content_ids'  => wp_json_encode( \WC_Facebookcommerce_Utils::get_fb_content_ids( $product ) ),
+					'content_name' => $product->get_name(),
 					'content_type' => 'product',
-					'contents'     => $this->get_cart_contents(),
+					'contents'     => wp_json_encode(
+						array(
+							array(
+							"id"	   => \WC_Facebookcommerce_Utils::get_fb_retailer_id( $product ),
+							"quantity" =>  $quantity,
+							),
+						)
+					),
 					'value'        => (float) $product->get_price() * $quantity,
 					'currency'     => get_woocommerce_currency(),
 				),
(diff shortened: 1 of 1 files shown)


jorgemd24 (commented):
Thanks, @rawdreeg for checking this! I have tested locally and can see that `value` gets the product price however looking into the code, `content_ids` or `content_name` is sending all the items in the cart. Shouldn't this contain only the product that is being added to the cart?

image

rawdreeg:
> however looking into the code, content_ids or content_name is sending all the items in the cart. Shouldn't this contain only the product that is being added to the cart?

You're right! Thanks for catching up. I updated this PR to cover the content_ids, content_name, and content properties as well. I have also updated the add_add_to_cart_event_fragment method to correctly generate the pixel script

rawdreeg:
@jorgemd24  this is ready for another round of reviews :-) Thanks

jorgemd24 (approved):
Thanks, @rawdreeg, it is working as expected! I see the events in the Pixel page with the right data. Something that I saw in Pixel but I think it is more related to my account or products is that the event properties, source and version, are displayed as '_removed_'.

image

rawdreeg:
> Thanks, @rawdreeg, it is working as expected! I see the events in the Pixel page with the right data. Something that I saw in Pixel but I think it is more related to my account or products is that the event properties, source and version, are displayed as '_removed_'.

I think FB might be hiding this in the event manager. At least according to this [article](https://www.pixelyoursite.com/custom-data-not-being-displayed). 

Thanks for reviewing!

PR #2217, opened 2022-05-24
Facebook Order sync action was not scheduled.

### LINKED ISSUE #2110: Orders placed on Facebook don't appear in WooCommerce
## ✍️ Describe the bug

Orders are placed through Facebook's checkout but don't appear in WooCommerce orders list. On the Facebook side of things, the messaging says the order "needs to be acknowledged in WooCommerce." 

Order details (like customer name) do not appear in log files if debug logging is enabled. The orders just aren't reaching WooCommerce.

https://wordpress.org/support/topic/order-pending-acknowlegement/

Also 4415708-zen

message-dimke:

Some time ago `commerce manage id` option stopped to be set and turned the condition to evaluate to false when checking for Facebook order sync scheduled action. So fixing the sync scheduled action check condition. Besides `commerce manage id` appeared to be an unknown parameter, not a single doc has the reference to such a parameter name, the closest thing to `commerce manage id` is `commerce account id`, but still `commerce manage id` was not used anywhere else, so removing is an checking for page token to validate the connection should be sufficient. Orders fetching does not require `commerce manage id`, only page token.

Closes #2110.

1. Before checking out the branch, go to WooCommerce > Scheduled Actions;
2. Search for wc_facebook_commerce_fetch_orders and see no results found;
WooCommerce_status_‹_Regular-WordPress_—_WordPress

3. Checkout the branch with the fix and repeat those two steps above;
4. You should see action scheduled;

### Changelog entry

> Fix - Facebook orders sync action is not scheduled.

### DIFF
--- a/includes/Commerce.php
+++ b/includes/Commerce.php
@@ -111,11 +111,9 @@ class Commerce {
 	 *
 	 * @return bool whether the site is connected
 	 */
-	public function is_connected() {
+	public function is_connected(): bool {
 
-		$connection_handler = facebook_for_woocommerce()->get_connection_handler();
-
-		$connected = (bool) strlen( $connection_handler->get_page_access_token() ) && ! empty( $connection_handler->get_commerce_manager_id() );
+		$connected = ! empty( facebook_for_woocommerce()->get_connection_handler()->get_page_access_token() );
 
 		/**
 		 * Filters whether the site is connected.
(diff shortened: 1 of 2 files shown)


rawdreeg (approved):
✅  Thanks @message-dimke. Looks good!

jconroy:
Nice work.

Lets hold off on shipping this one until we've aligned internally on order functionality as a whole

budzanowski:
Closed, we are not using order sync.

PR #2240, opened 2022-07-02
Add/unit test Facebook integration class.

message-dimke:

Adding unit tests for `WC_Facebookcommerce_Integration` class, a missing test suit to proceed with further refactoring. Enabling PHPCS on tests and `WC_Facebookcommerce_Integration` class.

See all the checks pass for the PR.


class-wc-facebookcommerce.php
```diff
@@ -127,7 +127,6 @@ public function __construct() {
 		 */
 		public function init() {
 
-			add_action( 'init', array( $this, 'get_integration' ) );
```
message-dimke, line 130: Early Facebook Integration object initialisation when Facebook access token is not here present. Removing it, since it is not ever necessary.

includes/fbgraph.php
```diff
@@ -757,14 +753,29 @@ public function get_tip_info( $external_merchant_settings_id ) {
 			return $connect_woo;
 		}
 
-		public function get_facebook_id( $facebook_catalog_id, $product_id ) {
+		/*public function get_facebook_id( $facebook_catalog_id, $product_id ) {
```
rawdreeg, line 756: Can we remove the commented-out code?

tests/Unit/WCFacebookCommerceGraphAPITest.php
```diff
@@ -689,4 +691,39 @@ public function test_next() {
...
+		$expected = [
+			'id'            => 'product-id',
+			'product_group' => [
+				'id' => 'product-group-id',
+			],
+		];
+
+		$response = function( $result, $parsed_args, $url ) {
+			$this->assertEquals( 'GET', $parsed_args['method'] );
+			$this->assertEquals( 'https://graph.facebook.com/v12.0/catalog:726635365295186:d2NfcG9zdF9pZF8xMjc=/?fields=id,product_group{id}', $url );
```
rawdreeg, line 714: We've since updated to v13.0.

rawdreeg (changes requested):
Thanks, @message-dimke

Great work on this. Looking really good. I have left a few minor comments.

Also, it seems we made some changes to WC_Facebookcommerce_Graph_API and a few other classes since this was last updated. It'd be good to refresh the branch with the latest changes.

Thanks,

includes/fbgraph.php
```diff
@@ -757,14 +753,29 @@ public function get_tip_info( $external_merchant_settings_id ) {
...
 			$param = 'catalog:' . (string) $facebook_catalog_id . ':' .
 			base64_encode( $product_id ) . '/?fields=id,product_group{id}';
 			$url   = $this->build_url( '', $param );
 			// success API call will return
 			// {id: <fb product id>, product_group{id} <fb product group id>}
 			// failure API will return {error: <error message>}
 			return self::_get( $url );
+		}*/
```
rawdreeg, line 764:
```suggestion
```

rawdreeg (changes requested):

facebook-commerce-pixel-event.php
```diff
@@ -575,8 +571,7 @@ public static function get_use_pii_key() {
 			if ( ! $fb_options ) {
 				return null;
```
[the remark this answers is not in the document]
message-dimke: Afaik `return` will evaluate to `NULL`, and I prefer functions to always return something, especially because everywhere in the code we use this function on the right side of the equality.

includes/fbgraph.php
message-dimke: Sure =)

message-dimke:
@rawdreeg , while removing SkyVerge dependency I found that the old API supports requests throttling out of the box and the new one does not. We have few options, either decide to move few endpoints from new into old, or implement throttling for the new api. Done lots of tests and spent time on this one, so better implement throttling here, not to loose my work.

Anyway, that is something to decide later, after SkyVerge removal is approved. Because I will definitely have to either implement few apis and tests for the old API, or integrate this one into the new codebase.

rawdreeg:
>  Done lots of tests and spent time on this one, so better implement throttling here, not to loose my work.
Sounds good, @message-dimke.

> Anyway, that is something to decide later, after SkyVerge removal is approved. Because I will definitely have to either implement a few APIs and tests for the old API or integrate this one into the new codebase.

Wouldn't it make sense to introduce the Skyverge removal here and use the tests to facilitate the API changes?

message-dimke:
> Wouldn't it make sense to introduce the Skyverge removal here and use the tests to facilitate the API changes?

Not sure about 'here', but maybe merge this one into #2271, because API is a smaller bit than SV removal as we discussed here https://github.com/woocommerce/facebook-for-woocommerce/pull/2271#issuecomment-1220465149

rawdreeg:
> > Wouldn't it make sense to introduce the Skyverge removal here and use the tests to facilitate the API changes?
> 
> Not sure about 'here', but maybe merge this one into #2271, because API is a smaller bit than SV removal as we discussed here [#2271 (comment)](https://github.com/woocommerce/facebook-for-woocommerce/pull/2271#issuecomment-1220465149)

Yes, you're right. The SV change is too massive. Let's keep it simple. We can always take a second pass at this to integrate this one into the new codebase as you said

message-dimke:
@rawdreeg , I'll merge this one into https://github.com/woocommerce/facebook-for-woocommerce/tree/feature/facebook-api-refactoring, since it does not make sense to review this one separately, especially when doing Smoke Testing.

PR #2263, opened 2022-08-02
Fix - Trashed products in WooCommerce remain listed on Facebook.

### LINKED ISSUE #2256: Trashed products in WooCommerce remain listed on Facebook
## 🔎 Isolate the bug

## ✍️ Describe the bug

When a product is deleted in WooCommerce is remains on the shop page in Facebook.

## 🚶‍♀️ Steps to reproduce

1. Create a product and sync to Facebook\
2. Trash the product from within WooCommerce
3. See product remains unchanged on Facebook.

## ✔️ Expected behavior

In an attempt to minimize the possibility of orphaned products ( #2129 ) or the need to run a script to delete all previously deleted products from Facebook ( #1964 ), when a product is deleted in WooCommerce a request to delete the product from within Facebook should be made. 

The request should be made for individually trashed items or when using bulk edit to delete multiple products.

rawdreeg:

When a product is deleted in WooCommerce it remains on the shop page on Facebook. This PR makes a few changes in the way we handle deleted products:

- We run on_product_delete on the `wp_trash_post` action hook, allowing the product to be also removed from the FB catalog.
- Allow variable products to bypass the `published_product_should_be_synced` check, as this will always return false because some variation products might already be removed. 
- We also add in `fb_restore_untrashed_variable_product` function that runs on the `untrashed_post` action hook. This ensures variable products are correctly published after all the variations have been untrashed.

Closes #2256 .

1. Create a product and sync it to Facebook
2. Trash the product from within WooCommerce
3. See product has also been removed on Facebook.
4. Restore the product; the product is recreated on Facebook.

### Changelog entry

> Fix - Ensure product is deleted from FB when moved to trash.

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -423,6 +423,8 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 			3
 		);
 
+		add_action( 'untrashed_post', array( $this, 'fb_restore_untrashed_variable_product' ));
+
 		// Product Set hooks.
 		add_action( 'fb_wc_product_set_sync', array( $this, 'create_or_update_product_set_item' ), 99, 2 );
 		add_action( 'fb_wc_product_set_delete', array( $this, 'delete_product_set_item' ), 99 );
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
Looks good and works as intended! Thanks, Rodrigue!

PR #2270, opened 2022-08-11
Gracefully handle `get_current_url` when `$_SERVER['HTTP_REFERER']` not present

### LINKED ISSUE #1729: PHP news in my debug.log
I'm receiving these warnings when debugging my website, why is this happening?

`PHP Notice:  Undefined index: HTTP_REFERER in /*/wp-content/plugins/facebook-for-woocommerce/includes/Events/Event.php on line 211`

ecaron:

Closes #1729.

Switch to using `wp_get_raw_referer()` instead of `$_SERVER['HTTP_REFERER']`. `wp_get_raw_referer()` was introduced in [WordPress 4.5.0](https://developer.wordpress.org/reference/functions/wp_get_raw_referer/) and it also accomodates AJAX calls that utilize the `wp_referer_field` functionality (which has been around since [2.0.4](https://developer.wordpress.org/reference/functions/wp_referer_field/))

### Changelog entry

> Fix - `Undefined array key "HTTP_REFERER"` not longer happens when `new Event` is triggered from an AJAX call that doesn't include a referrer (likely due to browser configuration)

### DIFF
--- a/includes/Events/Event.php
+++ b/includes/Events/Event.php
@@ -215,7 +215,7 @@ class Event {
 
 		if ( wp_doing_ajax() ) {
 
-			$url = $_SERVER['HTTP_REFERER'];
+			$url = wp_get_raw_referer();
 
 		} else {
 


rawdreeg (approved):
👍🏾  LGTM!

Thank you for your contribution @ecaron. We'll merge this!

PR #2271, opened 2022-08-11
Removing SV Framework dependency.

message-dimke:

Current PR completely removes SkyVerge Framework dependency out of the extension. 
The architecture of the SkyVerge framework is tightly coupled which makes migration in parts not practically possible. 

Test instructions are to checkout the branch and run [Smoke Tests](https://github.com/woocommerce/facebook-for-woocommerce/wiki/Smoke-tests) against the extension.

### Changelog entry

> Update - Remove SkyVerge extension composer dependency.


message-dimke:
@rawdreeg , I think it is a good idea to merge new API into this PR first to run Smoke Tests only once. What do you think?

rawdreeg:
> @rawdreeg, I think it is a good idea to merge new API into this PR first to run Smoke Tests only once. What do you think?

I agree; we should probably run the smoke tests on the new API.

Another thing I have noticed, on some of the files, we seem to have updated the phpcs (e.g., short array), This is probably not a show stopper, but we'll need to revisit this on another PR based on the discussions with @jconroy (p1660643441679189/1660635367.197539-slack-CK365S85V)

message-dimke:
> Another thing I have noticed, on some of the files, we seem to have updated the phpcs (e.g., short array), This is probably not a show stopper, but we'll need to revisit this on another PR based on the discussions with @jconroy (p1660643441679189/1660635367.197539-slack-CK365S85V)

I plan another PR in these series to enable phpcs on files. Current PR has all the files ignore code styling. So we configure phpcs and then in the next PR get all the syntax made right.

Issue #2171, opened 2022-03-10
Cannot disable Facebook Messenger


csmcneill:
## 🔎 Isolate the bug

- [x] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug
When I navigate to WooCommerce > Marketing > Facebook > Messenger and disable `Enable Messenger`, the settings do not persist after selecting `Save Changes`.

## 🚶‍♀️ Steps to reproduce
1. Navigate to WooCommerce > Marketing > Facebook > Messenger 
2. If `Enable Messenger` is disabled, enable it and save settings.
3. Disable `Enable Messenger` and save settings.
4. Page refreshes, and `Enable Messenger` is still enabled.

Here's an animated GIF:

![Screen Capture on 2022-03-10 at 13-59-29](https://user-images.githubusercontent.com/82999806/158838175-00f05dbe-eac0-407b-8ffc-45904ddece3b.gif)

## ✔️ Expected behavior
Disabling `Enable Messenger` should persist once the settings are saved.

## 🗃 Logs

Logs:

<details>

```
03-10-2022 @ 13:20:59 - Request
method: GET
uri: https://graph.facebook.com/v12.0/fbe_business/fbe_installs?fbe_external_business_id=wootestsite-61b0cecc72158
user-agent: Facebook-for-WooCommerce/2.6.12 (WooCommerce/6.3.1; WordPress/5.9.1)
body:
duration: 0.10883s

03-10-2022 @ 13:20:59 - Response
code: 200
message: OK
body: {"data":[{"business_manager_id":"577930360029345","commerce_merchant_settings_id":"697563351447721","onsite_eligible":false,"pixel_id":"2807517832872773","profiles":["103549555604221","108342031597485"],"ad_account_id":"0","catalog_id":"2013278605501102","pages":["103549555604221","108342031597485"],"token_type":"User"}]}

03-10-2022 @ 13:20:59 - Retrieving page access token
03-10-2022 @ 13:56:26 - Request
method: GET
uri: https://graph.facebook.com/v12.0/fbe_business?fbe_external_business_id=wootestsite-61b0cecc72158
user-agent: Facebook-for-WooCommerce/2.6.12 (WooCommerce/6.3.1; WordPress/5.9.1)
body:
[...]

PR #2291, opened 2022-09-12
The hearbeat cron shouldn't update messenger config.

rawdreeg:

We recently decided to keep the messenger `wc_facebook_enable_messenger` setting localized, regardless of the option configured on the FBE dashboard. 

The `refresh_business_configuration` process which runs hourly would update the option based on what's configured on FB, thus overriding the local option. This PR removes the code to update the messenger settings

Related to #2171.

1. Navigate to WooCommerce > Marketing > Facebook > Messenger
2. If Enable Messenger is disabled, enable it and save settings.
3. Disable Enable Messenger and save settings.
4. Enable Messenger remains disabled, and the chat plugin does not load on the front end.
5. Using [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/) and reschedule the facebook_for_woocommerce_daily_heartbeat_cron task to run now.
6. Observe that the `wc_facebook_enable_messenger` option is not overridden.

### Changelog entry

> Fix - Messenger settings are no longer overridden after business config refresh.

### DIFF
--- a/includes/Handlers/Connection.php
+++ b/includes/Handlers/Connection.php
@@ -148,24 +148,6 @@ class Connection {
 				$response->is_ig_cta_enabled()
 			);
 
-			// update the messenger settings
-			if ( $messenger_configuration = $response->get_messenger_configuration() ) {
-
-				// store the local "enabled" setting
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
Thank you, @rawdreeg!

PR #2292, opened 2022-09-13
Track purchase event flag in session variable instead post meta table.

### LINKED ISSUE #2125: Use session variables instead of postmeta on orders
You are using `_wc_facebook_for_woocommerce_order_placed` and `_wc_facebook_for_woocommerce_purchase_tracked` to stop duplicate purchase events from firing. I would suggest that you use `WC()->session->set()` instead. Just these two meta values take up 360 000 rows in our postmeta table, increasing with about 400 000 per year at this rate.

At the very least, maybe it would be possible to delete these when the order is set to relevant order statuses, like completed, failed, cancelled and refunded?

rawdreeg:

We're using _wc_facebook_for_woocommerce_order_placed and _wc_facebook_for_woocommerce_purchase_tracked to stop duplicate purchase events from firing. These values are not cleared and may pile in the users' DB. This PR moves these flags to session variables.

Closes #2125.

- [x ] Do the changed files pass `phpcs` checks? Please remove `phpcs:ignore` comments in changed files and fix any issues, or delete if not practical.

1. Complete a purchase.
2. Observe the Pixel purchase event is tracked.
3. Refresh the thank you page. The `inject_purchase_event` method should exit, and no additional purchase events tracked.

### Changelog entry

> Fix - Track purchase event flag in session variable instead post meta table.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -918,10 +918,10 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 			}
 
 			// use a session flag to ensure an order is tracked with any payment method, also when the order is placed through AJAX
-			$order_placed_session_flag = '_wc_' . facebook_for_woocommerce()->get_id() . '_order_placed';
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
LGTM! Thanks @rawdreeg!

PR #2297, opened 2022-09-16
Adding new filter to allow customizing a product group's default variation

urikanonov:
Today a product group's default variation is chosen based only on the default attributes of the product. If a listing has no default attributes then there is no way to set a default variation.
This affects the way that product prices are displayed in an Instagram shop connected to the Facebook feed (the prices are basically random - confusing the users).
This filter will allow advanced users to customize the selection based on other factors like price or custom attributes.

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -1485,12 +1485,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		);
 
 		// Figure out the matching default variation.
-		$default_product_fbid = apply_filters(
-			'wc_facebook_product_group_default_variation',
-			$this->get_product_group_default_variation( $woo_product, $fb_product_group_id ),
-			$woo_product,
-			$fb_product_group_id
-		);
+		$default_product_fbid = $this->get_product_group_default_variation( $woo_product, $fb_product_group_id );
 
 		if ( $default_product_fbid ) {
 			$product_group_data['default_product_id'] = $default_product_fbid;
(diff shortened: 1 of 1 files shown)


facebook-commerce.php
```diff
@@ -1485,7 +1485,12 @@ function update_product_group( $woo_product ) {
 		);
 
 		// Figure out the matching default variation.
-		$default_product_fbid = $this->get_product_group_default_variation( $woo_product, $fb_product_group_id );
+		$default_product_fbid = apply_filters(
```
rawdreeg, line 1488: I suggest applying the filter within the `get_product_default_variation` method rather:

https://github.com/woocommerce/facebook-for-woocommerce/blob/0b4eecc7de4e07bf02d84b746ff4bab0ee0c84aa/facebook-commerce.php#L1570


I would also recommend adding a doc block to the new filter as done here:

https://github.com/woocommerce/facebook-for-woocommerce/pull/2169/files#diff-8f3102bd79b0b163d0cbb9a7341231014b0dcd6a70d4cd25cdab2dc8f432ba9bR104-R112

rawdreeg (changes requested):

rawdreeg:
Hello @urikanonov,

Thanks for your contributions. I have requested some changes.

Thanks,

R

facebook-commerce.php
urikanonov: @rawdreeg - Thank you for the comments. I've pushed changes. Could you please take a look?

urikanonov:
> Hello @urikanonov,
> 
> Thanks for your contributions. I have requested some changes.
> 
> Thanks,
> 
> R

Thank you for the comments. I've pushed changes. Could you please take a look?

rawdreeg:
Hi @urikanonov,

Thanks for making these changes. I had a look, and I prefer having the default variation set [using the UI ](https://woocommerce.com/document/variable-product/#setting-defaults). Still, there might be some value in adding a filter for `$product_group_data` as `facebook_for_woocommerce_integration_prepare_product` filters [`$product_data`](https://github.com/woocommerce/facebook-for-woocommerce/blob/7050354be7a647cda5839eb157e25a622fb2a460/includes/fbproduct.php#L659-L663)

@budzanowski any thoughts?

urikanonov:
At least for my needs a code based filter will suffice (in fact simply UI
won't be enough). So please, I'd like to limit this checkin to just the
filter.

rawdreeg (approved):
Thanks for your contribution, @urikanonov ! Let's get this merged! 👍🏾

Issue #2301, opened 2022-09-22
Incorrect admin notification for user without permission


ibndawood:
## 🔎 Isolate the bug

- [x] I have confirmed this occurs in the most recent version of WordPress, WooCommerce, and Facebook for WooCommerce.
- [x] I have confirmed this occurs when only WooCommerce and Facebook for WooCommerce are active and when using a default WordPress or WooCommerce theme.

## ✍️ Describe the bug

Related to #2300. When Facebook for WooCommerce is activated without the activation of WooCommerce, an admin notice is displayed asking the user to update Facebook. This notification should be displayed only for users who have access to update the plugins and those who do not have access, for example, a subscriber.

![image](https://user-images.githubusercontent.com/179533/191747230-2f9ffc03-3547-4ea1-bb97-96bc40b02da2.png)
The Dashboard of a user whose role is Subscriber. The user does not have the capability to update plugins.

## 🚶‍♀️ Steps to reproduce

1. Install and activate Facebook for WooCommerce without activating the WooCommerce plugin.
2. Create a new user whose role is `Subscriber`. Log in to the admin panel, the notice is displayed asking the user to update the plugin. However, the user does not have access to the update feature in WordPress.

## ✔️ Expected behavior

The notice should only be displayed for users who have the capability to update plugins.

PR #2302, opened 2022-09-23
Adds helpful admin notices for correct user roles.

### LINKED ISSUE #2300: Incorrect admin notification displayed
## 🔎 Isolate the bug

## ✍️ Describe the bug

When Facebook for WooCommerce plugin is activated without WooCommerce active or installed. The header notification reads:
> Facebook for WooCommerce requires WooCommerce version 3.5.0 or higher. Please [update WooCommerce](https://example.com/wp-admin/update-core.php) to the latest version, or [download the minimum required version »](https://downloads.wordpress.org/plugin/woocommerce.3.5.0.zip)

The notification first recommends updating the plugin and redirects to the update admin page and then recommends downloading the minimum required version which is v3.5.0.

Here is the screenshot:
image

## 🚶‍♀️ Steps to reproduce

1. Set up WordPress and install Facebook for WooCommerce without activating WooCommerce.

## ✔️ Expected behavior

The notification should ask to install and activate WooCommerce instead of asking to update.

ibndawood:

Closes #2300 & #2301.

Admin notices that are displayed on activation of WooCommerce are not helpful. This PR adds the following helpful messages:
* Checks if WooCommerce is installed and prompts to install. The notice is displayed only for users who have the capability to install plugins. Earlier it was asking to update WooCommerce and to all types of users.
* Checks if WooCommerce is activated and prompts to activate. The notice is displayed only for users who can activate plugins.
* Checks if WooCommerce minimum WooCommerce version is installed and activated. The notice is displayed only for users who can update plugins.
* All the above admin notices are displayed only for users who have the required capabilities.
* The above notices are not displayed in the plugin update screen. 

This PR also has additional changes:
* Hooks `add_plugin_notices` to `admin_notices` hook instead of `admin_init`. This is because the function `get_screen` is not available at `admin_init`.
* Made the notices translatable.
* Methods that check for WooCommerce installation and activation.

#### When the plugin is activated without WooCommerce installed
image

#### When the plugin is activated with WooCommerce installed but not activated
image

1. Install and activate Facebook for WooCommerce without WooCommerce installed. The notice should read: 
> Facebook for WooCommerce requires WooCommerce to be installed and activated. Please install WooCommerce.
2. Click on **Install WooCommerce**. The link should download the latest version of WooCommerce.
3. Without activating WooCommerce, the notice should read:
> Facebook for WooCommerce requires WooCommerce to be activated. Please activate WooCommerce.
4. When updating/installing WooCommerce, the notice should not be displayed.
5. Click on **Activate WooCommerce**. The link should activate the plugin.
6. Create a new Subscriber user.
7. Log in as a subscriber and make sure the notices are not displayed. The notices are not required because the subscriber does not have the capability to install, update or activate plugins.

### Additional details:

> Tweak - added helpful admin notices, display them only for capable users, and made them translatable.

### DIFF
--- a/facebook-for-woocommerce.php
+++ b/facebook-for-woocommerce.php
@@ -223,7 +223,8 @@ class WC_Facebook_Loader {
 					'update_wordpress',
 					'error',
 					sprintf(
-						'%s requires WordPress version %s or higher. Please %supdate WordPress &raquo;%s',
+						/* translators: %1$s - plugin name, %2$s - minimum WordPress version required, %3$s - update WordPress link open, %4$s - update WordPress link close */
+						esc_html__( '%1$s requires WordPress version %2$s or higher. Please %3$supdate WordPress &raquo;%4$s', 'facebook-for-woocommerce' ),
 						'<strong>' . self::PLUGIN_NAME . '</strong>',
 						self::MINIMUM_WP_VERSION,
 						'<a href="' . esc_url( admin_url( 'update-core.php' ) ) . '">',


rawdreeg (approved):
LGTM! 👍🏾 

Thanks, @ibndawood!

PR #2340, opened 2022-10-26
Move Facebook menu notification to Inbox

### LINKED ISSUE #2299: Move The Facebook menu is now located... message to inbox.
image

The **Heads up! The Facebook menu is now located under the Marketing menu** message should only show for stores old stores prior to when this change was made, and it should move to being an inbox notice instead.

ibndawood:

Closes #2299 

In this PR:

1. Adds a class `SettingsMoved` note that inherits `NotesTrait` to display the "Facebook menu is moved..." admin message as an inbox note.
2. Removes "Facebook menu is moved..." admin notice.
3. Adds a method to `WC_Facebookcommerce` to access `lifecycle_handler` that allows us to get the lifecycle of the plugins. This method will return the last event of the lifecycle. The note should be displayed if the previous event is an upgrade and if it happened from v2.1.4.

### Screenshots:

image

1. Generate the build from this PR by running `npm run build`
2. Use local to set up 2 WordPress installations.
3. In one instance, download v2.1.4 from here: https://downloads.wordpress.org/plugin/facebook-for-woocommerce.2.1.4.zip, and install and activate Facebook for WooCommerce.
4. After activation, navigate to Plugins > Add New > Upload Plugin and upload the zip file generated from step 1. We are no upgrading from v2.1.4 to the version in the build.
5. Activate the plugin after the update and navigate to WooCommerce > Home. The Inbox should have a note as shown in the screenshot.
6. In another instance, install the build from the PR. Verify that no admin notices or inbox notes are added.
7. Verify the admin notice "Heads up! The Facebook menu is now located under the Marketing menu" is not displayed anywhere.

### Additional details:

The Facebook menu was moved under the Marketing menu in the WP admin panel in v2.2.0. This note should be displayed only to users updating the plugin from before v2.2.0.

The Lifecycle handler stores events such as installation and upgrade. It does not record downgrades. It also stores the latest version in the options table using this key: `wc_facebook_for_woocommerce_version`. This key is not updated on downgrades. Hence, it does not record events such as upgrades if we do them after installing a recent version. 

For example, if we install v2.6.28 and then downgrade to v2.1.4, and then upgrade to v2.6.27, the last upgrade will not be recorded. Hence for testing, we need to use 2 separate installations.

### Changelog entry

> Add - Inbox note about Facebook menu moved under the Marketing menu.

### DIFF
--- a/includes/Admin/Notes/SettingsMoved.php
+++ b/includes/Admin/Notes/SettingsMoved.php
@@ -39,7 +39,7 @@ class SettingsMoved {
 		 * only to users updating from a version prior to v2.2.0.
 		 */
 		$should_display = false;
-		$last_event     = \facebook_for_woocommerce()->get_last_event_from_history();
+		$last_event     = facebook_for_woocommerce()->get_last_event_from_history();
 
 		if ( isset( $last_event['name'] ) && 'upgrade' === $last_event['name'] ) {
 			$last_version = $last_event['data']['from_version'];


rawdreeg (commented):
Thanks, @ibndawood,

I tested this PR, and everything worked as expected. I left a small comment regarding the use of `\facebook_for_woocommerce()`.

PR #2356, opened 2022-11-04
Redirect the users to advertise tab of the plugin when FBE is installed

rahulmaddineni:
When FBE is successfully installed, instead of redirecting the user to settings tab redirect them to the advertise tab.

### Screenshots:
Tested with Local App running WooCommerce 7

https://user-images.githubusercontent.com/17508935/200050471-ad88e022-ab69-47d7-952e-5baaf919a0a4.mov

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -1013,6 +1013,19 @@ if ( ! class_exists( 'WC_Facebookcommerce' ) ) :
 
 			return admin_url( 'admin.php?page=wc-facebook' );
 		}
+		
+		/**
+         * Gets the advertise tab page URL.
+         *
+         * @since x.x.x
+         *
+         * @return string
+         */
+        public function get_advertise_tab_url() {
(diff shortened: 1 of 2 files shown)


budzanowski:
@rahulmaddineni,
We have reviewed the code. @rawdreeg has made one small addition. If the connection procedure fails we still want to redirect to settings. Going to Ads makes no sense when there is no connection. We have let ourselves to do the commits to speed things up a bit. I hope that this is OK with you!

budzanowski (approved):

rahulmaddineni:
> @rahulmaddineni, We have reviewed the code. @rawdreeg has made one small addition. If the connection procedure fails we still want to redirect to settings. Going to Ads makes no sense when there is no connection. We have let ourselves to do the commits to speed things up a bit. I hope that this is OK with you!

Hi @budzanowski & @rawdreeg good catch on the exception. Thanks for making the change & merging it.

PR #2395, opened 2022-11-24
Prevent class redeclaration error for class WC_Facebookcommerce_Utils

ibndawood:

Previously `WC_Facebookcommerce_Utils` had a `class_exists` check. It was removed. I've now added back the exact code to check for `class_exists` before loading the `WC_Facebookcommerce_Utils` class.

### Changelog entry

> Fix - Prevent class redeclaration error for class WC_Facebookcommerce_Utils


layoutd (approved):
Looks good 💯 

❔ Out of curiosity, what is the reason for using `if (): … endif;` over `{ … }` for this if block? Or is it just to match the original?

For reference, [this is the change that removed the original check](https://github.com/woocommerce/facebook-for-woocommerce/commit/c7869304890d49aeb6ee017de295077041d8301d#diff-8f3102bd79b0b163d0cbb9a7341231014b0dcd6a70d4cd25cdab2dc8f432ba9b) and this fix is related to [this support forum request](https://wordpress.org/support/topic/cannot-declare-class-wc_facebookcommerce_utils-because-the-name-is-already-in-u/).

PR #2457, opened 2023-01-20
Advandtage+ Shoppings Ads support in Advertise tab

vahidkay-meta:

This change changes the advertise tab so it runs an experiment for ASA campaigns. So based on the experiment parameters it shows either lwi-e or ASA view.

This is done as a task to simplify the way the ads are created in the FB-for-Woo plugin.

1- Connect your ad account
2- Set up your campaigns.
3- Check the insights and campaigns from Ads manager and this plugin to show the same values.


includes/fbutils.php
```diff
@@ -445,6 +445,17 @@ public static function get_store_name() {
 			return ( self::$store_name ) ? ( self::$store_name ) : 'A Store Has No Name';
 		}
 
+		public static function get_store_url() {
```
rawdreeg, line 448: not sure why this is needed

rawdreeg:
Thanks, @vahidkay-meta, for your momentous work on this PR.
It'll take some time to review everything, but here are some of my early general feedback:

1. `@since` in the doc bloc should be set to x.x.x as we replace this on release. eg: [here](https://github.com/woocommerce/facebook-for-woocommerce/pull/2457/files#diff-a1e6a6cf0a7255afea85e377040a757915fa8707bb9e88d7348348ee27d79f77R626)
2. Every string needs to be translatable. Seel [__()](https://developer.wordpress.org/reference/functions/__/)
3. I assume the echo statements are used for debugging and will be replaced when the PR is ready to release. You can use the log function to log any debugging statements. eg: [here](https://github.com/woocommerce/facebook-for-woocommerce/pull/2457/files#diff-5826d239f374e1cda2c64d674b49fc8164979b07a776c317b43bb89f6aa44612R1473)
4. We'll need to remove the ` phpcs:ignoreFile` comment to ensure all the files pass linting.

assets/js/admin/settings-advertise-asc.js
```diff
@@ -0,0 +1,321 @@
...
+      };
+    } catch (error) {
+      // skip
+      return '';
+    }
+
+  }
+
+  function check_for_changes( view ) {
+    hide_error();
```
ecgan, line 88: `hide_error` function has a `view` parameter, but it is not passed to the function call here.

ecgan (commented):
I took a bit of time to briefly review the JS and CSS code files for a short while, I have not tested the code yet. Leaving some early comments here. 

I suggest setting up linting and formatting with `@wordpress/scripts`, it would help address and prevent issues.

includes/AJAX.php
```diff
@@ -101,6 +107,104 @@ function( $name, $slug ) use ( $term ) {
...
+	 * Retrieves the changeset for each campaign type and posts them to the backend.
+	 * Makes sure that there is something to be sent.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string
+	 */
+	public function publish_ad_changes() {
+
+		$asc_rt = 'retargeting';
```
rawdreeg, line 122: 2 points here:

1. Can we use more descriptive variable names? eg: `retargeting` instead of `$rt`
2. `$asc_nb` and `$asc_rt` already exists as constants `NewBuyers::ID` and `Retargeting`. These are redeclared in a few places and not always consistently used.

includes/API/Insights/Response.php
```diff
@@ -0,0 +1,71 @@
...
+			'cart'      => $add_to_cart,
+			'purchases' => $purchase,
+		);
+
+		return array(
+			'spend'   => $spend,
+			'reach'   => $reach,
+			'actions' => $actions,
+		);
+	}
```
rawdreeg, line 63: I think this method can be simplified for easier readability:

``` PHP
public function get_result() {
    $data = $this->get_data();

    $result = array(
        'spend' => 0,
        'reach' => 0,
        'actions' => array(
            'clicks' => 0,
            'views' => 0,
            'cart' => 0,
            'purchases' => 0
        )
    );

    $result['spend'] = $data['spend'] ?? 0;
    $result['reach'] = $data['reach'] ?? 0;

    if ( array_key_exists('actions', $data) ) {
        foreach ( $data['actions'] as $action ) {
            if ( 'link_click' === $action['action_type'] ) {
                $result['actions']['clicks'] = $action['value'];
            } elseif ( 'view_content' === $action['action_type'] ) {
                $result['actions']['views'] = $action['value'];
            } elseif ( 'add_to_cart' === $action['action_type'] ) {
                $result['actions']['cart'] = $action['value'];
            } elseif ( 'purchase' === $action['action_type'] ) {
                $result['actions']['purchases'] = $action['value'];
            }
        }
    }

    return $result;
}
```

includes/fbutils.php
vahidkay-meta: I need the store url for the campaign to be created. 
So I first try to get the site_url, if it doesn't work I try with hostname

includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -287,7 +288,7 @@ private function try_render_experimental_view() {
 		$translate = function( $text ) {
 			return $this->get_escaped_translation( $text );
 		};
-		$translate_with_link = function( $hyperlink, $link_text, $pretext, $rest_of_text ) {
+		$translate_with_link = function( $hyperlink, $link_text, $pretext, $rest_of_text ) use(&$translate) {
 			return $translate( $pretext ) . " <a href='<?php echo " . $hyperlink . "?>' >" . $translate( $link_text ) . "</a>" . $translate( $rest_of_text );
 		};
```
rawdreeg: Why aren't these moved to separate methods? You could even add this to the [Util](https://github.com/woocommerce/facebook-for-woocommerce/blob/develop/includes/fbutils.php) class.

vahidkay-meta: Good point. I'll do that.

includes/API/Insights/Response.php
```diff
@@ -0,0 +1,57 @@
...
+					$result['actions']['clicks'] = $action['value'];
+				} elseif ( 'view_content' === $action['action_type'] ) {
+					$result['actions']['views'] = $action['value'];
+				} elseif ( 'add_to_cart' === $action['action_type'] ) {
+					$result['actions']['cart'] = $action['value'];
+				} elseif ( 'purchase' === $action['action_type'] ) {
+					$result['actions']['purchases'] = $action['value'];
+				}
+			}
+		}
```
rawdreeg, line 46: I think this block can be a bit more concise. Just a suggestion

```suggestion
$action_map = [
    'link_click' => 'clicks',
    'view_content' => 'views',
    'add_to_cart' => 'cart',
    'purchase' => 'purchases'
];

if ( isset( $data[ 'actions' ] ) ) {
    foreach ( $data[ 'actions' ] as $action ) {
        $action_type = $action[ 'action_type' ];
        if ( isset( $action_map[ $action_type ] ) ) {
            $result[ 'actions' ][ $action_map [ $action_type ] ] = $action[ 'value' ];
        }
    }
}

```

includes/Admin/Settings.php
```diff
@@ -181,16 +181,16 @@ private function connect_to_enhanced_admin( $screen_id ) {
...
-						$crumbs[] = __( 'Messenger', 'facebook-for-woocommerce' );
+						$crumbs[] = __( 'Facebook for WooCommerce', 'facebook-for-woocommerce' );
 						break;
 					case Settings_Screens\Product_Sync::ID:
-						$crumbs[] = __( 'Product sync', 'facebook-for-woocommerce' );
+						$crumbs[] = __( 'Facebook for WooCommerce', 'facebook-for-woocommerce' );
 						break;
 					case Settings_Screens\Advertise::ID:
-						$crumbs[] = __( 'Advertise', 'facebook-for-woocommerce' );
+						$crumbs[] = __( 'Facebook for WooCommerce', 'facebook-for-woocommerce' );
```
rawdreeg, line 193: Any reason for default the crumb to a Single value? If this is changing then the switch case seem redundant

includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -199,6 +250,222 @@ public function render() {
...
+
+	/*
+	 * Renders the ASC Experimental view.
+	 *
+	 * @since x.x.x
+	 *
+	 */
+	private function experimental_view_render() {
+
+		// if ( $this->can_try_experimental_view() ) {
```
rawdreeg, line 267: We'll need to clean up the commented up codes

includes/Admin/Settings_Screens/Advertise.php
```diff
@@ -199,6 +250,222 @@ public function render() {
...
+		 	$this->try_render_experimental_view();
+
+		// } else {
+
+		// 	$this->render_lwi_view();
+
+		// }
+	}
+
+	private function render_dashboard( $type, $text1, $text2 ) {
```
rawdreeg, line 278: Docbloc comments :-) and the args names need to be more descriptive

includes/AdvertiseASC/NewBuyers.php
```diff
@@ -0,0 +1,306 @@
...
+
+		// if ( $countries || $daily_budget ) {
+		// 	$this->apply_adset_changes( $countries, $daily_budget );
+		// }
+
+		// if ( array_key_exists( 'p1', $data ) ) {
+		// 	$this->set_ad_status( $data['p1'] );
+		// }
+
+		// return true;
```
rawdreeg, line 225: We'll need to clean up the commented out codes. This method isn't used anywhere it seems

package.json
```diff
@@ -37,5 +37,12 @@
   "engines": {
     "node": ">=12.22 <=16",
     "npm": ">=6.14 <=8"
+  },
+  "dependencies": {
+    "antd": "^5.9.2",
+    "dayjs": "^1.11.9",
+    "react": "^18.2.0",
```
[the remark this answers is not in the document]
rawdreeg: +1

rawdreeg (changes requested):
@vahidkay-meta,

I left some comments for you to review. I haven't looked into the technical implementation in depth, but I wanted to leave some early comments.

Two main points to address:

- We need to clean up the code -- remove unused methods, add documentation to methods, remove commented-out code, and check coding formatting.

- Review the react frontend implementation. We'll need to use WordPress react components instead of including our own react, as this might result in conflicts. Let us know if you have any questions

package.json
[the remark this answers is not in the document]
vahidkay-meta: Will remove the react. I didn't know WP had it

includes/AJAX.php
```diff
@@ -47,6 +49,14 @@ public function __construct() {
...
 		add_action( 'wp_ajax_wc_facebook_get_sync_status', array( $this, 'get_sync_status' ) );
 
+		// get the ad preview
+		add_action( 'wp_ajax_wc_facebook_get_ad_preview', array( $this, 'get_ad_preview' ) );
+
+		add_action( 'wp_ajax_wc_facebook_generate_ad_preview', array( $this, 'generate_ad_preview' ) );
+
+		// sync the ad/campaign changes with the marketing api.
+		add_action ( 'wp_ajax_wc_facebook_advertise_asc_publish_changes', array( $this, 'publish_ad_changes' ));
+
```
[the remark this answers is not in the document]
vahidkay-meta: Will look into it

assets/js/admin/advertise-asc-components/campaign-management-view.js
```diff
@@ -0,0 +1,141 @@
...
+
+        setPublishing(true);
+        fetch(facebook_for_woocommerce_settings_advertise_asc.ajax_url + '?action=wc_facebook_advertise_asc_publish_changes', {
+            method: 'post',
+            headers: { 'Content-Type': 'application/json' },
+            body: requestData
+        })
+            .then((response) => response.json())
+            .then((data) => {
+                //TODO: HANDLE ERROR, if success = false;
```
budzanowski, line 55: Error handling is missing.

assets/js/admin/advertise-asc-components/campaign-management-view.js
```diff
@@ -0,0 +1,141 @@
...
+                console.log(data);
```
budzanowski, line 56: console.log should be removed.

package.json
```diff
@@ -37,5 +37,12 @@
   "engines": {
     "node": ">=12.22 <=16",
     "npm": ">=6.14 <=8"
+  },
+  "dependencies": {
+    "antd": "^5.9.2",
```
budzanowski, line 42: WordPress and WooCommerce have their own native components systems.

assets/js/admin/advertise-asc-components/campaign-edit-view.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import { useState } from '@wordpress/element';
 import { Card, Col, Form, Input, InputNumber, Row, Select, Space, Switch, Tooltip } from 'antd';
```
rawdreeg, line 2: ❓ Any reason to use ant design components instead of wordpress/element ?

includes/Admin/Settings.php
vahidkay-meta: This is as per a conversation we had with Matthias. We want to show Facebook for Woocommerce on top of the page instead of tab name. 
You're right. I'll remove the switch case

assets/js/admin/advertise-asc-components/campaign-edit-view.js
vahidkay-meta: Just familiarity with the ant framework and the tight deadline.

rawdreeg: I understand, but Facebook for WooCommerce is one of our canonical extensions and we'd prefer if we make use of woocommerce/components? I am trying to avoid dependency on a third-party library we might have to maintain in the long run.

Could you have a look at https://www.npmjs.com/package/@woocommerce/components?activeTab=readme maybe and let me know if this is something you'd consider?

vahidkay-meta: We could consider it as a tech debt and switch when there's time, but for the current PR I don't think I can do it.

vahidkay-meta: Seems like the woocommerce/components package requires a higher version of php, etc. than what we have atm. 
We are on php 7.1 and also WP 14. So the plugin needs some housekeeping before we can use that @woocommerce/components package.

budzanowski: > We are on php 7.1

No we are on 7.4:

https://github.com/woocommerce/facebook-for-woocommerce/blob/51cbb2fd7554f168b2410494d1ce95f782ae28c2/composer.json#L13

What is WP 14?

package.json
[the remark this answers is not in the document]
Ismail-elkorchi: > Importing react and such directly can lead to a lot of issues.

Based on my understanding of the issue and the related pull requests below, the updated recommendation is to use React directly:

https://github.com/WordPress/gutenberg/issues/54074
https://github.com/WordPress/gutenberg/pull/54494
https://github.com/WordPress/gutenberg/pull/54908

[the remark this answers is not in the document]
budzanowski: This plugin is installed on exceptionally large number of machines - sometimes with very old versions of WP and Woo. Nobody knows how they will behave when we will introduce recommendations for the latest version of Gutenberg - which have not even been released yet. Besides that the Issues that you have mentioned if I read them correctly suggest having react as a peer dependency and not bundling it with the plugin. The size of the plugin is something that need to be addressed as well. Current  version  ( 3.0.34 ) zip size is 792 KB. The zip file that @vahidkay-meta  send for testing is 1,330 KB. That is clearly not a healthy change.

assets/js/admin/advertise-asc-components/campaign-edit-view.js
vahidkay-meta: Seems like we are on two different branches. Or maybe mine is a bit too old

vahidkay-meta: a simple installation of the woocommerce/components package shows many conflicts with the current @wordpress package that's being used. I cannot simply test everything before making such a change. This can risk the timely release of ASC

ecgan: I think this wasn't mentioned yet - there is also [`@wordpress/components`](https://www.npmjs.com/package/@wordpress/components) package (see [storybook](https://wordpress.github.io/gutenberg/)).

Using `@wordpress/components` - combined with https://www.npmjs.com/package/@wordpress/dependency-extraction-webpack-plugin that should already be setup in your use of https://www.npmjs.com/package/@wordpress/scripts - should make the plugin significantly lighter in size.

vahidkay-meta: Nice. I'll take a look.

vahidkay-meta:
I have done the smoke test. Could you please help merging this, so we can release this asap?

assets/js/admin/settings-advertise-asc.js
rawdreeg, line 1: Is this file still in use?

facebook-commerce.php
```diff
@@ -1527,6 +1533,7 @@ public function create_or_update_product_set_item( $product_set_data, $product_s
 				);
 			}
 		} catch ( ApiException $e ) {
+
 			$message = sprintf( 'There was an error trying to create/update a product set: %s', $e->getMessage() );
```
rawdreeg, line 1537: This along with all the messages displayed to the user will need to be made translatable

includes/API/Adset/Update/Response.php
```diff
@@ -0,0 +1,49 @@
...
+	 * @return string
+	 */
+	public function get_id() {
+		return $this->response_data['id'];
+	}
+
+	/**
+	 * Gets the countries for the adset.
+	 *
+	 * @since 2.0.0
```
rawdreeg, line 42: Some the 'since' values are obviously wrong. We'd need to ensure this is accurate.

rawdreeg (commented):
I left some more comments regarding coding style:

Some missing or incorrect docbloc comments

Some strings are not translated.

We need to log an issue to track this and fix this in a future.

rawdreeg (approved):

PR #2465, opened 2023-01-31
Add unit tests to update plugin version API call

ibndawood:

This PR:
* Adds unit tests for the `WooCommerce\Facebook\ExternalVersionUpdate\Update` class added in the PR #2442 
* Covers unit tests for methods `should_update_version` and `maybe_update_external_plugin_version` methods.

1. Open the terminal on the plugin folder.
2. Run ~~`composer update`~~ `composer dumpautoload` to regenerate the autoload files.
3. Run the command `vendor/bin/phpunit --filter=UpdateTest`
4. Verify all tests and assertions are passed.

### DIFF
--- a/tests/Unit/ExternalVersionUpdate/UpdateTest.php
+++ b/tests/Unit/ExternalVersionUpdate/UpdateTest.php
@@ -56,12 +56,10 @@ class UpdateTest extends WP_UnitTestCase {
 
 		// Connection Handler mock object to return is_connected as false.
 		$mock_connection_handler = $this->getMockBuilder( Connection::class )
-											->disableOriginalConstructor()
-											->setMethods( array( 'is_connected' ) )
-											->getMock();
-		$mock_connection_handler->expects( $this->any() )
-											->method( 'is_connected' )
-											->willReturn( false );
+									->disableOriginalConstructor()
+									->setMethods( array( 'is_connected' ) )
+									->getMock();
+		$mock_connection_handler->expects( $this->any() )->method( 'is_connected' )->willReturn( false );
 		$prop_connection_handler->setValue( $plugin, $mock_connection_handler );
 
 		update_option( 'facebook_for_woocommerce_latest_version_sent_to_server', '0.0.0' ); // Reset the option.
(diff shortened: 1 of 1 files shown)


tests/Unit/ExternalVersionUpdate/UpdateTest.php
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	/**
+	 * Test should_update_version
+	 */
+	public function test_should_update_version() {
+		$plugin = facebook_for_woocommerce();
+
+		// Assert update not required when the versions match.
+		update_option( 'facebook_for_woocommerce_latest_version_sent_to_server', \WC_Facebookcommerce_Utils::PLUGIN_VERSION );
```
JPry, line 38: I know that tests commonly call `update_option()`, but I feel that it is a bit cleaner to make use of the `pre_option_{$option}` filter to return the value that we want during the test. This can even be done with an anonymous function, which is then unhooked (rather than resetting the option). Something like this:

```php
$filter = function() {
	return WC_Facebookcommerce_Utils::PLUGIN_VERSION;
};

add_filter( 'pre_option_facebook_for_woocommerce_latest_version_sent_to_server', $filter );

// testing code, then later...
remove_filter( 'pre_option_facebook_for_woocommerce_latest_version_sent_to_server', $filter );

// add it again...
add_filter( 'pre_option_facebook_for_woocommerce_latest_version_sent_to_server', $filter );

// remove it again
remove_filter( 'pre_option_facebook_for_woocommerce_latest_version_sent_to_server', $filter );
```

tests/Unit/ExternalVersionUpdate/UpdateTest.php
```diff
@@ -0,0 +1,183 @@
...
+		/**
+		 * Set the $plugin->connection_handler and $plugin->api access to true. This will allow us
+		 * to assign the mock objects to these properties.
+		 */
+		$plugin_ref_obj          = new \ReflectionObject( $plugin );
+		$prop_connection_handler = $plugin_ref_obj->getProperty( 'connection_handler' );
+		$prop_connection_handler->setAccessible( true );
+
+		// Connection Handler mock object to return is_connected as false.
+		$mock_connection_handler = $this->getMockBuilder( \WooCommerce\Facebook\Handlers\Connection::class )
```
JPry, line 51: Minor: I won't comment on everything else, but I think every reference outside of this namespace should be imported.

tests/Unit/ExternalVersionUpdate/UpdateTest.php
```diff
@@ -0,0 +1,183 @@
...
+											->disableOriginalConstructor()
```
JPry, line 52: Minor: These chained calls should be indented only one extra tab, rather than attempting to align them farther to the right.

JPry (approved):
I made a few minor suggestions, but none of them are blocking. I can see the tests running successfully in my local environment.

JPry:
> Run `composer update` to regenerate the autoload files.

Running `composer dumpautoload` will regenerate the autoload files. Alternatively, it would be good to run `composer install`. If you run `update`, it will change the `composer.lock` file.

PR #2467, opened 2023-02-02
PHP 8.1 deprecated warnings

layoutd:

This fixes two sets of deprecated warnings in PHP 8.1:

1. Caused by missing method return type declarations in [`includes/Framework/Api/JSONResponse.php`](https://github.com/woocommerce/facebook-for-woocommerce/blob/develop/includes/Framework/Api/JSONResponse.php):
    image

    👉🏻  Adding the return types resolves this issue.

2. Caused by [passing the return value of a method that doesn't return anything to `wp_kses`](https://github.com/woocommerce/facebook-for-woocommerce/blob/6ceea0a49599c7d38a7095426de6d288fbbc43c3/includes/Admin/Product_Sets.php#L76-L83):
    image
    image

    👉🏻  Since the void methods display the HTML for the label and field directly, there's no need to sanitize or `echo` their return values.

1. With PHP 8.1 and the `develop` branch, connect a store to Facebook and visit the `Marketing > Facebook > Connection` tab (`/wp-admin/admin.php?page=wc-facebook&tab=connection`).
2. Confirm the presence of the deprecated warnings for `JSONResponse.php`.
3. Visit `Marketing > Facebook Product Sets` (`/wp-admin/edit-tags.php?taxonomy=fb_product_set&post_type=product`) and confirm the deprecated warnings near the bottom of the `Add new Facebook Product Set` form.
4. Create a Product Set, then click its Edit link.
5. Confirm the deprecated warnings in the last form label and field.
6. Switch to this PR branch, and confirm that all the deprecated warnings have been resolved.

### DIFF
--- a/includes/Framework/Api/JSONResponse.php
+++ b/includes/Framework/Api/JSONResponse.php
@@ -80,7 +80,7 @@ abstract class JSONResponse implements Response, ArrayAccess {
 	 *
 	 * @return bool Whether the offset exists.
 	 */
-	public function offsetExists( $offset ) {
+	public function offsetExists( $offset ): bool {
 		return array_key_exists( $offset, $this->response_data );
 	}
 
@@ -93,7 +93,7 @@ abstract class JSONResponse implements Response, ArrayAccess {
 	 *
 	 * @return mixed The offset value.
 	 */
-	public function offsetGet( $offset ) {
+	public function offsetGet( $offset ): mixed {
 		return $this->response_data[ $offset ];
 	}
 
@@ -107,7 +107,7 @@ abstract class JSONResponse implements Response, ArrayAccess {
 	 *
 	 * @return void
 	 */
-	public function offsetSet( $offset, $value ) {
+	public function offsetSet( $offset, $value ): void {
 		$this->response_data[ $offset ] = $value;
 	}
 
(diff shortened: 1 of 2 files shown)


layoutd:
Thanks @ibndawood, good point regarding the method names. I'll merge these changes now and be on the lookout for your subsequent PR!

includes/Framework/Api/JSONResponse.php
```diff
@@ -93,7 +93,7 @@ public function offsetExists( $offset ) {
 	 *
 	 * @return mixed The offset value.
 	 */
-	public function offsetGet( $offset ) {
+	public function offsetGet( $offset ): mixed {
```
mikkamp, line 96: @layoutd I've been working on PHP 8.1 compatibility as well. But from my understanding the mixed return type was added in PHP 8.0. Looks like the minimum PHP version is still 7.2 here. Is that going to cause problems?

layoutd: Good catch, thanks @mikkamp. I looked over [this table](https://www.php.net/manual/en/language.types.declarations.php) while making the changes, that one snuck past me. Reverted in https://github.com/woocommerce/facebook-for-woocommerce/pull/2469. I'll have to loop back on a few of my recent changes to double-check them too.

PR #2479, opened 2023-02-16
Replace is_ajax calls with wp_doing_ajax()

rawdreeg:

is_ajax() has been deprecated since version 6.1.0.

1. Install facebook-for-woocommerce on WP 6.1.1 install and observe that no warning notices are displayed in the backend

### Changelog entry

> Fix - Replace is_ajax calls with wp_doing_ajax()

### DIFF
--- a/includes/Framework/Helper.php
+++ b/includes/Framework/Helper.php
@@ -406,7 +406,7 @@ class Helper {
 	 */
 	public static function trigger_error( $message, $type = E_USER_NOTICE ) {
 
-		if ( is_callable( 'is_ajax' ) && is_ajax() ) {
+		if ( is_callable( 'wp_doing_ajax' ) && wp_doing_ajax() ) {
 
 			switch ( $type ) {
 
--- a/includes/Framework/Helper.php
+++ b/includes/Framework/Helper.php
@@ -394,41 +394,4 @@ class Helper {
 		return (bool) apply_filters( 'woocommerce_is_rest_api_request', $is_rest_api_request );
 	}
 
-
-	/**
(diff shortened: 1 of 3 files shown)


includes/Framework/Helper.php
```diff
@@ -406,7 +406,7 @@ public static function is_rest_api_request() {
 	 */
 	public static function trigger_error( $message, $type = E_USER_NOTICE ) {
 
-		if ( is_callable( 'is_ajax' ) && is_ajax() ) {
```
layoutd, line 409: 💅🏻 Just curious if `is_callable` is still necessary here. Facebook for WooCommerce [requires WP 5.6+](https://github.com/woocommerce/facebook-for-woocommerce/blob/1e2d5407b24114706fe9eb8374b5f9e490c00dbf/facebook-for-woocommerce.php#L15) and `wp_doing_ajax` has been around since [4.7.0](https://github.com/WordPress/wordpress-develop/blob/a8fd670237e4983b08a2f92d492946d570d8238b/src/wp-includes/load.php#L1531-L1540). (Although it looks like `Helper::trigger_error` isn't actually used anywhere in the extension).

layoutd (approved):
LGTM 👍🏻

includes/Framework/Helper.php
rawdreeg: Yes. Good catch, @layoutd. I removed trigger_error altogether.

 It looks like the whole class was ported from SkyVerge. Only a few of the methods in the class are used. I created https://github.com/woocommerce/facebook-for-woocommerce/issues/2480 to clean this up.

PR #2517, opened 2023-04-11
Fix - Fatal error when facebook_config option is empty.

### LINKED ISSUE #2516: PHP 8.1 fatal error with Facebook Pixel Event
In checking a website with query monitor, a merchant found sound issues after upgrading to PHP 8.1.17, reported in 6082196-zen. The exact text of the error is available in 2fb1b-pb/#php

They found the following:

- The code causing the exception is: `$fb_options[ self::PIXEL_ID_KEY ] = $pixel_id;`
- The data it is trying to access with the $fb_options offset is there but it's being returned serialized and isn't being unserialized

They applied the following fix on their site in file facebook-commerce-pixel-event.php

line 536, method set_pixel_id()

// Throws error
$fb_options = self::get_options();

// No error
$fb_options = unserialize(self::get_options());

The suggest that maybe the data needs unserializing somewhere else and hence this report.

rawdreeg:

We received a report of a fatal error in the Pixel Event class. After investigation, I noticed that if the `facebook_config` option is an empty string, the value returned by get_options would be an empty string, causing the exception: `error message: Uncaught TypeError: Cannot access offset of type string on string`. 

This assumes that `facebook_config` will always contain a serialized object, and the error is, in fact, caused by the `facebook_config` option being empty. 

Closes #2516.

1. log in as an admin, and navigate to wp-admin/options.php then delete the content of facebook_config.
2. navigate to any admin page, and you should receive the fatal error.
3. Switch to this branch, and the fatal should be gone.

### Changelog entry

> Fix - Fatal error when facebook_config option is empty.

### DIFF
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -657,14 +657,11 @@ class WC_Facebookcommerce_Pixel {
 		 * Get PixelID related settings.
 		 */
 		public static function get_options() {
-			return get_option(
-				self::SETTINGS_KEY,
-				array(
-					self::PIXEL_ID_KEY     => '0',
-					self::USE_PII_KEY      => 0,
-					self::USE_S2S_KEY      => false,
-					self::ACCESS_TOKEN_KEY => '',
-				)
+			return get_option( self::SETTINGS_KEY ) ?: array(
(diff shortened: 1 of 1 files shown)


budzanowski (approved):
Nice! :shipit:

PR #2524, opened 2023-04-20
Fixes `semgrep` violations and phpcs `WordPress.Security` violations

ibndawood:
### Changes proposed in this Pull Request:

This PR:
* Fixes `semgrep` violations
* Fixes phpcs `WordPress.Security` violations.

- [x] Do the changed files pass `phpcs` checks? Please remove `phpcs:ignore` comments in changed files and fix any issues, or delete if not practical.

### Detailed test instructions:

1. Install `semgrep` using `brew install semgrep`
2. Clone this repository in a top-level folder: `git clone git@github.com:Automattic/wpscan-semgrep-rules.git`
3. Run `semgrep --no-git-ignore --text -c audit ../facebook-for-woocommerce` for `semgrep` results. The results should not return any findings.
4. Verify your PHP `memory_limit` is set to a very high number like for example `2048M`. This is required to process large number of files.
5. Open the plugin folder in the terminal and run `vendor/bin/phpcs --standard=WooCommerce-Core --sniffs=WordPress.Security.ValidatedSanitizedInput,WordPress.Security.EscapeOutput --ignore-annotations includes` for the includes folder
6. The results should some violations.
7. Verify that those violations are false positives by checking if the line is marked as `phpcs:ignore`. Since we are using the `--ignore-annotations` in step 2, we are going to see the ignored violations as well.
8. Verify there are no other violations.
9. Repeat steps 2-5 for the main folder by running `vendor/bin/phpcs --standard=WooCommerce-Core --sniffs=WordPress.Security.ValidatedSanitizedInput,WordPress.Security.EscapeOutput --ignore-annotations *.php`
10. Perform [Smoke Tests](https://github.com/woocommerce/facebook-for-woocommerce/wiki/Smoke-tests) to make sure there are no breaking changes.

### Changelog entry

> Fix - Adds escaping to outputs and sanitizing to inputs.

PR #2529, opened 2023-05-02
Removes esc_url from false positives

ibndawood:

PR #2524 adds `esc_url` to all `semgrep` violations. This could cause issues in redirect. `esc_url` should be added to `add_query_arg` only during output and can be declared as a false positive in other cases.

1. Install `semgrep` using `brew install semgrep`
2. Clone this repository in a top-level folder: `git clone git@github.com:Automattic/wpscan-semgrep-rules.git`
3. Run `semgrep --no-git-ignore --text -c audit ../facebook-for-woocommerce` for `semgrep` results. The results should not return any findings.
4. Verify your PHP `memory_limit` is set to a very high number like for example `2048M`. This is required to process large number of files.
5. Search for products in the WooCommerce admin panel. Verify that the search results are returned.

### Changelog entry

> Fix - Issue in admin product search

### DIFF
--- a/includes/Framework/AdminMessageHandler.php
+++ b/includes/Framework/AdminMessageHandler.php
@@ -370,7 +370,9 @@ class AdminMessageHandler {
 		if ( $this->set_messages() ) {
 			$location = add_query_arg( self::MESSAGE_ID_GET_NAME, $this->get_message_id(), $location );
 		}
-		return esc_url( $location );
+
+		// nosemgrep: audit.php.wp.security.xss.query-arg
+		return $location;
 	}
 
 


nima-karimi (approved):
Thanks for the quick fix, @ibndawood. Looks good. I removed the `esc_url` from the following line too:

https://github.com/woocommerce/facebook-for-woocommerce/commit/575433350c6ac03acac0069295aa18e8631dd502#diff-925917f4c0af59cbcb56ea7e21afaeee99c929edccdc5106125aafd8738a0d90R373

PR #2539, opened 2023-05-04
Register delete action hook outside of is_admin check.

### LINKED ISSUE #2526: Deleted products via the WooCommerce Rest API remain in the Facebook catalogue
## 🔎 Isolate the bug

## ✍️ Describe the bug

When deleting a WooCommerce product via the rest API it remains in Facebook's catalogue. I tried adding the `force` parameter to delete the product permanently but it had no effect.

EDIT: Mannually re-syncing or deleting all products from the catalogue via `Status > Tools` doesn't solve this, also, due to these products being uploaded by WooCommerce, they can't be deleted from Facebook's side.

Link to image: https://d.pr/i/lhToe3

From what I checked for example in the [product sets template](https://github.com/woocommerce/facebook-for-woocommerce/blob/6dbe0634fb56f5bca09db2c312800d8aedabac32/includes/ProductSets/Sync.php#L134),  there is no hook included for [when a product gets deleted via the API](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L498). 

## 🚶‍♀️ Steps to reproduce

Delete any product using any API framework with `DELETE /wp-json/wc/v3/products/<id>`

## ✔️ Expected behavior

The product to be removed just like when deleted it from `All products > Product`.

rawdreeg:

When deleting a WooCommerce product via the rest API it remains in Facebook's catalogue.  This is because the delete hook was only registered in the admin context. This PR moves these hooks outside the is_admin check.

Closes #2526 .

_Replace this with a good description of your changes & reasoning._

1. Delete any product using any API framework with DELETE /wp-json/wc/v3/products/<id>
2. The product should be deleted in the FB catalog as well.

### Changelog entry

> Fix - Delete product in the Facebook catalog when products are deleted via WC Rest API.

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -303,11 +303,6 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 					[ $this, 'on_quick_and_bulk_edit_save' ]
 				);
 
-				add_action( 'before_delete_post', [ $this, 'on_product_delete' ] );
-
-				// Ensure product is deleted from FB when moved to trash.
-				add_action( 'wp_trash_post', [ $this, 'on_product_delete' ] );
-
 				add_action( 'add_meta_boxes', 'WooCommerce\Facebook\Admin\Product_Sync_Meta_Box::register', 10, 1 );
 
 				add_action(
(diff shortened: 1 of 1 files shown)


budzanowski (approved):
Makes sense! :shipit:

PR #2566, opened 2023-06-28
Adds priority to enhanced catalog attributes and display them as per their priority

ibndawood:

The enhanced attributes in the product edit are not assigned a priority. This results in related attributes appearing randomly instead of together, as shown in the screenshot below.

image

It is expected that the product dimensions such as length, width, height, and depth should appear together. This PR loops through recommended attributes and assigns a `priority` index. We'll then sort the recommended attributes based on the `priority`.

### Screenshots:

After this PR:

image

1. Install, activate, and connect Facebook for WooCommerce.
2. Navigate to `Products > All Products` and click on `Edit` to open the product editor.
3. In the `Product data` metabox, choose the `Facebook` tab.
4. For `Google Product Category` field, choose a category that has measurements as attributes. A list of all  category and their respective attributes can be found here: https://github.com/woocommerce/facebook-for-woocommerce/blob/develop/bin/fb_google_category_to_attribute_mapping.json
5. Verify that the category specific fields are in order: Length, Width, Height, Depth, other measurement attributes and others. Please note that not all measure fields are available for all attributes. The order should be for whatever attributes are available.

### Additional details:

Related forum issue: https://wordpress.org/support/topic/ui-issues-2/

### Changelog entry

> Add - Priority to enhanced catalog attributes and display as per priority.

### DIFF
--- a/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
+++ b/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
@@ -90,10 +90,10 @@ class Enhanced_Catalog_Attribute_Fields {
 		 */
 		$priorities = array(
 			'product_length' => 140,
-			'product_width' => 130,
+			'product_width'  => 130,
 			'product_height' => 120,
-			'product_depth' => 110,
-			'default' => 100,
+			'product_depth'  => 110,
+			'default'        => 100,
 		);
 
 		foreach ( $recommended_attributes as $key => &$attribute ) {
--- a/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
+++ b/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
@@ -88,9 +88,9 @@ class Enhanced_Catalog_Attribute_Fields {
 		/**
 		 * Attributes should be ordered by priority in this order: Length, Width, Height, Depth, Other measurement attributes, others.
 		 */
-		foreach( $recommended_attributes as $key => $attribute ) {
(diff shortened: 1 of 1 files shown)


includes/Admin/Enhanced_Catalog_Attribute_Fields.php
```diff
@@ -85,6 +85,37 @@ function( $attr ) {
...
+				$recommended_attributes[ $key ]['priority'] = $attr_priority;
+			} else {
+				$recommended_attributes[ $key ]['priority'] = 5;
+			}
+
+			$priority[ $key ] = $recommended_attributes[ $key ]['priority'];
+		}
+
+		array_multisort( $priority, SORT_DESC, $recommended_attributes );
+
```
budzanowski, line 118:
```suggestion
$priorities = array(
    'product_length' => 140,
    'product_width' => 130,
    'product_height' => 120,
    'product_depth' => 110,
    'default' => 100,
);

foreach ( $recommended_attributes as $key => &$attribute ) {
    $attribute['priority'] = 5; // Assign 5 initially to each attribute
    if ( 'measurement' === $attribute['type'] ) {
        $attribute['priority'] = isset( $priorities[ $attribute['key'] ] ) ? $priorities[ $attribute['key'] ] : $priorities['default'];
    }
    $priority[ $key ] = $attribute['priority'];
}

array_multisort( $priority, SORT_DESC, $recommended_attributes );

```

What about something like this ^? Just to remove the `switch`. This is a personal preference so no need to go with my suggestion.

Approving.

budzanowski (approved):

PR #2569, opened 2023-06-29
Deletes the product from facebook once it is switched to draft.

### LINKED ISSUE #2547: Products remain on Facebook when set to draft
## ✍️ Describe the bug

Typically, when you publish a product it's synced to Facebook. When you set the same product to private or hidden, it's removed from Facebook. However, unlike when hidden and set to private, products remain on Facebook when set to draft. 

## 🚶‍♀️ Steps to reproduce

1. Sync products to Facebook as usual
2. Set the product to draft
3. Notice the product remains on Facebook while the Facebook sync widget returns "Product is not published"

## ✔️ Expected behavior

Published products should be removed from Facebook once set to draft

krutidugade:

Products are visible on Facebook even after switching to draft.  This PR deletes the product from Meta Commerce Manager once the product status is switched to draft. 

Closes #2547.

1. Install and activate the Facebook for WooCommerce plugin
2. Go to Meta Commerce Manager
3. Look for a specific product and note the ID
4. Go to WP-Admin > Products and edit the same product
5. Change the product status to draft and update
6. Look for the same product under Meta Commerce Manager
7. Verify the product isn't visible - compare product ID

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -368,13 +368,8 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		add_action( 'untrashed_post', [ $this, 'fb_restore_untrashed_variable_product' ] );
 
-		// Ensure product is deleted from FB when moved to draft.
-		add_action(
-			'transition_post_status',
-			array( $this, 'delete_draft_product' ),
-			10,
-			3
-		);
+		// Ensure product is deleted from FB when status is changed to draft.
+		add_action( 'publish_to_draft', [ $this, 'on_product_delete' ] );
 
 		// Product Set hooks.
 		add_action( 'fb_wc_product_set_sync', [ $this, 'create_or_update_product_set_item' ], 99, 2 );
@@ -933,25 +928,6 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		// clear out both item and group IDs
 		delete_post_meta( $product_id, self::FB_PRODUCT_ITEM_ID );
 		delete_post_meta( $product_id, self::FB_PRODUCT_GROUP_ID );
-	}
-
-	/**
-	 * Deletes a product from Facebook when status is changed to draft.
-	 *
-	 * @param string   $new_status
-	 * @param string   $old_status
-	 * @param \WP_post $post
-	 */
-	public function delete_draft_product( $new_status, $old_status, $post ) {
-
-		if ( ! $post ) {
-			return;
-		}
-
-		if ( $old_status === 'publish' && $new_status === 'draft' ) {
-			$this->on_product_delete ( $post->ID );
-		}
-
 	}
 
 	/**
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -368,6 +368,14 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		add_action( 'untrashed_post', [ $this, 'fb_restore_untrashed_variable_product' ] );
 
+		// Ensure product is deleted from FB when moved to draft.
+		add_action(
+			'transition_post_status',
+			array( $this, 'delete_draft_product' ),
+			10,
+			3
+		);
+
 		// Product Set hooks.
 		add_action( 'fb_wc_product_set_sync', [ $this, 'create_or_update_product_set_item' ], 99, 2 );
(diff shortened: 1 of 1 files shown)


facebook-commerce.php
```diff
@@ -927,6 +935,26 @@ public function delete_fb_product( $product ) {
...
+	 */
+	public function delete_draft_product( $new_status, $old_status, $post ) {
+
+		if ( ! $post ) {
+			return;
+		}
+
+		if ( $old_status === 'publish' && $new_status === 'draft' ) {
+			$product_id = $post->ID;
+			$this->on_product_delete ( $product_id );
```
rawdreeg, line 953:
```suggestion
			$this->on_product_delete ( $post->ID );
```

I think the variable is not needed. but that's my personal preference, you don't have to apply the change :-)

facebook-commerce.php
```diff
@@ -368,6 +368,14 @@ public function __construct( WC_Facebookcommerce $facebook_for_woocommerce ) {
 
 		add_action( 'untrashed_post', [ $this, 'fb_restore_untrashed_variable_product' ] );
 
+		// Ensure product is deleted from FB when moved to draft.
+		add_action(
+			'transition_post_status',
+			array( $this, 'delete_draft_product' ),
+			10,
+			3
+		);
```
rawdreeg, line 377: I noticed that a few lines above, we have this action hook: https://github.com/woocommerce/facebook-for-woocommerce/blob/cccb9f88682c57166fa1b6709f19cc987fea7946/facebook-commerce.php#L367

They're doing the same thing. I'd remove that if it's not working, and I am not sure there's value in keeping both codes.

rawdreeg (approved):
Approved, with a couple of minor comments for your consideration.

budzanowski:
If you look inside the `./src/wp-includes/post.php` you will find that you can use this hook:

https://github.com/WordPress/wordpress-develop/blob/bea4307daa21a3f97d159898b0ac676332e0e7de/src/wp-includes/post.php#L5363

Which uses a dynamic action name. This means that you can use a more direct action `publish_to_draft` and you are not required to do the check:

```PHP
if ( $old_status === 'publish' && $new_status === 'draft' ) 
```

facebook-commerce.php
budzanowski: I don't understand this comment 😓 

One hook is for deleting, one if for trashing and one is for changing to draft :) Which one should we delete? Do I understand this correctly?

facebook-commerce.php
budzanowski: I think that this would work with just `$post`.

Because the first line of `on_product_detele` is:

```PHP
	$product = wc_get_product( $product_id );
``` 

and the first param to `wc_get_product` is described as:
```PHP
 * @param mixed $the_product Post object or post ID of the product.
```

facebook-commerce.php
krutidugade: We need to keep both. The new action hook only deletes the product if it is changed to draft. The existing one `add_action( 'wp_trash_post', [ $this, 'on_product_delete' ] ); ` deletes the product if it is moved to trash.

rawdreeg: Had a bit of a brain fart here. Draft and trash aren't the same. You're both correct. I will resolve the threat.

facebook-commerce.php
krutidugade: @budzanowski Are you suggesting we make changes to `on_product_detele` method? If not then we'd get an error since the method accepts int parameter.

budzanowski: We could but lets keep it as is and just use what @rawdreeg has suggested.

krutidugade:
@rawdreeg I found an issue with the merged code so requested a revert. I have pushed the correct code in the same branch - https://github.com/woocommerce/facebook-for-woocommerce/commit/21ae4192fde8e327510baf2021ccc2c1cd39f07a

budzanowski:
@krutidugade In cases like this please do the revert ( if it is really just a revert change it should be OK to bypass the security review check and just merge ).

Next please create a separate PR with the correct fix.

krutidugade:
> @krutidugade In cases like this please do the revert ( if it is really just a revert change it should be OK to bypass the security review check and just merge ).

Thanks for confirming. 

> Next please create a separate PR with the correct fix.
On it!

PR #2597, opened 2023-08-02
Adds a filter to Business Manager ID

ibndawood:

It would be useful to have a filter that will allow third-party plugins to override the Facebook Business Manager ID. This will be useful when overriding the connection to establish a manual connection,

### Changelog entry

> Dev - Filters Business Manager ID.


ibndawood:
Thank you, @rawdreeg, for your time to review the PR.

PR #2600, opened 2023-08-03
Use the Dashicon FB icon in product data options tabs

layoutd:

During smoke testing I noticed the icon in Product Data could be more descriptive, so this PR changes it from the generic wrench to the [existing Dashicons Facebook alt logo](https://developer.wordpress.org/resource/dashicons/#facebook). This is especially useful in reduced/mobile resolutions.

### Screenshots:

**Before (Desktop/Mobile)**
image  image

**After (Desktop/Mobile)**
image image

1. Activate WooCommerce and Facebook for WooCommerce.
2. Behold! (may require a hard refresh).

### Changelog entry

> Tweak - Update product data options icon.

### DIFF
--- a/assets/css/admin/facebook-for-woocommerce-products-admin.css
+++ b/assets/css/admin/facebook-for-woocommerce-products-admin.css
@@ -6,6 +6,10 @@
 	width: auto;
 }
 
+#woocommerce-product-data ul.wc-tabs li.fb_commerce_tab_options a::before {
+	content: "\f304";
+}
+
 /* added #woocommerce-product-data to increase selector specificity */
 #woocommerce-product-data .fb-product-image-source-field .wc-radios li {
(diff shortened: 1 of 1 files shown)


ibndawood (approved):
Thank you, @layoutd, for the changes. The tab now looks much better. 

I notice that the alt Facebook logo is `305` and the regular squared logo is `304`. Did you intend to use the regular or alt? I am want to confirm because the PR desc says alt logo and maybe the alt logo would look better than the squared logo.

layoutd:
You're right, I ended up not using alt because the original seemed better. Take a look (modify the `::before` element). What do you think? On second look, the alt may mesh better with other channel icons (like Google).

ibndawood:
I agree. The squared one looks better than the alt one.

PR #2603, opened 2023-08-03
make get_options a bit more robust.

rawdreeg:

We assume that if the serial object is “correct”, meaning an array, then we return it to the caller to check if the index is there. This PR will set an index to the default when such index does not exist in the config

1. Switch to this branch.
2. log in as an admin, navigate to wp-admin/options.php, then set the value of the content of facebook_config to a random string. e.g: 10k
3. navigate to any admin page. Then check the value of facebook_config and confirm this is a valid serialized object. And repeat this test by setting facebook_config to an empty value.

### Changelog entry

> Fix - WC_Facebookcommerce_Pixel::get_options() throwing a fatal error if facebook_config is invalid

### DIFF
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -657,15 +657,27 @@ class WC_Facebookcommerce_Pixel {
 		 * Get PixelID related settings.
 		 */
 		public static function get_options() {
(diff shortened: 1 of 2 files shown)


ibndawood (approved):
Thank you, @rawdreeg, for the changes. I tidied up the spacing a bit in the unit test case file.

PR #2612, opened 2023-08-08
Fix WPML product hiding logic

### LINKED ISSUE #2584: WPML + Facebook for WooCommerce can trigger a fatal error on PHP 8+
## 🔎 Isolate the bug

## ✍️ Describe the bug

As described in [this support request](https://wordpress.org/support/topic/backend-fatal-error-php-fatal-error-uncaught-typeerror-illegal-offset-type/), there are circumstances where the `WC_Facebook_WPML_Injector::should_hide()` method generates a fatal error.

Looking in the code for this method, we're relying on the results of calling:

```php
$product_lang = apply_filters( 'wpml_post_language_details', null, $wp_id );
```

In WPML, I found that the `wpml_get_language_information()` function is hooked to this filter, and it has two possible return values:

1. An instance of `WP_Error`. This happens when the post ID is not provided, or when the post object cannot be retrieved using `get_post()`. 
2. An array, like this:

```php
	return [
		'language_code'      => $language,
		'locale'             => $sitepress->get_locale( $language ),
		'text_direction'     => $sitepress->is_rtl( $language ),
		'display_name'       => $sitepress->get_display_language_name( $language, $current_language ),
		'native_name'        => isset( $language_information['display_name'] ) ? $language_information['display_name'] : '',
		'different_language' => $language !== $current_language,
	];
```

Considering that the default value of `null` is passed to the filter, it is also possible that the return value is `null`.

## Discussion needed

The main question is this: what should the `should_hid()` method return in each of the above scenarios?

## 🚶‍♀️ Steps to reproduce

Pending response to [this comment](https://wordpress.org/support/topic/backend-fatal-error-php-fatal-error-uncaught-typeerror-illegal-offset-type/#post-16901967)

## 🗃 Logs

The status report can be seen in the WordPress.org forum thread above.

JPry:

Closes #2584.

This updates the logic within the `\WC_Facebook_WPML_Injector::should_hide()` method to more robustly handle various results that could be returned from the applied filter. It also adds unit tests covering this functionality.

### Changelog entry

> Fix - Update WPML compatibility to prevent fatal errors on PHP 8+

### DIFF
--- a/includes/fbwpml.php
+++ b/includes/fbwpml.php
@@ -87,6 +87,7 @@ class WC_Facebook_WPML_Injector {
 	}
 
 	public function wpml_ajax_support( $call, $REQUEST ) {
+		/** @var SitePress $sitepress */
 		global $sitepress;
 		if ( isset( $REQUEST['icl_ajx_action'] ) ) {
 			$call = $REQUEST['icl_ajx_action'];
@@ -111,7 +112,7 @@ class WC_Facebook_WPML_Injector {
 	 * The section is shown at the bottom of the WPML > Languages settings page.
 	 */
 	public function wpml_support() {
-		/** @var object $sitepress */
+		/** @var SitePress $sitepress */
 		global $sitepress;
 
 		// there is no nonce to check here and the value of $_GET['page] is being compared against a known and safe string
--- a/includes/fbwpml.php
+++ b/includes/fbwpml.php
@@ -1,5 +1,4 @@
 <?php
-// phpcs:ignoreFile
 /**
  * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
  *
@@ -7,6 +6,11 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @package FacebookCommerce
+ *
+ * phpcs:disable Squiz.Commenting.ClassComment.Missing
+ * phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
+ * phpcs:disable Squiz.Commenting.VariableComment.Missing
+ * phpcs:disable WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -82,7 +86,7 @@ class WC_Facebook_WPML_Injector {
 		}
 
 		// Hide products from non-active languages, of if the language status isn't visible.
-		$should_hide = ! isset( self::$settings[ $language_code ] ) || self::$settings[ $language_code ] !== FB_WPML_Language_Status::VISIBLE;
+		$should_hide = ! isset( self::$settings[ $language_code ] ) || FB_WPML_Language_Status::VISIBLE !== self::$settings[ $language_code ];
 
 		return (bool) apply_filters( 'wc_facebook_wpml_should_hide_product', $should_hide, $product_id, $product_lang );
 	}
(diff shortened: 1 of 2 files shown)


includes/fbwpml.php
```diff
@@ -37,26 +37,57 @@ public function __construct() {
...
+	 * Whether a product should be hidden from the Facebook catalog.
+	 *
+	 * @since 1.9.10
+	 *
+	 * @param int $wp_id The product ID.
+	 *
+	 * @return bool
+	 */
 	public static function should_hide( $wp_id ) {
+		// Apply WPML filters to the product.
```
ibndawood, line 50: `phpcs` throws this error: 

```
A "hook" comment must be a "/**" style docblock comment.
```

includes/fbwpml.php
```diff
@@ -37,26 +37,57 @@ public function __construct() {
...
+		$language_code = $product_lang['language_code'] ?? '';
+
+		// Option doesn't exist: Backwards Compatibility
+		if ( ! self::$settings ) {
+			return (bool) apply_filters( 'wc_facebook_wpml_should_hide_product', self::$default_lang !== $language_code, $wp_id, $product_lang );
 		}
-		return $settings[ $product_lang ] !== FB_WPML_Language_Status::VISIBLE;
+
+		// Hide products from non-active languages, of if the language status isn't visible.
+		$should_hide = ! isset( self::$settings[ $language_code ] ) || self::$settings[ $language_code ] !== FB_WPML_Language_Status::VISIBLE;
```
ibndawood, line 84: Another `phpcs` error:

```
Use Yoda Condition checks, you must.
```

includes/fbwpml.php
```diff
@@ -37,26 +37,57 @@ public function __construct() {
...
-			return ( $product_lang && self::$default_lang !== $product_lang );
+		/*
+		 * The wpml_post_language_details filter is applied to obtain the language data for the product.
+		 *
+		 * Possible values for $product_lang after the filter is applied:
+		 *
+		 * 1. null - The WPML filter wasn't applied.
+		 * 2. array - WPML returned the language data from wpml_get_language_information().
+		 * 3. WP_Error - WPML didn't recognize the post ID.
+		 */
```
ibndawood, line 61: We can move this docblock comment to the `wpml_post_language_details` filter.

tests/Unit/WPMLTest.php
```diff
@@ -0,0 +1,95 @@
...
+	 * @return void
+	 */
+	public function tear_down() {
+		WC_Facebook_WPML_Injector::$settings     = null;
+		WC_Facebook_WPML_Injector::$default_lang = null;
+		remove_all_filters( 'wpml_post_language_details' );
+	}
+
+	public function test_should_hide_product_when_wpml_filter_not_applied() {
+		$this->assertTrue( WC_Facebook_WPML_Injector::should_hide( 1 ) );
```
ibndawood, line 19: A small suggestion for better readability. 
```suggestion
        $product_id = 1;
        $this->assertTrue( WC_Facebook_WPML_Injector::should_hide( $product_id ) );
```

ibndawood: If you agree with this suggestion, you may apply it for the rest of the test cases.

ibndawood (changes requested):
Thank you, @JPry, for the changes. Excellent work on the unit tests. I've added a couple of comments regarding `phpcs` errors and a suggestion which is a take it or leave it.

ibndawood:
One additional comment is that I'd suggest renaming the method parameter from `$wp_id` to `$product_id` as it would improve the readability of the code.

tests/Unit/WPMLTest.php
JPry: Thanks for the suggestion, I did it a little bit differently in 66922dc0.

JPry:
@ibndawood Thanks for the review and the suggestions! I've made some updates based on your feedback.

ibndawood (approved):
Thank you, @JPry, for the changes. They look good to me and tests well with WPML.

PR #2624, opened 2023-08-23
Updates PHP unit tests matrix and dev dependencies for PHP8.2 compatibility.

krutidugade:

As part of PHP 8.1+ compatibility testing, we are updating unit test tools to the latest version of PHP. This PR updates the php-unit-tests.yml to run unit tests for PHP 7.4, 8.0, 8.1, and 8.2. We have bumped the PHP requirement to 7.4 and phpunit to 8.5. Also, we have updated yoast/phpunit-polyfills dependency to 2.0.

No code change. Verify all the git workflow unit test checks pass.

### Changelog entry

> Dev - Updates PHP unit tests matrix in git workflow and versions of dev dependencies in composer.

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -30,11 +30,11 @@ jobs:
       WP_TESTS_DIR: "/tmp/wordpress/tests/phpunit"
     strategy:
       matrix:
-        php: [7.4]
-        wp-version: [5.8, 5.9, latest]
+        php: [7.4, 8.0, 8.1, 8.2]
+        wp-version: [latest]
         include:
-          - php: 8.0
-            wp-version: latest
+          - php: 7.4
+            wp-version: 5.9
 
     steps:
       - name: Checkout repository
@@ -50,10 +50,6 @@ jobs:
 
       - name: Install WP tests
         run: ./bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wp-version }}
-
-      - name: PHP8, PHPUnit9 with WordPress v5.9+
-        if: matrix.php >= 8
-        run: composer require phpunit/phpunit "^8" --dev --update-with-all-dependencies
 
       - name: Run PHP unit tests
         run: composer test-unit
--- a/composer.json
+++ b/composer.json
@@ -10,15 +10,15 @@
     }
   ],
   "require": {
-    "php": ">=7.2",
+    "php": ">=7.4",
     "woocommerce/action-scheduler-job-framework": "2.0.0",
     "composer/installers": "~1.0"
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^v0.7",
-    "phpunit/phpunit": "^7.5",
+    "phpunit/phpunit": "^8.5",
     "wp-coding-standards/wpcs": "^2.3",
-    "yoast/phpunit-polyfills": "^1.0",
+    "yoast/phpunit-polyfills": "^2.0",
     "wp-cli/i18n-command": "^2.2",
     "woocommerce/woocommerce-sniffs": "^0.1.3"
   },
@@ -34,7 +34,7 @@
   },
   "config": {
     "platform": {
-      "php": "7.2"
+      "php": "7.4"
     },
     "allow-plugins": {
       "dealerdirect/phpcodesniffer-composer-installer": true,
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ee8b2ef17c17be9ccf196202955701af",
+    "content-hash": "aab7b7c7b053e27342c18225c8446b03",
     "packages": [
         {
             "name": "composer/installers",
@@ -405,16 +405,16 @@
         },
         {
             "name": "gettext/gettext",
-            "version": "v4.8.8",
+            "version": "v4.8.11",
             "source": {
                 "type": "git",
                 "url": "https://github.com/php-gettext/Gettext.git",
-                "reference": "302a00aa9d6762c92c884d879c15d3ed05d6a37d"
+                "reference": "b632aaf5e4579d0b2ae8bc61785e238bff4c5156"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/php-gettext/Gettext/zipball/302a00aa9d6762c92c884d879c15d3ed05d6a37d",
-                "reference": "302a00aa9d6762c92c884d879c15d3ed05d6a37d",
+                "url": "https://api.github.com/repos/php-gettext/Gettext/zipball/b632aaf5e4579d0b2ae8bc61785e238bff4c5156",
+                "reference": "b632aaf5e4579d0b2ae8bc61785e238bff4c5156",
                 "shasum": ""
             },
             "require": {
@@ -466,7 +466,7 @@
             "support": {
                 "email": "oom@oscarotero.com",
                 "issues": "https://github.com/oscarotero/Gettext/issues",
-                "source": "https://github.com/php-gettext/Gettext/tree/v4.8.8"
+                "source": "https://github.com/php-gettext/Gettext/tree/v4.8.11"
             },
             "funding": [
                 {
@@ -482,7 +482,7 @@
                     "type": "patreon"
                 }
             ],
-            "time": "2022-12-08T11:59:50+00:00"
+            "time": "2023-08-14T15:15:05+00:00"
         },
         {
             "name": "gettext/languages",
@@ -560,16 +560,16 @@
         },
         {
             "name": "mck89/peast",
-            "version": "v1.15.1",
+            "version": "v1.15.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/mck89/peast.git",
-                "reference": "cf06286910b7efc9dce7503553ebee314df3d3d3"
+                "reference": "1df4dc28a6b5bb7ab117ab073c1712256e954e18"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/mck89/peast/zipball/cf06286910b7efc9dce7503553ebee314df3d3d3",
-                "reference": "cf06286910b7efc9dce7503553ebee314df3d3d3",
+                "url": "https://api.github.com/repos/mck89/peast/zipball/1df4dc28a6b5bb7ab117ab073c1712256e954e18",
+                "reference": "1df4dc28a6b5bb7ab117ab073c1712256e954e18",
                 "shasum": ""
             },
             "require": {
@@ -582,13 +582,12 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "1.15.1-dev"
+                    "dev-master": "1.15.4-dev"
                 }
             },
             "autoload": {
                 "psr-4": {
-                    "Peast\\": "lib/Peast/",
-                    "Peast\\test\\": "test/Peast/"
+                    "Peast\\": "lib/Peast/"
                 }
             },
             "notification-url": "https://packagist.org/downloads/",
@@ -604,9 +603,9 @@
             "description": "Peast is PHP library that generates AST for JavaScript code",
             "support": {
                 "issues": "https://github.com/mck89/peast/issues",
-                "source": "https://github.com/mck89/peast/tree/v1.15.1"
+                "source": "https://github.com/mck89/peast/tree/v1.15.4"
             },
-            "time": "2023-01-21T13:18:17+00:00"
+            "time": "2023-08-12T08:29:29+00:00"
         },
         {
             "name": "mustache/mustache",
@@ -660,16 +659,16 @@
         },
         {
             "name": "myclabs/deep-copy",
-            "version": "1.11.0",
+            "version": "1.11.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/myclabs/DeepCopy.git",
-                "reference": "14daed4296fae74d9e3201d2c4925d1acb7aa614"
+                "reference": "7284c22080590fb39f2ffa3e9057f10a4ddd0e0c"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/myclabs/DeepCopy/zipball/14daed4296fae74d9e3201d2c4925d1acb7aa614",
-                "reference": "14daed4296fae74d9e3201d2c4925d1acb7aa614",
+                "url": "https://api.github.com/repos/myclabs/DeepCopy/zipball/7284c22080590fb39f2ffa3e9057f10a4ddd0e0c",
+                "reference": "7284c22080590fb39f2ffa3e9057f10a4ddd0e0c",
                 "shasum": ""
             },
             "require": {
@@ -707,7 +706,7 @@
             ],
             "support": {
                 "issues": "https://github.com/myclabs/DeepCopy/issues",
-                "source": "https://github.com/myclabs/DeepCopy/tree/1.11.0"
+                "source": "https://github.com/myclabs/DeepCopy/tree/1.11.1"
             },
             "funding": [
                 {
@@ -715,32 +714,33 @@
                     "type": "tidelift"
                 }
             ],
-            "time": "2022-03-03T13:19:32+00:00"
+            "time": "2023-03-08T13:26:56+00:00"
         },
         {
             "name": "phar-io/manifest",
-            "version": "1.0.3",
+            "version": "2.0.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phar-io/manifest.git",
-                "reference": "7761fcacf03b4d4f16e7ccb606d4879ca431fcf4"
+                "reference": "97803eca37d319dfa7826cc2437fc020857acb53"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phar-io/manifest/zipball/7761fcacf03b4d4f16e7ccb606d4879ca431fcf4",
-                "reference": "7761fcacf03b4d4f16e7ccb606d4879ca431fcf4",
+                "url": "https://api.github.com/repos/phar-io/manifest/zipball/97803eca37d319dfa7826cc2437fc020857acb53",
+                "reference": "97803eca37d319dfa7826cc2437fc020857acb53",
                 "shasum": ""
             },
             "require": {
                 "ext-dom": "*",
                 "ext-phar": "*",
-                "phar-io/version": "^2.0",
-                "php": "^5.6 || ^7.0"
+                "ext-xmlwriter": "*",
+                "phar-io/version": "^3.0.1",
+                "php": "^7.2 || ^8.0"
             },
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "1.0.x-dev"
+                    "dev-master": "2.0.x-dev"
                 }
             },
             "autoload": {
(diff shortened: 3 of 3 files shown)


.github/workflows/php-unit-tests.yml
```diff
@@ -30,11 +30,11 @@ jobs:
...
       matrix:
-        php: [7.4]
-        wp-version: [5.8, 5.9, latest]
+        php: [7.4, 8.0, 8.1, 8.2]
+        wp-version: [latest]
         include:
-          - php: 8.0
-            wp-version: latest
+          - php: 7.4
+            wp-version: 5.9
```
ecgan, line 37: ❓ Why is this 5.9? The latest WP version is 6.3. If let's say we follow "L-2" policy, then the version here should be 6.1, no? 🤔

ecgan (changes requested):
The changes here look okay. I have one comment about the WP version being set to 5.9, which I have marked with ❓ .

I did some testing and code search on this PR. Here are my findings. Important ones are marked with ❓ . I think we need to make a few more changes in this PR. 

## `npm run lint:php` and `npm run lint:php:summary`

With PHP 7.4, I see 0 errors and 3 warnings:

With PHP 8.2, I see there are 98 errors and 0 warnings, and they are mostly about `"The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated"`.

A reference on how to fix the error: [Migration to PHP 8.1 - how to fix Deprecated Passing null to parameter error - rename build in functions](https://stackoverflow.com/q/71707325/379641)

📜 I think we can fix this is a separate PR, and also put this linting into GitHub Actions, which we don't have now.

## `composer test-unit`

I also tried to run the unit test locally, but apparently it is not that straightforward as it requires some installation step with `./bin/install-wp-tests.sh` with database access. It makes me wonder why _unit test_ requires _database access_. 🤔 I think it would be good to have more documentation on this, and also add an npm script so that we can have all the scripts listed in `package.json` file.  
 
## `npm run i18n`

PHP 7.4: OK.
PHP 8.2: OK.

## `npm run generate:category_attribute_json`

PHP 7.4: OK.
PHP 8.2: OK.

## `npm run archive`

PHP 7.4: OK.
PHP 8.2: OK.

## PHP version find

I did some PHP version search in the code base.

### "7.2"

I think this line can be removed, because the WooCommerce core plugin already has PHP version specified and our extension should just follow it (related discussion: pb0Spc-4fD-p2#comment-8238):

https://github.com/woocommerce/facebook-for-woocommerce/blob/0f6298cbc6a38fdeaa4a5eb5252166dc4cbd2ce1/facebook-for-woocommerce.php#L20

The following variable and its usage should be removed too, this would probably require a bit more testing, so we can do it in a separate PR:

https://github.com/woocommerce/facebook-for-woocommerce/blob/0f6298cbc6a38fdeaa4a5eb5252166dc4cbd2ce1/facebook-for-woocommerce.php#L50-L51

❓ The following should be updated to "7.4" (reference: https://github.com/PHPCompatibility/PHPCompatibility#sniffing-your-code-for-compatibility-with-specific-php-versions):

https://github.com/woocommerce/facebook-for-woocommerce/blob/0f6298cbc6a38fdeaa4a5eb5252166dc4cbd2ce1/phpcs.xml.dist#L7

❓ In the following PHP coding standards configuration, we are using PHP 7.2, and this can be seen in the GitHub Action run too. It should be updated to "7.4" (see https://github.com/StephaneBour/actions-php-lint); I wonder if we can put 7.4 and 8.2 into a matrix configuration, perhaps in a separate PR 🤔 :

https://github.com/woocommerce/facebook-for-woocommerce/blob/0f6298cbc6a38fdeaa4a5eb5252166dc4cbd2ce1/.github/workflows/php-coding-standards.yml#L40

.github/workflows/php-unit-tests.yml
krutidugade: The idea was to preserve one old check PHP7.4 + WP5.9 from the previous setup. So if any new code doesn't work with this setup, we would know what is going wrong.

>WooCommerce now uses L-1 support policy.
I think it is fine to use latest wp-version, so that we can catch issues early, instead of using older wp-version.

Given that supporting older versions isn't within our scope and everything works fine up until this point, we are good to remove these and just have the latest WordPress as you suggest.

composer.json
```diff
@@ -10,15 +10,15 @@
...
   "require": {
-    "php": ">=7.2",
+    "php": ">=7.4",
     "woocommerce/action-scheduler-job-framework": "2.0.0",
     "composer/installers": "~1.0"
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^v0.7",
-    "phpunit/phpunit": "^7.5",
+    "phpunit/phpunit": "^8.5",
```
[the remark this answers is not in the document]
krutidugade: Nice catch. I updated PHPUnit with `composer require phpunit/phpunit --dev --update-with-all-dependencies` and it selected 8.5 automatically. For PHP 8.1+, I made changes to php-unit-test.yml file to have the workflow update PHPUnit to ^10 which can be seen in this [test PR commit](https://github.com/woocommerce/facebook-for-woocommerce/pull/2623/commits/e4ddad40781d97fbf195ccf9947265c7a7c0b5e5#diff-626cf1025fab6d0ecd9b6b485d250f06348b4893aae8b6ad792ac6b409000490). Eventually, I ended up removing this and didn't test with ^9. I'm going to test and make changes accordingly. Thanks for bringing this to my notice.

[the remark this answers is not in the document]
krutidugade: Fixed in [this commit](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/b8f3cb9e3a5be2080349bdfef0a135cfac1f0403).

.github/workflows/php-unit-tests.yml
krutidugade: Fixed in [this commit](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/b8f3cb9e3a5be2080349bdfef0a135cfac1f0403).

krutidugade:
>I did some PHP version search in the code base.
"7.2"
Requires PHP: 7.2 
const MINIMUM_PHP_VERSION = '7.2.0'; 

Thank you for the suggestion. I have made the changes in [this commit](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/e4d119087c213acfcc7e234344cc66fdf4ded8b3). I'll be working through the rest of your feedback shortly.

krutidugade:
>The following should be updated to "7.4" (reference: https://github.com/PHPCompatibility/PHPCompatibility#sniffing-your-code-for-compatibility-with-specific-php-versions):
 <config name="testVersion" value="7.2-" /> 
 
✅  I have updated this to 7.4-.

>With PHP 8.2, I see there are 98 errors and 0 warnings, and they are mostly about "The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated".
A reference on how to fix the error: [Migration to PHP 8.1 - how to fix Deprecated Passing null to parameter error - rename build in functions](https://stackoverflow.com/q/71707325/379641)
📜 I think we can fix this is a separate PR, and also put this linting into GitHub Actions, which we do not have currently.

I did some digging on this. We get a bunch of errors but they are just repetitions of the following two errors.
```
An error occurred during processing; checking has been aborted. The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /abc/wp-content/plugins/facebook-for-woocommerce/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php on line 280 (Internal.Exception)

An error occurred during processing; checking has been aborted. The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /abc/wp-content/plugins/facebook-for-woocommerce/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WP/I18nSniff.php on line 194 (Internal.Exception)
```

These errors come from wp-coding-standards/wpcs. I tried updating wp-coding-standards/wpcs dev dependency to 3.0.0 hoping it will fix these errors given that the [code is refactored](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/I18nSniff.php) in the latest release. However, we can't update it as `"wp-coding-standards/wpcs": "^2.3"` is a requirement for `"woocommerce/woocommerce-sniffs": "^0.1.3"`. It's been more than a year since woocommerce-sniffs was updated. I'm wondering what should we do next. Should we figure out a way to display a notice to add `?? ''`  at specific locations for PHP versions >=8.1 to be able to run unit tests locally? @ecgan 

> In the following PHP coding standards configuration, we are using PHP 7.2, and this can be seen in the GitHub Action run too. It should be updated to "7.4" (see https://github.com/StephaneBour/actions-php-lint); I wonder if we can put 7.4 and 8.2 into a matrix configuration, perhaps in a separate PR 🤔 :

Because of the reason mentioned above, I have updated it to 8.0 `StephaneBour/actions-php-lint@8.0`.

krutidugade:
>I also tried to run the unit test locally, but apparently it is not that straightforward as it requires some installation step with ./bin/install-wp-tests.sh with database access. It makes me wonder why unit test requires database access. 🤔 I think it would be good to have more documentation on this, and also add an npm script so that we can have all the scripts listed in package.json file.

That's a good suggestion. I have added the script in package.json and updated [README.md](https://github.com/woocommerce/facebook-for-woocommerce#readme) file in [this commit](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/64d88cf92764eacd50f125f6641bedaa1d042b56).

krutidugade:
>What I meant to say is that if it requires database access, then it is more of an integration test than a unit test. Even the title of the page that you linked to is called "Plugin Integration Tests". 😂 But anyway, I'll just ignore the naming of it now, don't want to get too caught up by it.

That makes sense. I'm still learning this whole testing thing so it is interesting to read about this bit. Thanks for highlighting it.

ecgan:
> These errors come from wp-coding-standards/wpcs. I tried updating wp-coding-standards/wpcs dev dependency to 3.0.0 hoping it will fix these errors given that the [code is refactored](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/I18nSniff.php?rgh-link-date=2023-08-25T10%3A51%3A23Z) in the latest release. However, we can't update it as `"wp-coding-standards/wpcs": "^2.3"` is a requirement for `"woocommerce/woocommerce-sniffs": "^0.1.3"`. It's been more than a year since [woocommerce-sniffs was updated](https://github.com/woocommerce/woocommerce-sniffs). I'm wondering what should we do next. Should we figure out a way to display a notice to add `?? ''` at specific locations for PHP versions >=8.1 to be able to run unit tests locally? @ecgan

I did a some more testings on this. At first, I thought it is a problem with `woocommerce/woocommerce-sniffs`, so I uninstalled the package, remove the related PHPCS configuration and rerun PHPCS in PHP 8.2, but I still see the same problem.

I did some googling on the error message and I found this issue https://github.com/WordPress/WordPress-Coding-Standards/issues/2035. In short, `"wp-coding-standards/wpcs": "^2.3"` only works with PHP 7.4 and does not work with PHP 8.0+. To make it work with PHP 8.0+, we need to use `"wp-coding-standards/wpcs": "^3.0"`. 

There is also an alternative, which is to put in the following into phpcs.xml configuration file, to suppress reporting on deprecated error:

```xml
<!--
Prevent errors caused by WordPress Coding Standards not supporting PHP 8.0+.
See https://github.com/WordPress/WordPress-Coding-Standards/issues/2035
-->
<ini name="error_reporting" value="E_ALL &#38; ~E_DEPRECATED" />
```

I tested it and it seems to work as expected, but I'm not sure if it will give us trouble down the road, e.g. suppressing deprecated error in our own plugin code, not just in WordPress Coding Standards.

So, I think what we should do is that when we run phpcs / `npm run lint:php`, we should only use PHP 7.4 (we should mention this in the documentation). When WooCommerce core moves to using minimum PHP version 8.0+, hopefully by then woocommerce-sniffs will be updated to work with WPCS 3.0, and then we shall update the packages in our plugins to use the newer woocommerce-sniffs and WPCS.

Side note: I also look into the PHP Coding Standards GitHub Action that runs PHPCS. The PHP version used in the GitHub Action depends on the `woocommerce/grow/prepare-php@actions-v1`: 

https://github.com/woocommerce/facebook-for-woocommerce/blob/0f6298cbc6a38fdeaa4a5eb5252166dc4cbd2ce1/.github/workflows/php-coding-standards.yml#L28-L29

... which currently has a default value of "7.4" (see https://github.com/woocommerce/grow/blob/9ea43833593986daeb1edbaa54404acacb62527b/packages/github-actions/actions/prepare-php/action.yml#L8). That's why it runs okay with no errors.

ecgan:
> > In the following PHP coding standards configuration, we are using PHP 7.2, and this can be seen in the GitHub Action run too. It should be updated to "7.4" (see [StephaneBour/actions-php-lint](https://github.com/StephaneBour/actions-php-lint)); I wonder if we can put 7.4 and 8.2 into a matrix configuration, perhaps in a separate PR 🤔 :
> 
> _Because of the reason mentioned above_, I have updated it to 8.0 `StephaneBour/actions-php-lint@8.0`.

Sorry, I don't quite get it, but what do you mean by "the reason mentioned above"? Why do we use 8.0? 🤔

krutidugade:
>Then, I updated wp-coding-standards/wpcs to version 3.0, and it works with no errors.

Amazing troubleshooting, well done @ecgan . You confirmed my assumption that the errors won't appear with wp-coding-standards/wpcs 3.0.

>In short, "wp-coding-standards/wpcs": "^2.3" only works with PHP 7.4 and does not work with PHP 8.0+
Sorry, I don't quite get it, but what do you mean by "the reason mentioned above"? Why do we use 8.0? 🤔

I beg to differ. I have tested it with PHP version 8.0 and I don't get the errors. This can be confirmed in this [PHPCS job](https://github.com/woocommerce/facebook-for-woocommerce/actions/runs/5976688578/job/16215064185?pr=2624) where PHP8.0 is installed and the check has passed. This is why I have used `StephaneBour/actions-php-lint@8.0` and explicitly mentioned PHP version 8.0. We get the errors for PHP8.1+ versions.

>There is also an alternative, which is to put in the following into phpcs.xml configuration file, to suppress reporting on deprecated error:
I tested it and it seems to work as expected, but I'm not sure if it will give us trouble down the road, e.g. suppressing deprecated error in our own plugin code, not just in WordPress Coding Standards.

I'm not in favor of suppressing warnings. We may end up not fixing these warnings for a long time so I would rather have these visible until they are fixed.

>So, I think what we should do is that when we run phpcs / npm run lint:php, we should only use PHP 7.4 (we should mention this in the documentation). When WooCommerce core moves to using minimum PHP version 8.0+, hopefully by then woocommerce-sniffs will be updated to work with WPCS 3.0, and then we shall update the packages in our plugins to use the newer woocommerce-sniffs and WPCS.

Yes, I'm on board with this idea. Since I have tested with 8.0, I'm happy to add a note suggesting the use of PHP8.0 for PHPCS linting.

>Side note: I also look into the PHP Coding Standards GitHub Action that runs PHPCS. The PHP version used in the GitHub Action depends on the woocommerce/grow/prepare-php@actions-v1:

It says `description: Specify PHP version. "7.4" by default.`. So if we don't specify the PHP version in our YAML file then it will pick PHP7.4. Since we have mentioned 8.0, our job runs with PHP8.0.

krutidugade:
>So, I think what we should do is that when we run phpcs / npm run lint:php, we should only use PHP 7.4 (we should mention this in the documentation). When WooCommerce core moves to using minimum PHP version 8.0+, hopefully by then woocommerce-sniffs will be updated to work with WPCS 3.0, and then we shall update the packages in our plugins to use the newer woocommerce-sniffs and WPCS.

I have made these changes in [0800642](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/0800642383c547cf75721ad544bddbb5caf58b8b) commit.

ecgan:
> > In short, "wp-coding-standards/wpcs": "^2.3" only works with PHP 7.4 and does not work with PHP 8.0+
> > Sorry, I don't quite get it, but what do you mean by "the reason mentioned above"? Why do we use 8.0? 🤔
> 
> I beg to differ. I have tested it with PHP version 8.0 and I don't get the errors. This can be confirmed in this [PHPCS job](https://github.com/woocommerce/facebook-for-woocommerce/actions/runs/5976688578/job/16215064185?pr=2624) where PHP8.0 is installed and the check has passed. This is why I have used `StephaneBour/actions-php-lint@8.0` and explicitly mentioned PHP version 8.0. We get the errors for PHP8.1+ versions.

Sorry, I should have been clearer. What I meant is that in the issue comment https://github.com/WordPress/WordPress-Coding-Standards/issues/2035#issuecomment-1077737225, it says: 

> The last WPCS release does not support PHP 8.0 or 8.1. Either run it on PHP < 8.0 or use `develop` for the time being until the next release.

"The last WPCS release" refers to the "^2.3" version. It does not support PHP 8.0, at least officially.

But I guess if it works for us, it is okay for us to use it with PHP 8.0.

> I'm not in favor of suppressing warnings. We may end up not fixing these warnings for a long time so I would rather have these visible until they are fixed.

I agree. 🙂 👍 

> > Side note: I also look into the PHP Coding Standards GitHub Action that runs PHPCS. The PHP version used in the GitHub Action depends on the woocommerce/grow/prepare-php@actions-v1:
> 
> It says `description: Specify PHP version. "7.4" by default.`. So if we don't specify the PHP version in our YAML file then it will pick PHP7.4. Since we have mentioned 8.0, our job runs with PHP8.0.

Ahh, got it. I think when I was testing and reviewing your PR with that side note, I haven't got your latest commit at that time. I see them now. Thank you. 🙂

.github/workflows/php-coding-standards.yml
```diff
@@ -28,6 +28,8 @@ jobs:
       - name: Prepare PHP
         uses: woocommerce/grow/prepare-php@actions-v1
         with:
+        # TODO: Update to PHP8.2 once WPCS (3.0) and woocommerce-sniffs (> 0.1.3) are updated to latest versions.
+          php-version: 8.0
```
ecgan, line 32: 👍 Thank you for the comments! I think this is useful because when someone check the `php-version` in the future, they can see the `TODO` easily and then act on it. 🙂

.github/workflows/php-coding-standards.yml
```diff
@@ -37,4 +39,4 @@ jobs:
         run: vendor/bin/phpcs -q -n --report=checkstyle | cs2pr
 
       - name: PHP Syntax checker
-        uses: StephaneBour/actions-php-lint@7.2
\ No newline at end of file
+        uses: StephaneBour/actions-php-lint@8.0
```
ecgan, line 42: I'm thinking we should add a line break so that we don't see the "missing end of line" warning.

Reference: 

- [Why should text files end with a newline?](https://stackoverflow.com/q/729692/379641)

.gitignore
```diff
@@ -37,5 +37,8 @@ codeception.yml
 php.ini
 wp-config.php
 
+#PHP Unit
+.phpunit.result.cache
+
 # IDE files
 .vscode/*
```
ecgan, line 44: Same thing about the end of line here.

README.md
```diff
@@ -16,20 +16,29 @@ If you have a WooCommerce.com account, you can [start a chat or open a ticket on
...
 
 ### Releasing
-Refer to the [wiki for details of how to build and release the plugin](https://github.com/woocommerce/facebook-for-woocommerce/wiki/Build-&-Release). 
+Refer to the [wiki for details of how to build and release the plugin](https://github.com/woocommerce/facebook-for-woocommerce/wiki/Build-&-Release).
+
+### PHPCS Linting and PHP 8.1+
+
+We currently do not support PHPCS on PHP 8.1+ versions. Please run PHPCS checks on PHP 8.0 or lower versions.
+
+Alternately, you can run PHPCS checks on PHP8.1+ versions by appending `?? ''` code within `trim()` at Line 280 of file /vendor/wp-coding-standards/wpcs/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php and Line 194 of file /vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WP/I18nSniff.php
```
ecgan, line 44: I'm leaning towards not putting the code and line details here in the readme, because it can get outdated quite fast and easily, e.g. when someone modify the files, then the line number mentioned here may not be accurate anymore.

Alternatively, we can include a link to our discussion comment in this PR, so that people can get more context details if they want to.

If you do make the changes here, do also put an "end of line" at the end of the file, even though there is no warning symbol in GitHub. 😄

ecgan (approved):
The changes here look good. Tested and they run okay. I'm approving this PR. 👍 

I am leaving a few minor comments below. I'll trust your judgement on them. 🙂 

Thanks so much @krutidugade !

krutidugade:
>Sorry, I should have been clearer. What I meant is that in the issue comment https://github.com/WordPress/WordPress-Coding-Standards/issues/2035#issuecomment-1077737225, it says:
>>The last WPCS release does not support PHP 8.0 or 8.1. Either run it on PHP < 8.0 or use develop for the time being until the next release.

>"The last WPCS release" refers to the "^2.3" version. It does not support PHP 8.0, at least officially.
But I guess if it works for us, it is okay for us to use it with PHP 8.0.

Thank you for sharing additional context. I understand you better now. And yes, let's still go ahead with 8.0 given it works for us.

README.md
krutidugade: >I'm leaning towards not putting the code and line details here in the readme, because it can get outdated quite fast and easily, e.g. when someone modify the files, then the line number mentioned here may not be accurate anymore.

This makes sense. Also, we can't assume that others will see exact same errors as us in their local setup.

>Alternatively, we can include a link to our discussion comment in this PR, so that people can get more context details if they want to.

Sounds good. Folks should be able to make the desired changes and make things work with PHP8.1+ if they read through this PR.

.gitignore
krutidugade: Made the changes in [07d3fe4](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/07d3fe4f5ddfaba0b99825ab4799728177f812be) commit.

.github/workflows/php-coding-standards.yml
krutidugade: Made the changes in [07d3fe4](https://github.com/woocommerce/facebook-for-woocommerce/pull/2624/commits/07d3fe4f5ddfaba0b99825ab4799728177f812be) commit.

PR #2632, opened 2023-09-28
Add filter for FB product image size.

### LINKED ISSUE #2503: Bug/feature - Ability to change attachment image size
## Feature requests

It would be very handy if we could change the image size url on product feed:
https://github.com/woocommerce/facebook-for-woocommerce/blob/6dbe0634fb56f5bca09db2c312800d8aedabac32/includes/fbproduct.php#L171

Currently, we can only add custom image via meta values. But a filter would be handy if we could affect the printed product image (eg different dimensions) via **wp_get_attachment_image_url** and replace the current **wp_get_attachment_url**

rawdreeg:

This PR adds a filter to update the image image size sent to FB.

Closes #2503.

1. This assumes you have FB for WooCommerce connected to your FB catalog.
2. Checkout this branch
3.  Use the `facebook_for_woocommerce_fb_product_image_size` filter as follow:
```
add_filter('facebook_for_woocommerce_fb_product_image_size', function () {
	return 'thumbnail';
});

```
4. Create/update a product and sync the change to your FB catalog, verify that the image uploaded is the correct size.

### Changelog entry

> Add - Filter the size of the  Facebook product image.

### DIFF
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -233,7 +233,7 @@ class WC_Facebook_Product {
 		 *
 		 * @param string $size The image size. e.g. 'full', 'medium', 'thumbnail'.
 		 */
-		$image_size 			  = apply_filters( 'facebook_for_woocommerce_fb_product_image_size', 'full' );
+		$image_size               = apply_filters( 'facebook_for_woocommerce_fb_product_image_size', 'full' );
 		$product_image_url        = wp_get_attachment_image_url( $this->woo_product->get_image_id(), $image_size ); ;
 		$parent_product_image_url = null;
 		$custom_image_url         = $this->woo_product->get_meta( self::FB_PRODUCT_IMAGE );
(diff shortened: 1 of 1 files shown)


includes/fbproduct.php
```diff
@@ -226,15 +226,23 @@ public function get_all_image_urls() {
...
 
-		$product_image_url        = wp_get_attachment_url( $this->woo_product->get_image_id() );
+		/**
+		 * Filters the FB product image size.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param string $size The image size. e.g. 'full', 'medium', 'thumbnail'.
+		 */
+		$image_size 			  = apply_filters( 'facebook_for_woocommerce_fb_product_image_size', 'full' );
```
budzanowski, line 236: Indent looks off.

budzanowski (approved):
Looks good with a small nitpick!

includes/fbproduct.php
rawdreeg: I can never remember if we do tabs or spaces :-D. I fixed it 👍🏾

PR #2654, opened 2023-10-24
Product Item batch api sync

rahulraina7:
[1/2]
Currently the video syncing is only supported by [batch-api](https://developers.facebook.com/docs/marketing-api/reference/product-catalog/items_batch/) / feed upload

This pull requests adds the following
* Replace update product item calls with batch api call

### flows are affected
* Move updating/saving a single product to batch api (This IMO is the major change here, as it changes an existing functionality completely)

### Tests
* Integration tests are updated and CI passing in forked repo - https://github.com/rahulraina7/facebook-for-woocommerce/actions/runs/6556265345
* **E2E Tests**
   - Single Product Update
       * Go to individual product.
       * Click Publish->Update
       * Outcome : Item synced to facebook


budzanowski:
> Note - Creating a simple product is still being synced through [Product-Item POST](https://developers.facebook.com/docs/marketing-api/reference/product-item/) api. This can be looked in a further pull request.

Honestly we should be doing that from the start. In case a product is just created and the video is being added initially, and later on never updated the https://github.com/woocommerce/facebook-for-woocommerce/pull/2655 will be not enough to make video feature work.

Beside having create feature excluded makes the code less coherent.

rawdreeg:
> Note - Creating a simple product is still being synced through [Product-Item POST](https://developers.facebook.com/docs/marketing-api/reference/product-item/) API. This can be looked at in a further pull request.

I agree with @budzanowski. I want to understand: is there any reason why this is excluded from this PR?

rahulraina7:
> > Note - Creating a simple product is still being synced through [Product-Item POST](https://developers.facebook.com/docs/marketing-api/reference/product-item/) API. This can be looked at in a further pull request.
> 
> I agree with @budzanowski. I want to understand: is there any reason why this is excluded from this PR?

I will take a stab at implementing this, honestly not a solid reason apart from it seemed a little less straight forward. Also I just assume that people do regular updates to products, so we will get the syncs eventually.

rahulraina7:
Hello @budzanowski @rawdreeg  - I have implemented creating products using batch api.

There's only one difference since batch API is a async API. We don't receive product-item-ids immediately after API call, and thus they are not cached  immediately using 

```
update_post_meta(
  $woo_product->get_id(),
  self::FB_PRODUCT_ITEM_ID,
  $fb_product_item_id
);			
```

That's called using the old create product API. I have also udpated the tests and description.

includes/Products/Sync/Background.php
```diff
@@ -342,7 +250,9 @@ private function process_item_delete( $prefixed_retailer_id ) {
 	private function send_item_updates( array $requests ): array {
 		$facebook_catalog_id = facebook_for_woocommerce()->get_integration()->get_product_catalog_id();
 		$response            = facebook_for_woocommerce()->get_api()->send_item_updates( $facebook_catalog_id, $requests );
-		$handles             = ( isset( $response->handles ) && is_array( $response->handles ) ) ? $response->handles : [];
+		$response_handles    = $response->handles;
+		// For some reason isset( $response->handles ) is false while isset( $response_handles ) is true
```
budzanowski, line 254: This is a tricky behaviour but it is known. The `isset` is not a function - it is a language construct and it works on variables only. In your case, the `$response->handles` is not a variable. `handles` is a property of the `$this->response_data` defined in the base class `JSONResponse`. You access the `handles` via the magic function `__get` from the `JSONResponse`. As it was mentioned `isset` is not able to operate on anything else than a variable and here you have a return value of a function.

Complicated for sure :) Your code is good and you can remove the comment.

budzanowski:
@rahulraina7 is this review ready after the requested changes?

rahulraina7:
> @rahulraina7 is this review ready after the requested changes?

Yes added the docbloc and the simple product creation via Batch API

budzanowski:
We did an evaluation and it looks like in general this should be good. However we are worried about the lack of the id in the metabox. This is an established flow so this may cause some issues with merchants. Could you please evaluate and add a simple message to the meta box that would indicate a waiting state? Something like: `Syncing product` when the batch call is being scheduled for a single product.

budzanowski:
@rahulraina7 a tip: whenever you finish a piece of work after a review you can use re-request for review:
image

budzanowski:
@rahulraina7 So the idea is to show that the sync process has started. It does not need to update live.

image

This information should be just replaced with the text that the sync has started - in case it has started. If that will be hard to do then I believe that we can leave it as it is now. The current UI/UX is not perfect and would require a general overhaul so any improvement that makes understanding the flow ( from the merchant perspective ) is highly welcomed.

rahulraina7:
Hello @budzanowski 

I have added support for display item id after batch api. Making it similar to earlier implementation, which gives feedback to the user that product was synced to facebook.

See attached video for reference 


https://github.com/woocommerce/facebook-for-woocommerce/assets/4215464/b946abf2-e3be-494d-a965-1998ebbaca97

budzanowski:
Thank you for the updates. We can start the final review before the merge. Before we will do I have one request:

Please create a new branch out of `develop` and move only the changeset. I just want to clean up the problematic commits and the history. After that create a fresh PR. I can help you with that in case there are any issues.

You have been added as a contributor to this repository so you can create PRs directly in repo without a fork.

rahulraina7:
Closing in favor of https://github.com/woocommerce/facebook-for-woocommerce/pull/2672

PR #2657, opened 2023-11-02
Update docs domain name to Woo.com

layoutd:

Update the old WooCommerce.com → the new Woo.com in several places, especially for documentation links (pcShBQ-1cw-p2).

Also updates some text and URLs regarding support requests.

Also changes the `homepage` in `package.json`, shouldn't be an issue should it?
**Note**: I wonder if we should also change the `author` there to Woo / WooCommerce.

1. No real code changes (only in `get_sales_page_url()` which is never actually used in favor of the `get_reviews_url()` value; [base method](https://github.com/woocommerce/facebook-for-woocommerce/blob/39793867e27e0a30f02243e85ccfaf54d51af6fd/includes/Framework/Plugin.php#L734-L736) & [child overwriting method](https://github.com/woocommerce/facebook-for-woocommerce/blob/76ab385f0d98378a23092c7fcc9bec476648e013/class-wc-facebookcommerce.php#L748-L750)).
2. Double check all new URLs actually go to a page (connect.woocommerce.com doesn't seem to be redirecting, for example).

### Changelog entry

> Doc - Use new Woo.com domain.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -34,7 +34,7 @@ This Code of Conduct applies both within project spaces and in public spaces whe
 
 ## Enforcement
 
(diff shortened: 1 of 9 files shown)


ecgan (approved):
I tested the links, they work as expected. I also searched the code base and I only found a handful of `connect.woocommerce.com` which do not need to be changed. LGTM 👍

layoutd:
Thanks @ecgan !

PR #2658, opened 2023-11-02
Create the build:dev npm script aliased to the current build script

layoutd:

Adding the `build:dev` npm script for consistency with other projects. It's currently just aliased to the existing `build` script, although it seems like `build` in most repos actually ends by building the archive.

1. See that both `npm run build` and `npm run build:dev` work and do the same thing (build assets and translations.


ecgan (commented):
Tested. They kinda work the same - they both build assets and translations.

However, `npm run build:dev` does not trigger `prebuild` (to run `composer install`) and `postbuild` (creating zip archive).

I am wondering: why do we want to add this `build:dev` script, since it is doing "the same thing"? Is there a pain point that you are experiencing from its inconsistency with other projects? 🤔 

I looked into https://github.com/woocommerce/google-listings-and-ads/blob/e95a1151cc3ed9dcb902ead95e65c1cd3cc6ff1b/package.json#L93 and I think it has pretty good setup there (look for `build` and `dev` npm scripts). Maybe we can use that as a reference?

layoutd:
> I am wondering: why do we want to add this `build:dev` script, since it is doing "the same thing"? Is there a pain point that you are experiencing from its inconsistency with other projects? 🤔

My thought process is essentially to reduce brain cycles. We have a bit of a hodgepodge of commands: some repos use `build:dev`, some use `dev`, and this one currently uses `build`. My end goal is standardize the command to build all our extensions for easier testing when jumping across repos (think: compatibility testing, etc.).

ecgan (approved):
Okay, let's do it! 👍 We can iterate on it along the way. 🙂

layoutd:
Thanks for the assistance and feedback, @ecgan 🙇🏻

PR #2665, opened 2023-11-17
Remove sync not possible modal.

### LINKED ISSUE #2636: Unable to create new product in excluded category
## 🔎 Isolate the bug

## ✍️ Describe the bug
If a category has been added to Marketing > Facebook > Product sync > **Exclude categories from sync**, and I try to create a new product in that category (that I don't care if it doesn't sync), the resulting warning only tells me to either remove the category from excluded categories, or remove it from the product.

It should also mentioned that if I mark the product as "Do not sync", I can create it. 

## 🚶‍♀️ Steps to reproduce

1. Add a category to **Exclude categories from sync**.
1. Create a new product and use the excluded category.
2. See that it's impossible to create the product in the category I want (unless I enable sync for the category, which I don't want).

## ✔️ Expected behavior
The message should indicate, "To create this product anyway, please disable sync in the Facebook details section"

budzanowski:

Closes #2636

In this PR I remove the modal that warns the merchant before crating a product if certain conditions are met. The modal was shown during the creation or update action of a product when:
- that product was set to sync
- product had a category assigned and that category was excluded from Facebook sync.

The original modal was intended to warn the merchant about the excluding sync conditions. The interaction looked as follows:

https://github.com/woocommerce/facebook-for-woocommerce/assets/17271089/a4152e39-33c1-4b33-a72d-309457df4cfe

The intended behaviour was to "force" merchant to change the settings and make the product setup consistent. But this has backfired and is causing merchants a lot of problems. The product that is using the excluded from sync category can't be updated in any way.

The simplest fix is to remove the modal entirely. This solution is supported by the fact that the Facebook Product sync section will show what is blocking the sync for that product so no information is being lost. This feature has been implemented some time ago and looks like this:

1. Create a product category
2. Go to `Marketing > Facebook > Product sync > Exclude categories from sync` and add that category to the excluded list
3. Create a new product and assign the excluded category to that product
4. Try to publish/update that product - it should work
5. Compare to the `develop` (rebuild required) branch where this operation should fail

### Additional details:

The modal was displayed in two cases:
1. When the product was created or updated
2. When the merchant was opening a product editor for a product with the excluded setup

I am removing modal for both cases. For the case 2 the modal was never displayed due to a bug.

The `<script>` added to the footer was doing an AJAX call with a wrong parameter

https://github.com/woocommerce/facebook-for-woocommerce/blob/c4c7b6a52d4ed66f6a0f497a9bd6ba2d79a5f7fd/includes/Admin.php#L1553

when a different name was expected by the AJAX handler:

https://github.com/woocommerce/facebook-for-woocommerce/blob/c4c7b6a52d4ed66f6a0f497a9bd6ba2d79a5f7fd/includes/AJAX.php#L177

This makes the total removal of the popup modal safe.

### Changelog entry

> Fix - Remove popup modal blocking product edit when product is using category excluded from sync.


budzanowski:
Zip for testing based on https://github.com/woocommerce/facebook-for-woocommerce/pull/2665/commits/77d138e8562e6b10894f1442629df3314a6e7864:

[facebook-for-woocommerce.zip](https://github.com/woocommerce/facebook-for-woocommerce/files/13389252/facebook-for-woocommerce.zip)

sukafia:
I can't publish drafts or new products when the test copy is installed on my site:



For the new products, the Publish button just doesn't work. 


Here's a screen recording of my test: 
https://github.com/woocommerce/facebook-for-woocommerce/assets/60121732/eedd95d6-2d34-4f41-91d2-b59a9b002ca6

budzanowski:
The problem found by @sukafia has been the result of cached js files. After clearing the cache the issue was no longer there.
When this will be a part of a new release this problem will not happen.

camilolunacom:
@budzanowski this is amazing, and how things should be like! I know I'm saying this for many WC users like me: THANK YOU!!!

budzanowski:
@camilolunacom I am happy that you like it.

This should be released on Tuesday.

Cheers!

PR #2672, opened 2023-12-14
Create/Update products sync to facebook with Batch API

rahulraina7:
[1/2]
Currently the video syncing is only supported by [batch-api](https://developers.facebook.com/docs/marketing-api/reference/product-catalog/items_batch/) / feed upload

This pull requests adds the following
* Replace update product item calls with batch api call
* Replace create product item calls with batch api call

### flows are affected
* Move updating/saving a single product to batch api (This IMO is the major change here, as it changes an existing functionality completely)

### Tests
* Integration tests are updated and CI passing in forked repo - https://github.com/rahulraina7/facebook-for-woocommerce/actions/runs/6556265345
* **E2E Tests**
   - Single Product Update
       * Go to individual product.
       * Click Publish->Update
       * Outcome : Item synced to facebook

### DIFF
--- a/facebook-commerce.php
+++ b/facebook-commerce.php
@@ -16,6 +16,7 @@ use WooCommerce\Facebook\Framework\Helper;
 use WooCommerce\Facebook\Framework\Plugin\Exception as PluginException;
 use WooCommerce\Facebook\Products;
 use WooCommerce\Facebook\Products\Feed;
+use WooCommerce\Facebook\Products\Sync;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -313,6 +314,8 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 				);
 
 				add_action( 'add_meta_boxes', 'WooCommerce\Facebook\Admin\Product_Sync_Meta_Box::register', 10, 1 );
+
+				add_action( 'add_meta_boxes', [ $this, 'display_batch_api_completed' ], 10, 2 );
 
 				add_action(
 					'wp_ajax_ajax_fb_toggle_visibility',
@@ -1175,7 +1178,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 
 		if ( $fb_product_item_id ) {
 			$woo_product->fb_visibility = Products::is_product_visible( $woo_product->woo_product );
-			$this->update_product_item( $woo_product, $fb_product_item_id );
+			$this->update_product_item_batch_api( $woo_product, $fb_product_item_id );
 			return $fb_product_item_id;
 		} else {
 			// Check if this is a new product item for an existing product group
@@ -1232,8 +1235,7 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 		}
 
 		if ( $fb_product_group_id ) {
-			$fb_product_item_id = $this->create_product_item( $woo_product, $retailer_id, $fb_product_group_id );
-			return $fb_product_item_id;
+			return $this->create_product_item_batch_api( $woo_product, $retailer_id, $fb_product_group_id );
 		}
 		return '';
 	}
@@ -1337,6 +1339,35 @@ class WC_Facebookcommerce_Integration extends WC_Integration {
 			$message = sprintf( 'There was an error trying to update Product Group %s: %s', $fb_product_group_id, $e->getMessage() );
 			WC_Facebookcommerce_Utils::log( $message );
 		}
+	}
+
+	/**
+	 * Creates a product item using the facebook Catalog Batch API. This replaces existing functionality,
+	 * which is currently using facebook Product Item API implemented by `WC_Facebookcommerce_Integration::create_product_item`
+	 *
+	 * @since x.x.x
+	 * @param WC_Facebook_Product $woo_product
+	 * @param string              $retailer_id
+	 **/
+	public function create_product_item_batch_api( $woo_product, $retailer_id, $product_group_id ): string {
+		try {
+			$product_data        = $woo_product->prepare_product( $retailer_id, \WC_Facebook_Product::PRODUCT_PREP_TYPE_ITEMS_BATCH );
+			$requests            = WC_Facebookcommerce_Utils::prepare_product_requests_items_batch( $product_data );
+			$facebook_catalog_id = $this->get_product_catalog_id();
+			$response            = $this->facebook_for_woocommerce->get_api()->send_item_updates( $facebook_catalog_id, $requests );
+
+			if ( $response->handles ) {
+				return '';
+			} else {
+				$this->display_error_message(
+					'Updated product on Facebook has failed.'
+				);
+			}
+		} catch ( ApiException $e ) {
+			$message = sprintf( 'There was an error trying to create a product item: %s', $e->getMessage() );
+			WC_Facebookcommerce_Utils::log( $message );
+		}
+		return '';
 	}
 
 	public function create_product_item( $woo_product, $retailer_id, $product_group_id ): string {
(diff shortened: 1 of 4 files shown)


ibndawood:
Thank you, @rahulraina7 for the PR. I'll do a review now.

includes/Products/Sync/Background.php
```diff
@@ -342,7 +250,8 @@ private function process_item_delete( $prefixed_retailer_id ) {
 	private function send_item_updates( array $requests ): array {
 		$facebook_catalog_id = facebook_for_woocommerce()->get_integration()->get_product_catalog_id();
 		$response            = facebook_for_woocommerce()->get_api()->send_item_updates( $facebook_catalog_id, $requests );
-		$handles             = ( isset( $response->handles ) && is_array( $response->handles ) ) ? $response->handles : [];
+		$response_handles    = $response->handles;
+		$handles             = ( isset( $response_handles ) && is_array( $response_handles ) ) ? $response_handles : array();
```
ibndawood, line 254: If we are checking for the existence of property `handles` in line 254, then we should not access it in line 253. Otherwise, it would throw an `Undefined property: $handles` warning if the property is absent.
```suggestion
		$handles             = ( isset( $response->handles ) && is_array( $response->handles ) ) ? $response->handles : array();
```

facebook-commerce.php
```diff
@@ -1339,6 +1341,35 @@ public function update_product_group( WC_Facebook_Product $woo_product ) {
...
+	 * @param string              $retailer_id
+	 **/
+	public function create_product_item_batch_api( $woo_product, $retailer_id, $product_group_id ): string {
+		try {
+			$product_data        = $woo_product->prepare_product( $retailer_id, \WC_Facebook_Product::PRODUCT_PREP_TYPE_ITEMS_BATCH );
+			$requests            = WC_Facebookcommerce_Utils::prepare_product_requests_items_batch( $product_data );
+			$facebook_catalog_id = $this->get_product_catalog_id();
+			$response            = $this->facebook_for_woocommerce->get_api()->send_item_updates( $facebook_catalog_id, $requests );
+
+			if ( $response->handles ) {
```
ibndawood, line 1359: I noticed that in other places, we do this. We can keep this if we are sure about the `handles` property in the response:

``` php
if ( isset( $response_handles ) && is_array( $response_handles ) )

File: facebook-commerce.php [hunk 3]
```diff
@@ -1465,6 +1496,37 @@ private function get_product_variation_attributes( array $variation ): array {
...
+	 * @return void
+	 */
+	public function update_product_item_batch_api( WC_Facebook_Product $woo_product, string $fb_product_item_id ): void {
+		$product = $woo_product->prepare_product( null, \WC_Facebook_Product::PRODUCT_PREP_TYPE_ITEMS_BATCH );
+		$requests = WC_Facebookcommerce_Utils::prepare_product_requests_items_batch( $product );
+
+		try {
+			$facebook_catalog_id = $this->get_product_catalog_id();
+			$response            = $this->facebook_for_woocommerce->get_api()->send_item_updates( $facebook_catalog_id, $requests );
+			if ( $response->handles ) {
```
ibndawood, line 1513: I noticed that in other places, we do this. We can keep this if we are sure about the `handles` property in the response:

``` php
if ( isset( $response_handles ) && is_array( $response_handles ) )

File: includes/fbutils.php [hunk 4]
```diff
@@ -622,6 +623,127 @@ public static function get_cached_best_tip() {
...
+
+			// NB: Changing this to get items_batch to work
+			// retailer_id cannot be included in the data object
+			unset( $product_data['retailer_id'] );
+			$product_data['id'] = $retailer_id;
+
+			$requests = array([
+				'method' => Sync::ACTION_UPDATE,
+				'data'   => $product_data,
+			]);
```
ibndawood, line 714:
```suggestion
			$requests = array( [
				'method' => Sync::ACTION_UPDATE,
				'data'   => $product_data,
			] );
```

includes/fbutils.php
```diff
@@ -622,6 +623,127 @@ public static function get_cached_best_tip() {
...
+		 * Prepares the requests array to be included in a batch api request.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param array $product Array
+		 * @return array
+		 */
+		public static function prepare_product_requests_items_batch( $product ) {
+			$product['item_group_id'] = $product['retailer_id'];
+			$product_data = self::normalize_product_data_for_items_batch( $product );
```
ibndawood, line 701:
```suggestion
			$product['item_group_id'] = $product['retailer_id'];
			$product_data             = self::normalize_product_data_for_items_batch( $product );
```

ibndawood (approved):
Thank you, @rahulraina7, for your patience. The changes look good to me. I've tested the simple product, variable product creation and update. They work as expected. I also looked into the logs and saw `items_batch` being called instead of the `Graph ID`.

However, there is only one blocker related to the code: the usage of `update_product_item` within the `update_fb_visibility` method. Please verify if we should also switch this to the batch API call.

I've added some minor questions and suggestions. They are not blockers. 

I've approved the PR in advance as it works as described. I'll trust you to resolve the blocker and merge the PR.

facebook-commerce.php
```diff
@@ -1175,7 +1178,7 @@ public function on_simple_product_publish( $wp_id, $woo_product = null, &$parent
 
 		if ( $fb_product_item_id ) {
 			$woo_product->fb_visibility = Products::is_product_visible( $woo_product->woo_product );
-			$this->update_product_item( $woo_product, $fb_product_item_id );
+			$this->update_product_item_batch_api( $woo_product, $fb_product_item_id );
```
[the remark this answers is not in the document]
rahulraina7: @ibndawood  - thank you for pointing this out. I am of the opinion that lets leave this function as it is, as this is merely changing the visibility of the product. Even though we could possible migrate to the new API, but that prepares an entire product again with a new Woo Product class and makes the code a bit more verbose IMO, because the current way of preparing a product for sending batch updates sends the entire product, instead of just say the visibility in this case.

Let me know what you think ?

includes/Products/Sync/Background.php
rahulraina7: Can you see @budzanowski comment here - https://github.com/woocommerce/facebook-for-woocommerce/pull/2654#discussion_r1372755189 

> This is a tricky behaviour but it is known. The isset is not a function - it is a language construct and it works on variables only. In your case, the $response->handles is not a variable. handles is a property of the $this->response_data defined in the base class JSONResponse. You access the handles via the magic function __get from the JSONResponse. As it was mentioned isset is not able to operate on anything else than a variable and here you have a return value of a function.

> Complicated for sure :) Your code is good and you can remove the comment.

facebook-commerce.php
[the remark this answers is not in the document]
ibndawood: I understand. We can proceed without changing this function.

budzanowski:
@rahulraina7 is this all or are you working on some things still?

@ibndawood how does this look after the last updates?

ibndawood:
@budzanowski - Looks good to me. I've already approved.

rahulraina7:
> @rahulraina7 is this all or are you working on some things still?
> 
> @ibndawood how does this look after the last updates?

Hello @budzanowski  - yes all is good, I am just waiting for the holiday period to be over. To have no surprise issues, I will merge this in first week of January. I am not sure of the release cycles of the package how frequent it is done.

budzanowski:
@rahulraina7 

We release every week ( Tuesday/Wednesday ) if there is something to release. Please ping after the merge.
Thanks!

PR #2704, opened 2024-03-14
Check if condition is set before setting a default value.

### LINKED ISSUE #2700: $data['condition'] = 'new'; Hard coded with no way to change it.
I am using the filter facebook_for_woocommerce_integration_prepare_product to adjust values for my products during the synchronisation with Facebook and I need to be able to specify if the item is new or used.  But I am unable to do this.

I looked through the code and noticed that in the file fbutils.php on line 733 you are hardcoding the following:
`$data['condition'] = 'new';`

As this line is called during the function normalize_product_data_for_items_batch, it is happening after the facebook_for_woocommerce_integration_prepare_product is applied so my value is being overwritten.

There is no way for me to override this hard coded value. The only way for me to fix this is to change the code of the plugin which is not ideal as it would be overwritten on next update.

My plan is to change the code at line 733 of fbutils.php to read the following:

```
if(!isset($data['condition'])||!in_array($data['condition'],['refurbished','used','new'])){
                $data['condition'] = 'new';
            }
```

With this new code, the 'new' value will still be used as default but valid alternative values will not be overwritten.

Could you please update the codebase for the next release to include the above change or something which would achieve the same results?

Also it would be very helpful to see more documentation about all of the options that can be used with the 'facebook_for_woocommerce_integration_prepare_product' filter generally. 

Thank you for your help.

```[tasklist]
### Tasks
```

rawdreeg:

In #2700, the reporter uses the `facebook_for_woocommerce_integration_prepare_product` filter to modify the condition. but we default this value to `new` when we normalize the data. This PR adds a check to avoid defaulting the condition value to `new` if the value value has been set and is valid.

Closes #2700 .

1. Use the `facebook_for_woocommerce_integration_prepare_product` filter to set a condition value in your theme function.php. And confirm the value isn't overriden

### Changelog entry

> Tweak - Check if condition is set before setting a default value.

### DIFF
--- a/includes/fbutils.php
+++ b/includes/fbutils.php
@@ -729,8 +729,14 @@ if ( ! class_exists( 'WC_Facebookcommerce_Utils' ) ) :
 		 * @return array
 		 */
 		public static function normalize_product_data_for_items_batch( $data ) {
-			// Allowed values are 'refurbished', 'used', and 'new', but the plugin has always used the latter.
-			$data['condition'] = 'new';
+			/*
+			 * To avoid overriding the condition value, we check if the value is set or is not one of
+			 * the allowed values before setting it to 'new'. Allowed values are 'refurbished', 'used', and 'new'.
+			 */
+			if ( ! isset( $data['condition'] ) || ! in_array( $data['condition'], [ 'refurbished', 'used', 'new' ], true ) ) {
+				$data['condition'] = 'new';
+			}
+
 			// Attributes other than size, color, pattern, or gender need to be included in the additional_variant_attributes field.
 			if ( isset( $data['custom_data'] ) && is_array( $data['custom_data'] ) ) {
 				$attributes = [];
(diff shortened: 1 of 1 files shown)


ibndawood (approved):
Thank you, @rawdreeg. The changes look good. I just left a minor comment.

includes/fbutils.php
```diff
@@ -729,8 +729,14 @@ public static function get_cached_best_tip() {
 		 * @return array
 		 */
 		public static function normalize_product_data_for_items_batch( $data ) {
-			// Allowed values are 'refurbished', 'used', and 'new', but the plugin has always used the latter.
-			$data['condition'] = 'new';
+			/*
+			 * To avoid overriding the condition value, we check if the value is set or is not one of
+			 * the allowed values before setting it to 'new'. Allowed values are 'refurbished', 'used', and 'new'.
+			 */
+			if ( ! isset( $data['condition'] ) || ! in_array( $data['condition'], [ 'refurbished', 'used', 'new' ], true ) ) {
```
ibndawood, line 736: Short array syntax is disallowed in WooCommerce core but allowed in this repo. We might want to take a call here if we wish to align our code with the core going forward.

rawdreeg: Fixed this here. but you're correct, short array syntaxes are used throughout the code base. We can create a separate issue to align coding styles.

PR #2742, opened 2024-05-10
Fix excluded categories in product sets check

### LINKED ISSUE #2741: Product Sets JS alert about excluded categories always displays
## 🔎 Isolate the bug

## ✍️ Describe the bug
- As reported in https://github.com/woocommerce/facebook-for-woocommerce/issues/2740#issuecomment-2102472131
- Product Sets - The excluded categories warning is always displayed, I'll add an issue pointing to this issue shortly.

## 🚶‍♀️ Steps to reproduce
1. Connect FB and exclude a category with products.
1. Create a product set that includes the excluded category.
2. The JS alert `You have selected one or more categories currently excluded from the Facebook sync. Products belonging to the excluded categories will not be added to your Facebook Product Set.` displays.
3. Create another product set containing a NON-excluded category.
4. The JS alert still displays.

## ✔️ Expected behavior
The JS alert should only display if an excluded category is actually included in the product set.

## 🔧 Solution
The culprit is the comparison in this function:
https://github.com/woocommerce/facebook-for-woocommerce/blob/2eab66267d706fc46beed78e2c7506ccc347130f/assets/js/admin/product-sets-admin.js#L58-L68

Obviously, `excludedCategories[i]` will always be in the `excludedCategories` array! 😆

layoutd:

Closes #2741 .

This PR fixes an errant JavaScript check that checks for each excluded category within the list of excluded categories itself, instead of within the list of categories selected for the given product set.

See the test instructions for #2741. The only difference is that Product Sets that have no excluded categories should be saved without a JavaScript notice.

### Changelog entry

> Fix - Incorrect alert for Product Sets without excluded categories.

### DIFF
--- a/assets/js/admin/product-sets-admin.js
+++ b/assets/js/admin/product-sets-admin.js
@@ -60,7 +60,7 @@ function hasExcludedCategories( selectedCategories, excludedCategoryIDs ) {
 	let counter = 0;
 
 	for ( let i = 0; i < excludedCategoryIDs.length; i++ ) {
-		if ( excludedCategoryIDs.includes( excludedCategoryIDs[i] ) ) counter++;
+		if ( selectedCategories.includes( excludedCategoryIDs[i] ) ) counter++;
 	}
 
 	return counter > 0;


ibndawood (approved):
Thank you, @layoutd, for catching this issue and fixing it.

PR #2767, opened 2024-07-11
Align PHP require version to composer.json

rawdreeg:

This PR make changes to align PHP require version to composer.json

### Changelog entry

>Tweak - Align PHP require version to composer.json

### DIFF
--- a/facebook-for-woocommerce.php
+++ b/facebook-for-woocommerce.php
@@ -12,7 +12,7 @@
  * Author: Facebook
  * Author URI: https://www.facebook.com/
  * Version: 3.2.5
- * Requires at least: 6.4
+ * Requires at least: 5.6
  * Requires PHP: 7.4
  * Text Domain: facebook-for-woocommerce
  * Requires Plugins: woocommerce
--- a/facebook-for-woocommerce.php
+++ b/facebook-for-woocommerce.php
@@ -12,7 +12,7 @@
  * Author: Facebook
  * Author URI: https://www.facebook.com/
  * Version: 3.2.5
- * Requires at least: 5.6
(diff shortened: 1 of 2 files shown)


facebook-for-woocommerce.php
```diff
@@ -12,7 +12,7 @@
  * Author: Facebook
  * Author URI: https://www.facebook.com/
  * Version: 3.2.5
- * Requires at least: 5.6
+ * Requires at least: 7.4
```
ibndawood, line 15: This is the WordPress version. I am setting the same as the one used in WooCommerce..
```suggestion
 * Requires at least: 6.4
 * Requires PHP: 7.4
```

ibndawood (changes requested):
Thank you, @rawdreeg. I've made some changes. Here is the plugin header details: https://developer.wordpress.org/plugins/plugin-basics/header-requirements/

rawdreeg:
Thanks @ibndawood  , for the Requires at least, shouldn't that be 5.6? then we can update readme as well.

ibndawood:
I just used the same value as in WooCommerce core here: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php

ibndawood (approved):

rawdreeg:
> I just used the same value as in WooCommerce core here: [woocommerce/woocommerce@trunk/plugins/woocommerce/woocommerce.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php?rgh-link-date=2024-07-11T13%3A34%3A27Z)

Ah neat. We can make another PR to adjust the minimum required WP.

PR #2781, opened 2024-08-08
Use get_title() for content_name to match catalog name.

### LINKED ISSUE #2774: Inconsistent use of get_title vs. get_name on product objects
You sometimes use `$product->get_title()` for content_name and sometimes use `$product->get_name()`.

This creates inconsistencies between events and the catalog.

rawdreeg:

We received a report ( #2774 ) pointing out inconsistencies in populating the content_name property in events we send to Meta, sometimes using `get_title()` vs `get_name()`. I have looked into this and found that `get_title()` applies the `woocommerce_product_title` filter to the product name prop. This may result in different values being sent to Facebook. I also find that [we use get_title for the catalog data](https://github.com/woocommerce/facebook-for-woocommerce/blob/df03a97e449b242c1ef59ddc70eac2de966cc3e8/includes/fbproduct.php#L657), so I propose we use the get_title method

Closes #2774.

1. Use the `woocommerce_product_title` filter. E.g.:

```
function custom_woocommerce_product_title_prefix( $title, $product ) {
    // Add the prefix "Special: " to the product title
    $title = 'Special: ' . $title;

    return $title;
}
add_filter( 'woocommerce_product_title', 'custom_woocommerce_product_title_prefix', 10, 2 );

```

2. Using the FB pixel helper browser extension, visit a product and confirm that the ViewContent event fires and all the details are confirmed.
3. Repeat this with the addToCart, Checkout and Purchase Events

### Changelog entry

> Update - Use get_title() for content_name to match catalog name.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -583,7 +583,7 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 				'event_name'  => 'AddToCart',
 				'custom_data' => array(
 					'content_ids'  => wp_json_encode( \WC_Facebookcommerce_Utils::get_fb_content_ids( $product ) ),
-					'content_name' => $product->get_name(),
+					'content_name' => $product->get_title(),
 					'content_type' => 'product',
 					'contents'     => wp_json_encode(
 						array(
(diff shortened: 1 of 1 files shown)


nima-karimi (approved):
LGTM ✅ All events include the correct title with the added prefix.

rawdreeg:
Thanks @nima-karimi for the review.

Dekadinious:
Not to be a party pooper here, but given that this is in a WooCommerce context and that WooCommerce also has filters for `$product->get_name();`  one would think that using that everywhere would be more future proof as WooCommerce might move products to custom tables just as they did with orders (HPOS).

Using native getters and setters on objects is the preferable way to code for WooCommerce anyway, so people using the post title filter might be considered to be "doing_it_wrong".

Dekadinious:
Sorry, had a brain fart there for a second. Move on, nothing to see here! :)

PR #2814, opened 2024-10-26
Show Settings navigation tabs on Product Sets taxonomy screens

layoutd:

This PR renders the basic navigation tabs used on "Connection", "Product Sync", and "Advertise" pages, on the "Product Sets" pages as well. 

They're rendered (hidden) at the top of the corresponding list + edit pages, and then moved in the DOM and displayed. Depending on page load speed, they might not be available for short period, but I've seen the tabs flicker on all the other settings pages of the extension as well.

### Screenshots:

**Before**
image

**After**
image
image

**Other Settings Pages (for comparison)**
image

1. Connect a store.
2. Navigate around the different sections of the Facebook for WooCommerce extension.
3. Confirm the navigation tabs appear correctly in the different screens related to Product Sets.
4. Try with several browser resolutions (mobile, desktop, etc).
5. Also confirm that the correct menu item is highlighted always (Marketing > Facebook).

### Changelog entry

> Add - Extension navigation tabs on Product Sets screens.

### DIFF
--- a/includes/Admin/Settings.php
+++ b/includes/Admin/Settings.php
@@ -14,7 +14,6 @@ use Automattic\WooCommerce\Admin\Features\Features as WooAdminFeatures;
 use WooCommerce\Facebook\Admin\Settings_Screens;
 use WooCommerce\Facebook\Admin\Settings_Screens\Connection;
 use WooCommerce\Facebook\Framework\Helper;
-use WooCommerce\Facebook\Framework\Plugin\Compatibility;
 use WooCommerce\Facebook\Framework\Plugin\Exception as PluginException;
 
 defined( 'ABSPATH' ) || exit;
@@ -53,7 +52,7 @@ class Settings {
 		add_action( 'wp_loaded', array( $this, 'save' ) );
 		add_filter( 'parent_file', array( $this, 'set_parent_and_submenu_file' ) );
 
-		add_action( 'admin_head', array( $this, 'add_tabs_to_product_sets_taxonomy' ) );
+		add_action( 'all_admin_notices', array( $this, 'add_tabs_to_product_sets_taxonomy' ) );
 	}
 
 	/**
@@ -349,6 +348,7 @@ class Settings {
 	 * @since 3.3.0
 	 */
 	public function add_tabs_to_product_sets_taxonomy() {
+
 		// Only load this on the edit-tags.php page
 		$screen                  = get_current_screen();
 		$is_taxonomy_list_page   = 'edit-tags' === $screen->base;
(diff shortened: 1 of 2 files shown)


includes/Admin/Settings.php
```diff
@@ -312,4 +340,70 @@ public function get_tabs() {
...
+					<?php endif; ?>
+				</style>
+				<script>
+						document.addEventListener('DOMContentLoaded', function() {
+								// Insert custom tabs into #wpbody
+								let tabs = document.querySelector('nav.facebook-for-woocommerce-tabs');
+								if (tabs) {
+										let wpbody = document.querySelector('#wpbody-content > .wrap');
+										if (wpbody) {		
+											// wpbody.querySelector('h1').remove();
```
brezocordero, line 394: Nitpick: Looks like a leftover?

includes/Admin/Settings.php
```diff
@@ -312,4 +340,70 @@ public function get_tabs() {
...
+								padding-top: 0;
+								position: relative;
+								top: -10px;
+								margin-bottom: -10px;
+							}
+						}
+					<?php endif; ?>
+				</style>
+				<script>
+						document.addEventListener('DOMContentLoaded', function() {
```
brezocordero, line 388: @layoutd 
This approach works, but out of curiosity, have you considered inserting the tabs using the `all_admin_notices` action? It might be a cleaner solution, as it ensures the tabs are added after the screen options and notices, avoiding the need for addEventListener for DOM manipulation.

brezocordero (approved):
@layoutd 
Tested and it works.
I am pre-approving but please remove the commented code.
I have a question about the approach but it is not blocking as this works. It could also be added as a future improvement.

includes/Admin/Settings.php
layoutd: Good spot, thanks. At least I got all the `console.log`s 😆 Removed in 31752c1f061a1a12b0f4aca21893b111d649dc43

includes/Admin/Settings.php
layoutd: ZOMG thank you! I tried a few times to remember the exact hook to use, but even looking at [Actions Run During an Admin Page Request](https://developer.wordpress.org/apis/hooks/action-reference/#actions-run-during-an-admin-page-request) all my brain could provide was `admin_notices`, which didn't work.
I've made the change in ea089fa7f9cd0f4a18fd715099ee0795ddb69bd3 and cleaned up the JS (and CSS a bit, too)

brezocordero:
@layoutd Thanks for making the changes.
Tested and looks good. 🚢

layoutd:
Thanks @brezocordero!

PR #2815, opened 2024-10-26
Configuration label and tooltip cleanup

layoutd:

This PR revises some fields labels related to Facebook Sync, adds a few tooltips, and adds links to more information at the bottom of Advertise, Product sync, and in the Debug section options.

Pending is adding a documentation link at the bottom of Product Sets, but would probably need some DOM element manipulation with JS, so leaving that one for a different PR.

It also clears up some Product Editor form issues where the Custom Image URL field was always displayed initially, and makes the field contingent on the "Use custom image" option being selected.

### Screenshots:

image

image

image

image

1. Check that all learn more links work correctly.
2. Check text of learn more links and tooltips.
3. Confirm that the Custom Image URL field works correctly (displays and hides, saves, etc) on Simple and Variable products.

### Changelog entry

> Tweak - Tooltips, config labels, documentation links.

### DIFF
--- a/assets/css/admin/facebook-for-woocommerce-connection.css
+++ b/assets/css/admin/facebook-for-woocommerce-connection.css
@@ -53,6 +53,11 @@
 	background-image: url( '../../images/icon-2.png' );
 }
 
+#wc-facebook-connection-box .actions {
+	display: flex;
+	align-items: center;
+}
+
 #wc-facebook-connection-box .button {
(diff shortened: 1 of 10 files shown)


layoutd:
Thanks @message-dimke!

I forgot to mention that I centered the Disconnect link vertically, very trivial but I think it's much better

**Before**
image

**After**
image

PR #2840, opened 2024-12-16
Fixes the item not found error by using filter in the product endpoint

halilozanakgul:

Updates the ProductCatalog/Products/Id API call to use endpoint filter instead of directly getting the product with id.

This API call is mainly used for checking if the product item exists on Meta. When the item does not exists it threw an error. With this change the endpoint will stop erroring and will return empty list as response.

1. Create any product on admin page that will be synced to Meta.
2. Check for errors.

### Changelog entry

> Dev - Use product GET endpoint filter function to reduce the item not found errors.


tests/Unit/ApiTest.php
```diff
@@ -540,8 +540,9 @@ public function test_get_product_facebook_ids_creates_get_ids_request() {
...
 			$this->assertEquals( 'GET', $parsed_args['method'] );
 
-			$path = "catalog:{$facebook_product_catalog_id}:" . base64_encode( $facebook_product_retailer_id );
-			$path = "/{$path}/?fields=id,product_group{id}";
+			$filter = urlencode('{"retailer_id":{"eq":"' . $facebook_product_retailer_id . '"}}');
+			$fields = urlencode('id,product_group{id}');
+			$path = "/{$facebook_product_catalog_id}/products?filter={$filter}&fields={$fields}";
 
 			$this->assertEquals( "{$this->endpoint}{$this->version}{$path}", $url );
 			return [
```
mshymon, line 548: Just making sure: is the api response structure in production still matches one in the test here after you updated request?

facebook-commerce.php
```diff
@@ -2933,15 +2933,21 @@ public function get_product_fbid( string $fbid_type, int $wp_id, $woo_product =
...
 
 		try {
-			$facebook_ids = $this->facebook_for_woocommerce->get_api()->get_product_facebook_ids(
+			$response = $this->facebook_for_woocommerce->get_api()->get_product_facebook_ids(
 				$this->get_product_catalog_id(),
 				$fb_retailer_id
 			);
 
-			if ( $facebook_ids->id ) {
+			if ( !$response->data || empty( $response->data ) ) {
```
mshymon, line 2941: if ( !$response->data ) alone will check both conditions. 

```suggestion
			if ( !$response->data ) {
```

mshymon (changes requested):
PR checks are failing, please fix and make sure you addressed or answered to all the PR comments. 

By the way, you can remove line " Do the changed files pass phpcs checks? Please remove phpcs:ignore comments in changed files and fix any issues, or delete if not practical." on your PR description.

includes/API/ProductCatalog/Products/Id/Request.php
```diff
@@ -8,9 +8,9 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Request object for Product Catalog > Product Groups > Products > Update Graph Api.
+ * Request object for Product Catalog > Products > Get Graph Api.
  *
- * @link https://developers.facebook.com/docs/marketing-api/reference/product-group/products/
+ * @link https://developers.facebook.com/docs/marketing-api/reference/product-catalog/products/
  */
 class Request extends ApiRequest {
```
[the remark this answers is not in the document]
halilozanakgul: But this is only used to fetch the ids. I think the intent here was to make sure we can still have a separate 'read' endpoint file later if necessary. And the rest of the API calls are not consistent either.

tests/Unit/ApiTest.php
halilozanakgul: Thanks for pointing out, it was not matching, I fixed it.

facebook-commerce.php
```diff
@@ -1107,9 +1107,11 @@ public function on_variable_product_publish( $wp_id, $woo_product = null ) {
 			return;
 		}
 
+		$use_filter_endpoint = ($woo_product->woo_product->get_date_created() == $woo_product->woo_product->get_date_modified());
```
devbodaghe, line 1110: variable is a boolean, it should be named to clearly indicate its purpose. Could be renamed to `should_use_filter_endpoint`

facebook-commerce.php
```diff
@@ -2920,7 +2924,7 @@ public function on_quick_and_bulk_edit_save( $product ) {
 	 * @param WC_Facebook_Product|null $woo_product product
 	 * @return string facebook product id or an empty string
 	 */
```
devbodaghe, line 2926: Also add description for the variable you are adding to the function

facebook-commerce.php
```diff
@@ -2935,12 +2939,25 @@ public function get_product_fbid( string $fbid_type, int $wp_id, $woo_product =
...
-			if ( $facebook_ids->id ) {
+			if ( $use_filter_endpoint ) {
+				if ( empty( $response->data ) ) {
+					return null;
+				}
+
+				$facebook_ids = $response->data[0];
+			}
+			else
+			{
```
devbodaghe, line 2956: the linting can be fixed here

includes/API.php
```diff
@@ -443,8 +443,8 @@ public function delete_product_item( string $facebook_product_id ): API\ProductC
 	 * @throws ApiException In case of network request error.
 	 * @throws API\Exceptions\Request_Limit_Reached
 	 */
-	public function get_product_facebook_ids( string $facebook_product_catalog_id, string $facebook_retailer_id ): API\ProductCatalog\Products\Id\Response {
-		$request = new API\ProductCatalog\Products\Id\Request( $facebook_product_catalog_id, $facebook_retailer_id );
+	public function get_product_facebook_ids( string $facebook_product_catalog_id, string $facebook_retailer_id, $use_filter_endpoint ): API\ProductCatalog\Products\Id\Response {
```
devbodaghe, line 446: add description for the variable above

PR #2841, opened 2024-12-17
Fix Product Syncing via Feeds

mshymon:

The current state of the product syncing via Feeds is broken - while plugin code base has business logic to generate feed file daily and serve the feed file on request, there is no logic for triggering a feed one time upload. 

In this PR I am suggesting the following fixes to unblock product sync via feeds:

1. Trigger a daily feed upload sessions.

      - Sending a one time  create feed upload request to Meta would happen on feed file generation completion (this is to ensure we upload as feed file fresh data as possible).
      - Added logic of choosing or creating the correct feed to which such feed upload would happen. This included local caching of the integrated feed ID, validating it is valid, choosing a feed from catalog or creating a new one.  

2. Update Feed file content by:

      - Adding `GTIN` to the feed file
      - Adding `quantity_to_sell_on_facebook` to the feed file
      - Adding `woo_product_id` to the feed file. This is to resolve any ambiguity of what is the woo id if needed due to retailer_id is a construct of both `woo_product_id` and `sku`.
 

1. Unit Tests
  A. Run new tests:
      - `./vendor/bin/phpunit --filter ApiTest`
      - `./vendor/bin/phpunit --filter fbproductTest`
  B. Run all tests: 
      - `npm run test:php `
2. Manual Tests.
      - Steps were added via an internal dogfooding doc and requires access to Meta internal tools to validate feed upload sessions status. 

### Changelog entry

Fix - Fixed feeds by requesting a feed file upload session after feed file is generated and added missing new fields to the feed file.

>

### DIFF
--- a/includes/Products/Feed.php
+++ b/includes/Products/Feed.php
@@ -39,6 +39,8 @@ class Feed {
 	/** @var string the WordPress option name where the secret included in the feed URL is stored */
 	const OPTION_FEED_URL_SECRET = 'wc_facebook_feed_url_secret';
 
+	// TODO: add description
+	const FEED_NAME = 'Product Feed by Facebook for WooCommerce plugin. DO NOT DELETE.';
 
 	/**
 	 * Feed constructor.
@@ -189,9 +191,9 @@ class Feed {
 	 * TODO: description and signiture
 	 */
 	public function send_request_to_upload_feed() {
-		$feed_id = self::retrieve_integration_feed_id();
+		$feed_id = self::retrieve_or_create_integration_feed_id();
 		if ($feed_id === null || $feed_id === '') {
-			WC_Facebookcommerce_Utils::log( 'Feed: integration feed ID was not retrieved, feed will not be upload' );
+			WC_Facebookcommerce_Utils::log( 'Feed: integration feed ID is null or empty, feed will not be uploaded.' );
 			return;
 		}
 
@@ -202,14 +204,14 @@ class Feed {
 		try {
 			facebook_for_woocommerce()->get_api()->create_upload($feed_id, $data );
 		} catch ( Exception $exception ) {
-			facebook_for_woocommerce()->log( 'Could not send create feed upload request: ' . $exception->getMessage() );
+			facebook_for_woocommerce()->log( 'Could not send create feed upload via request: ' . $exception->getMessage() );
 		}
 	}
 
 	/**
 	 * TODO: description and signiture
 	 */
-	public function retrieve_integration_feed_id() {
+	public function retrieve_or_create_integration_feed_id() {
 		// Step 1 - Get feed ID if it is already available in local cache
 		$feed_id = facebook_for_woocommerce()->get_integration()->get_feed_id();
 		if ($feed_id !== null && $feed_id !== '') {
@@ -228,15 +230,20 @@ class Feed {
 		}
 
 		// Step 3 - Create a new feed
+		$feed_id = self::create_feed_id();
+		if ($feed_id !== null && $feed_id !== '') {
+			facebook_for_woocommerce()->get_integration()->update_feed_id($feed_id);
+			WC_Facebookcommerce_Utils::log( 'Feed: retrieve_integration_feed_id(): feed_id = '.$feed_id.', created a new feed via Meta API.');
+			return $feed_id;
+		}
 
-		// TODO: create a new feed in this case
 		return '';
 	}
 
 	/**
 	 * TODO: description and signiture
 	 */
-	public function query_and_filter_integration_feed_id() {
+	private function query_and_filter_integration_feed_id() {
 		$catalog_id = facebook_for_woocommerce()->get_integration()->get_product_catalog_id();
 
 		// No catalog id. Most probably means that we don't have a valid connection.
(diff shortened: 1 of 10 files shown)


includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+			if ( self::validate_feed_exists($feed_id) ) {
+				WC_Facebookcommerce_Utils::log( 'Feed: retrieve_integration_feed_id(): feed_id = '.$feed_id.', from local cache.');
+				return $feed_id;
+			} else {
+				WC_Facebookcommerce_Utils::log( 'Feed: retrieve_integration_feed_id(): feed_id = '.$feed_id.', from local cache was invalidated.');
+			}
+		}
+
+		// Step 2 - Query feeds data from Meta and filter the right one
+		$feed_id = self::query_and_filter_integration_feed_id();
```
Jmencab, line 233: This could throw an error that is currently not being caught

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+			foreach ( $feed_nodes as $feed ) {
+				try {
+					$feed_metadata = facebook_for_woocommerce()->get_api()->read_feed( $feed['id'] );
+				} catch ( Exception $e ) {
+					$message = sprintf( 'There was an error trying to get feed metadata: %s', $e->getMessage() );
+					WC_Facebookcommerce_Utils::log( $message );
+					continue;
+				}
+
+				$woo_feed_name_option_1 = self::FEED_NAME;
```
Jmencab, line 331: suggest factor out to private function named something like `private function pattern_match_feed_names()`

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+	 * the plugin integration feed ID
+	 * 
+	 * @return		string the integration feed ID
+	 * 
+	 * @internal
+	 */
+	private function query_and_filter_integration_feed_id() {
+		$catalog_id = facebook_for_woocommerce()->get_integration()->get_product_catalog_id();
+		if ( '' === $catalog_id ) {
+			throw new Error( 'No catalog ID' );
```
Jmencab, line 294: not being caught in retrieve_or_create_integrationFeed_id()

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
 	}
 
 
+	/**
```
Jmencab, line 190: change comment on 181 that says it defaults to 15 min if we are doing it every day

halilozanakgul (changes requested):

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+	 * Sends request to Meta to start a one-time feed file upload session.
+	 *
+	 * @internal
+	 */
+	public function send_request_to_upload_feed() {
+		$feed_id = self::retrieve_or_create_integration_feed_id();
```
devbodaghe, line 196: can be simplified to 
`if (empty($feed_id)) {
    WC_Facebookcommerce_Utils::log('Feed: integration feed ID is null or empty, feed will not be uploaded.');
    return;
}`

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+		}
+
+		$data = [
+			'url' => Feed::get_feed_data_url(),
+		];
+
+		try {
+			facebook_for_woocommerce()->get_api()->create_upload($feed_id, $data );
+		} catch ( Exception $exception ) {
+			facebook_for_woocommerce()->log( 'Could not send create feed upload via request: ' . $exception->getMessage() );
```
devbodaghe, line 209: `facebook_for_woocommerce()->log('Failed to create feed upload request: ' . $exception->getMessage());
`

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+				WC_Facebookcommerce_Utils::log( 'Feed: retrieve_integration_feed_id(): feed_id = '.$feed_id.', from local cache.');
+				return $feed_id;
+			} else {
+				WC_Facebookcommerce_Utils::log( 'Feed: retrieve_integration_feed_id(): feed_id = '.$feed_id.', from local cache was invalidated.');
+			}
+		}
+
+		// Step 2 - Query feeds data from Meta and filter the right one
+		$feed_id = self::query_and_filter_integration_feed_id();
+		if ($feed_id !== null && $feed_id !== '') {
```
devbodaghe, line 234: `if (!empty($feed_id)) {`

includes/Products/Feed.php
```diff
@@ -179,6 +187,192 @@ public function schedule_feed_generation() {
...
+	 * @param 		string $feed_id the feed ID
+	 * 
+	 * @return		bool true if the feed ID is valid
+	 * 
+	 * @internal
+	 */
+	private function validate_feed_exists($feed_id) {
+		$catalog_id = facebook_for_woocommerce()->get_integration()->get_product_catalog_id();
+		if ( '' === $catalog_id ) {
+			throw new Error( 'No catalog ID' );
```
devbodaghe, line 263: Are the errors stored in the logs or was it a conscious decision to not add any logs here

tests/Unit/ApiTest.php
```diff
@@ -715,4 +715,63 @@ public function test_read_feeds_creates_read_feeds_request() {
 			$response->data
 		);
 	}
+
```
devbodaghe, line 718: Can you add a test to check unhappy path
If errors are thrown or logs are generated

includes/Products/Feed.php
```diff
@@ -62,6 +67,9 @@ private function add_hooks() {
 
 		// handle the feed data request
 		add_action( 'woocommerce_api_' . self::REQUEST_FEED_ACTION, array( $this, 'handle_feed_data_request' ) );
+
+		// Send request fir feed one time upload after feed file generated
+ 		add_action( 'wc_facebook_feed_generation_completed', array( $this, 'send_request_to_upload_feed' ) );
 	}
```
Jmencab, line 75: // fpassthru might be disabled in some hosts (like Flywheel)
^ what do we do if fpassthru is allowed? We don't write anything in this case as far as I can tell @mshymon

tests/Unit/ApiTest.php
mshymon: I think it won't bring value in this case, as we just mock response. Test with error mocks were added for other calls. that should be enough I think.

mshymon (commented):
Addressed the PR comments. Thanks everyone for reviewing the code!

vinkmeta (approved):

PR #2842, opened 2024-12-18
Expose Brand & MPN to Woocommerce UI

devbodaghe:
Changes proposed in this Pull Request:

This PR introduces a new feature to the Facebook WooCommerce Plugin, allowing users to add a brand field to their products. With this update, we can seamlessly synchronize the brand field with the Facebook Commerce Manager platform.
Screenshot 2025-01-10 at 18 57 18

Screenshot 2024-10-30 at 18 08 36

### DIFF
--- a/includes/Products/FBCategories.php
+++ b/includes/Products/FBCategories.php
@@ -22,6 +22,8 @@ defined( 'ABSPATH' ) || exit;
  */
 class FBCategories {
 
+	private $keys_to_exclude = ['brand' => true];
+
 	/**
 	 * Fetches the attribute from a category using attribute key.
 	 *
@@ -129,7 +131,14 @@ class FBCategories {
 		$return_attributes = array();
 		foreach ( $category['attributes'] as $attribute_hash ) {
 			// Get attribute array from the stored hash version
-			$return_attributes[] = $this->get_attribute_field_by_hash( $attribute_hash );
+			$attribute = $this->get_attribute_field_by_hash( $attribute_hash );
+
+			// Skip if attribute is invalid or its key is in the exclude list
+			if ( ! is_array( $attribute ) || empty( $attribute['key'] ) || isset( $this->keys_to_exclude[ $attribute['key'] ] ) ) {
+				continue;
+			}
+
+			$return_attributes[] = $attribute;
 		}
 
 		return $return_attributes;
--- a/includes/fbproduct.php
+++ b/includes/fbproduct.php
@@ -36,6 +36,8 @@ class WC_Facebook_Product {
 	const FB_VARIANT_IMAGE       = 'fb_image';
 	const FB_VISIBILITY          = 'fb_visibility';
 	const FB_REMOVE_FROM_SYNC    = 'fb_remove_from_sync';
+	const FB_BRAND               = 'fb_brand';
+	const FB_MPN              	 = 'fb_mpn';
 
 	const MIN_DATE_1 = '1970-01-29';
 	const MIN_DATE_2 = '1970-01-30';
@@ -358,6 +360,28 @@ class WC_Facebook_Product {
 		}
 	}
 
+	public function set_fb_brand( $fb_brand ) {
+		$fb_brand = stripslashes(
+			WC_Facebookcommerce_Utils::clean_string( $fb_brand )
+		);
+		update_post_meta(
+			$this->id,
+			self::FB_BRAND,
+			$fb_brand
+		);
+	}
+
+	public function set_fb_mpn( $fb_mpn ) {
+		$fb_mpn = stripslashes(
+			WC_Facebookcommerce_Utils::clean_string( $fb_mpn )
+		);
+		update_post_meta(
+			$this->id,
+			self::FB_MPN,
+			$fb_mpn
+		);
+	}
+
 	public function set_price( $price ) {
 		if ( is_numeric( $price ) ) {
 			update_post_meta(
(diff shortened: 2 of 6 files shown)


mshymon:
@devbodaghe, is the screenshot  in the description missing MPN field? 
Also, it is easier to compare the change if you include before / after screenshots and potentially highlight area with the new elements in red.  

Also, any particular reason why Brand goes between image URL and Price?

includes/fbproduct.php
```diff
@@ -391,6 +431,48 @@ public function set_use_parent_image( $setting ) {
...
+		}
+
+		// If not set, try to get brand from post meta
+		if ( empty( $fb_brand ) ) {
+			$fb_brand = get_post_meta(
+				$this->id,
+				self::FB_BRAND,
+				true
+			);
+		}
```
mshymon, line 450: Why do we need to keep variable $this->fb_brand, why not always read it from the DB via get_post_meta() when we need the value? Maintaining an intermediate variable  in sync can lead to bugs.

includes/fbproduct.php
```diff
@@ -391,6 +431,48 @@ public function set_use_parent_image( $setting ) {
...
+			if ( $brand ) {
+				$fb_brand = WC_Facebookcommerce_Utils::clean_string( $brand );
+			} elseif ( !is_wp_error( $brand_taxonomy ) && $brand_taxonomy ) {
+				$fb_brand = WC_Facebookcommerce_Utils::clean_string( $brand_taxonomy );
+			} else {
+				$fb_brand = wp_strip_all_tags( WC_Facebookcommerce_Utils::get_store_name() );
+			}
+		}
+
+		return WC_Facebookcommerce_Utils::clean_string( $fb_brand );
```
mshymon, line 473: If we call WC_Facebookcommerce_Utils::clean_string at the end, can we remove previous such calls in the code of this method?

includes/fbproduct.php
```diff
@@ -845,7 +944,11 @@ function( $key ) {
 		$matched_attributes = array_filter(
 			$all_attributes,
 			function( $attribute ) use ( $sanitized_keys ) {
-				return in_array( $attribute['key'], $sanitized_keys );
+				// Check if $attribute is an array and has the 'key' index
+				if (is_array($attribute) && isset($attribute['key'])) {
```
mshymon, line 948: Formatting, check also through the code

```suggestion
				if ( is_array( $attribute ) && isset( $attribute['key'] ) ) {
```

mshymon (changes requested):

devbodaghe:
> @devbodaghe, is the screenshot in the description missing MPN field? Also, it is easier to compare the change if you include before / after screenshots and potentially highlight area with the new elements in red.
> 
> Also, any particular reason why Brand goes between image URL and Price?



> @devbodaghe, is the screenshot in the description missing MPN field? Also, it is easier to compare the change if you include before / after screenshots and potentially highlight area with the new elements in red.
> 
> Also, any particular reason why Brand goes between image URL and Price?

It's updated now

includes/fbproduct.php
devbodaghe: You make a good point and I have updated this without the need for this value.

includes/fbproduct.php
devbodaghe: Great point

includes/fbproduct.php
```diff
@@ -845,7 +909,11 @@ function( $key ) {
 		$matched_attributes = array_filter(
 			$all_attributes,
 			function( $attribute ) use ( $sanitized_keys ) {
-				return in_array( $attribute['key'], $sanitized_keys );
+				// Check if $attribute is an array and has the 'key' index
```
mshymon, line 912: Comment is saying the same think as the code. Remove or add value why we do this check (not what we do)

mshymon (approved):

bin/fb_google_category_to_attribute_mapping.json
vinkmeta, line 1: It is not a good idea to modify this json as its a standard static file that is not supposed to be modified. The alternative approach is to filter Brand when surfacing GPC attributes so that it never get surfaced and you can introduce a new field called 'Brand' that will eventually use the existing brand field's values.

includes/Products/FBCategories.php
```diff
@@ -22,6 +22,8 @@
  */
 class FBCategories {
 
+	private $keys_to_exclude = ['brand' => true];
```
devbodaghe, line 25: Added associative array here to avoid lopping over regular array

includes/Products/FBCategories.php
```diff
@@ -200,6 +209,7 @@ public function get_categories() {
 	 */
 	protected function get_attribute_field_by_hash( $hash ) {
 		$fields_data = $this->get_raw_attributes_fields_data();
+		// error_log('fields_data' . print_r($fields_data, true));
```
vinkmeta, line 212: Lets remove this error log

includes/fbproduct.php
```diff
@@ -89,6 +91,7 @@ class WC_Facebook_Product {
 	 */
 	public $fb_visibility;
 
+
```
vinkmeta, line 94: Lets remove this extra line

vinkmeta:
I manually tested this PR locally and discovered a critical bug: If a value for the Brand attribute is already saved and I then upgrade the plugin version, the Brand value appears empty. To fix this, you may need to fetch the Brand value while rendering this attribute in the Facebook.

vinkmeta (approved):

PR #2855, opened 2025-01-20
Install svn in unit test file

halilozanakgul:

Installs svn in `php-unit-tests.yml` file because the tests error otherwise.


halilozanakgul:
#2857 fixes the same thing

PR #2862, opened 2025-01-21
[WIP] Add WordPress and WooCommerce versions to the unit tests

halilozanakgul:
This PR adds WordPress and WooCommerce versions matrix in the unit tests so we can make sure new code works in older WordPress and WooCommerce versions too.

See Github's checks running multiple WC and WP versions.

### Changelog entry
Dev - Added WordPress and WooCommerce versions to the tests

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -31,8 +31,8 @@ jobs:
     strategy:
       matrix:
         php: [7.4, 8.3]
-        wp-version: [latest]
-        wc-version: [latest]
+        wp-version: [5.7, 6.6, 6.7]
+        wc-version: [9.6.0, 9.7,0]
 
     steps:
       - name: Install SVN
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -31,8 +31,8 @@ jobs:
     strategy:
       matrix:
         php: [7.4, 8.3]
-        wp-version: [5.7, 6.6, 6.7]
-        wc-version: [9.6.0, 9.7,0]
+        wp-version: [latest]
+        wc-version: [latest]
 
     steps:
       - name: Install SVN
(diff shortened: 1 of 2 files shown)


.github/workflows/php-unit-tests.yml
```diff
@@ -23,15 +23,16 @@ concurrency:
...
+    name: PHP unit tests - PHP ${{ matrix.php }}, WP ${{ matrix.wp-version }}, WC ${{ matrix.wc-version }}
     runs-on: ubuntu-latest
     env:
       WP_CORE_DIR: "/tmp/wordpress/src"
       WP_TESTS_DIR: "/tmp/wordpress/tests/phpunit"
     strategy:
       matrix:
         php: [7.4, 8.3]
-        wp-version: [latest]
+        wp-version: [6.6, 6.7]
```
vahidkay-meta, line 34: I think for wp-version, we want to also test 5.7 as it's in our supported list.

iodic, line 34: @halilozanakgul, would you be open to me adding an update which fetches the last two minor versions dynamically?

We can do this for WP & Woo by fetching the versions from these sources:
- WP: https://api.wordpress.org/core/version-check/1.7/
- Woo: https://api.github.com/repos/woocommerce/woocommerce/releases

@vahidkay-meta and I also talked about updating the PHP versions used so we cover the range between 7.4 and 8.3. I can include that in a commit (but we should rename the PR then to include changes to PHP, WP & Woo versions…).

halilozanakgul: Where can I find the versions we support?

halilozanakgul: Do you think we can just use the `latest` tag for this?

vahidkay-meta: No, we need to cover all of the supported versions. 
You can find the supported versions in readme.txt file.

iodic:
We have removed WP version 5.7 from the tests so the phpunit version is no longer an issue and the tests pass for all of the version combinations we have in place now.

iodic:
The work from this PR is continued on https://github.com/facebook/facebook-for-woocommerce/pull/2913, so I will close this one.

PR #2872, opened 2025-02-06
[Ramp up] Adding test for Product feed Create API call

jczhuoMeta:
Adding tests to Product Feed Creation to as ramp up task

Test: npm run test:php

### DIFF
--- a/tests/Unit/Api/ProductCatalog/ProductFeeds/Create/RequestTest.php
+++ b/tests/Unit/Api/ProductCatalog/ProductFeeds/Create/RequestTest.php
@@ -1,0 +1,21 @@
+<?php
+declare( strict_types=1 );
+/**
+ * Api unit test clas.
+ */
+class RequestTest extends WP_UnitTestCase {
+	/**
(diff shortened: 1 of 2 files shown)


tests/Unit/Api/ProductCatalog/ProductFeeds/Create/RequestTest.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+class RequestTest extends WP_UnitTestCase {
+	/**
+	 * @return void
+	 */
+	public function test_request() {
+		$product          = WC_Helper_Product::create_simple_product();
+		$facebook_product = new WC_Facebook_Product( $product );
+		$product_group_id = 'facebook-product-group-id';
+		$data             = $facebook_product->prepare_product();
```
mradmeta, line 21: So I know the contents of the data variable here dont actually matter because its not typed and acutally used for anything in this test, but it feels weird to set what reads as product data as the data to create a ProductFeed

PR #2873, opened 2025-02-07
release/3.3.3

vahidkay-meta:
This branch contains the release version updates


includes/fbproduct.php
```diff
@@ -726,6 +726,8 @@ public function prepare_product( $retailer_id = null, $type_to_prepare_for = sel
 
 		$custom_fields = $this->get_facebook_specific_fields();
 
+		$custom_fields = $this->get_facebook_specific_fields();
```
iodic, line 729: Looks like this line was duplicated in the previous release, can we delete it?

iodic (changes requested):
Can we just remove the duplicated line in `includes/fbproduct.php:729`?

PR #2875, opened 2025-02-08
Common Feed Upload Framework

Jmencab:
Changes in detail here: https://docs.google.com/document/d/1PaXHKK1PRVtiBEwJtH_8ZP8h0UitDCXZnpFFDWo-eik/edit?tab=t.0#heading=h.m8p14s8600vn
The main idea is to add a FeedManager singleton that will manage all the distinct Feed singletons. The Feed singletons will inherit from an AbstractFeed which will delegate the actual file creation to the Feed Handler or Feed Generator given the admin preferences. With that being said, the sequence will look similar to the existing product flow, but with the above simplifications:

image

Closes # .
Detailed test instructions:
Added an API test as done in this PR: https://github.com/facebook/facebook-for-woocommerce/pull/2841
Integration test not yet possible as dependent on Mice framework addition and Ent creation
Added comments for how to test/dogfood as well as included instructions in the above linked doc
Successful upload via Feed Handler:
https://www.internalfb.com/intern/cat/sessions/642042718535396/
Successful upload via Feed Generator:
https://www.internalfb.com/intern/cat/sessions/642101398529528/

**See branch common_framework_test to see an example proof of concept that utilizes the base classes.**
This will just include what will be checked in via diff.

Unit Tests
A. Run new tests:
./vendor/bin/phpunit --filter ApiTest
B. Run all tests:
npm run test:php 
Manual Tests.
Steps were added via an internal dogfooding doc and requires access to Meta internal tools to validate feed upload sessions status.


iodic:
Looking good so far! I will refrain from commenting on adding extra logging until you push the final version. :)

Jmencab:
> Looking good so far! I will refrain from commenting on adding extra logging until you push the final version. :)

Thank you! I think this is good to review now. I have also added some verification steps in the doc and a simple unit test that matches what was used in the product feed upload PR.

class-wc-facebookcommerce.php
```diff
@@ -187,7 +190,9 @@ public function init() {
 			$this->heartbeat = new Heartbeat( WC()->queue() );
 			$this->heartbeat->init();
 
-			$this->checkout           			 = new WooCommerce\Facebook\Checkout();
+			// Leave feed manger out for now as will just make example feed. Use for dogfooding.
+			$this->feed_manager              = new WooCommerce\Facebook\Feed\FeedManager();
```
nrostrow-meta, line 194: So this won't be present in the diff but will be a requirement to be added when we implement the first feed upload?

facebook-commerce.php
```diff
@@ -2434,6 +2434,17 @@ public function get_facebook_pixel_id() {
...
+	/**
+	 * Retrieve the commerce partner integration ID used in GraphPartnerIntegrationFileUpdatePost call
+	 * Hard coded for now. Update once Mice settings are available.
+	 *
+	 * @return string
+	 * @since 3.5.0
+	 */
+	public function get_commerce_partner_integration_id(): string {
+		return '24316596247984028';
+	}
```
nrostrow-meta, line 2446: Can be removed

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+	public function send_request_to_upload_feed(): void {
+		$name = $this->data_stream_name;
+		$data = array(
+			'url'         => self::get_feed_data_url(),
+			'feed_type'   => $this->feed_type,
+			'update_type' => 'CREATE',
+		);
+
+		try {
+			$cpi_id = \WC_Facebookcommerce::instance()->get_integration()->get_commerce_partner_integration_id();
```
nrostrow-meta, line 144: I think that this can just be replaced with: get_option( 'wc_facebook_commerce_partner_integration_id', '' )

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+	 * The interval in seconds for the feed generation.
+	 *
+	 * @var int
+	 */
+	protected int $gen_feed_interval;
+
+	/**
+	 * Schedules the recurring feed generation.
+	 *
+	 * This method must be implemented by the concrete feed class, usually by providing a recurring interval
```
nrostrow-meta, line 97: Update comment since this logic is now defined in the base class

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+
+	/**
+	 * Regenerates the example feed based on the defined schedule.
+	 *
+	 * @since 3.5.0
+	 */
+	public function regenerate_feed(): void {
+		// Maybe use new ( experimental ), feed generation framework.
```
nrostrow-meta, line 120: I think it would be good to have a comment here explaining the difference between these two flows

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+		// regenerate if the file doesn't exist.
+		if ( ! file_exists( $file_path ) ) {
+			$this->feed_handler->generate_feed_file();
+		}
+
+		try {
+			// bail early if the feed secret is not included or is not valid.
+			if ( self::get_feed_secret() !== Helper::get_requested_value( 'secret' ) ) {
+				throw new PluginException( "{$name} feed: Invalid secret provided.", 401 );
+			}
```
nrostrow-meta, line 215: What exactly is the secret here?

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+
+			// bail early if the file can't be read.
+			if ( ! is_readable( $file_path ) ) {
+
```
nrostrow-meta, line 219: nit: remove newline

includes/Feed/ExampleFeed.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.5.0
+	 */
+	public function __construct() {
+		$this->data_stream_name            = FeedManager::EXAMPLE;
+		$this->gen_feed_interval           = DAY_IN_SECONDS;
+		$this->feed_type                   = 'PRODUCT_RATINGS_AND_REVIEWS';
+		$this->feed_url_secret_option_name = self::OPTION_FEED_URL_SECRET . $this->data_stream_name;
```
nrostrow-meta, line 43: What is the importance/use of this value?

includes/Feed/ExampleFeed.php
```diff
@@ -0,0 +1,100 @@
...
+		try {
+			$cpi_id = \WC_Facebookcommerce::instance()->get_integration()->get_commerce_partner_integration_id();
+			facebook_for_woocommerce()->
+			get_api()->
+			create_common_data_feed_upload( $cpi_id, $data );
+		} catch ( Exception $e ) {
+			// Log the error and continue.
+			\WC_Facebookcommerce_Utils::log( "{$name} feed: Failed to create feed upload request: " . $e->getMessage() );
+		}
+	}
```
nrostrow-meta, line 89: This shouldn't have to be implemented in future feeds since it was added to the base class right?

includes/Feed/ExampleFeed.php
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Gets the secret value that should be included in the legacy WooCommerce REST API URL.
+	 *
+	 * @since 3.5.0
+	 * @return string
+	 */
+	public function get_feed_secret(): string {
+		return 'secret';
+	}
```
nrostrow-meta, line 99: Does this just need to be a unique value for each feed? Does it actually need to be secret?

includes/Feed/ExampleFeedGenerator.php
```diff
@@ -0,0 +1,207 @@
...
+	/**
+	 * Retrieves items for a specific batch.
+	 *
+	 * @param int   $batch_number The batch number.
+	 * @param array $args Additional arguments.
+	 * @return array The items for the batch.
+	 * @inheritdoc
+	 * @since 3.5.0
+	 */
+	protected function get_items_for_batch( int $batch_number, array $args ): array {
```
nrostrow-meta, line 74: For the promotions, ratings and reviews, and site navigation feeds my understanding is that we wouldn't do anything with this function right? The logic to fetch the necessary data and write it to the file would happen in the feed handler classes?

class-wc-facebookcommerce.php
Jmencab: actually, I think we settled on creating the feed manager but not making any feeds in its constructor
s

facebook-commerce.php
```diff
@@ -2434,6 +2434,17 @@ public function get_facebook_pixel_id() {
...
 	}
 
+	/**
+	 * Retrieve the commerce partner integration ID used in GraphPartnerIntegrationFileUpdatePost call
+	 * Hard coded for now. Update once Mice settings are available.
+	 *
+	 * @return string
+	 * @since 3.5.0
+	 */
+	public function get_commerce_partner_integration_id(): string {
```
Jmencab, line 2444: will not be present in final diff

includes/API.php
```diff
@@ -101,6 +102,35 @@ protected function perform_request( $request ): API\Response {
 		return parent::perform_request( $request );
 	}
 
+	/**
```
Jmencab, line 105: will not be present in final diff

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @return string
+	 */
+	public function get_feed_secret(): string {
+/*		$secret = get_option( $this->feed_url_secret_option_name, '' );
+		if ( ! $secret ) {
+			$secret = wp_hash( 'example-feed-' . time() );
+			update_option( $this->feed_url_secret_option_name, $secret );
+		}
+
+		return $secret;*/
```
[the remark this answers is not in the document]
Jmencab: This snapshot is what you need for testing this. need to be able to know what the "secret" will be so you can easily hit the url yourself. 
For the actual implementation, it will be the commented out code

includes/Feed/AbstractFeed.php
Jmencab: it's a hash that gets appended to the file names. you verify the secret here to make sure it's the correct file

includes/Feed/AbstractFeed.php
```diff
@@ -0,0 +1,255 @@
...
+			header( 'Content-Type: text/csv; charset=utf-8' );
+			header( 'Content-Description: File Transfer' );
+			header( 'Content-Disposition: attachment; filename="' . basename( $file_path ) . '"' );
+			header( 'Expires: 0' );
+			header( 'Cache-Control: must-revalidate, post-check=0, pre-check=0' );
+			header( 'Pragma: public' );
+			header( 'Content-Length:' . filesize( $file_path ) );
+
+			// phpcs:ignore
+			$file = @fopen( $file_path, 'rb' );
```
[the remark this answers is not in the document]
Jmencab: `if ( \WC_Facebookcommerce_Utils::is_fpassthru_disabled() || ! @fpassthru( $file ) ) {`
Is doing a lot... if fpassthru returns true, then it was passed to Meta, if it did not work then it needs to be streamed

includes/Feed/ExampleFeed.php
Jmencab: true, but you can still override if need be. In this case, I'm overriding for the 'url' field in the $data array

includes/Feed/ExampleFeed.php
Jmencab: It's a secret in so far as it shouldn't be a url others can easily go onto. For testing, we know to hit wc-api...secret=secret, but that wouldn't be good for the actual public urls. 
For each implementing class, the secret value will change per instantiation since it is based on the time it is created and then hashed. So if the correct class is calling it, it should never through that 401 error in handle_feed_data_request but would if someone guessing the url guessed incorrectly.

includes/Feed/ExampleFeedGenerator.php
Jmencab: It will depend on the specifics of each data stream you want to support. Think of the batch number as a page number. if you have data that's paginated, it makes sense to use with this. The sql queries in the product case are numbered 1...n. So for ratings and reviews could use this if for example, 
 Scenario: Wp_Product[ [reviewId: ....], i.e. the review info is nested in the product info, you can almost exactly copy the product implementation and just get the necessary review information out of it.

PR #2918, opened 2025-03-04
Include root directory files in code coverage report

carterbuce:
Our code coverage report currently only runs for files in the `includes` directory. We do have tests, however, for some of the files in the root directory and should be tracking the coverage for those as well.

### Screenshots:

image

vendor/bin/phpunit --coverage-html=reports/coverage 

### Changelog entry
Including root directory php files in phpunit code coverage report

>

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -20,6 +20,12 @@
 	<coverage>
 		<include>
 			<directory>includes</directory>
+			<file>./class-wc-facebookcommerce.php</file>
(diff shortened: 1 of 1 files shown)


carterbuce:
@vahidkay-meta, any ideas why the code sniff and unit test workflows are stuck in pending here? I don't see any workflows started for this PR, even after re-pushing it. I'm guessing shipit would break unless these are in a completed state

PR #2930, opened 2025-03-07
Improve custom checkout UI

ajello-meta:
This PR introduces an improved UI for the custom checkout permalink of the user's WooCommerce store. These changes renders the custom checkout UI to be exactly the same as the native checkout UI at `/checkout`.

Original template created in https://github.com/facebook/facebook-for-woocommerce/pull/2887.

### Screenshots:
Custom Checkout (Before):
Screenshot 2025-03-11 at 5 59 44 PM

Custom Checkout (After):
Screenshot 2025-03-11 at 5 57 01 PM

Default Checkout:
Screenshot 2025-03-11 at 5 56 54 PM

1. Test checkout permalink with one product (e.g. `/fb-checkout?products=<ID>:<#>`)
2. Test checkout permalink with two product (e.g. `/fb-checkout?products=<ID>:<#>,<ID>:<#>`)
3. Test checkout permalink with products and a coupon (e.g. `/fb-checkout?products=<ID>:<#>,:<ID>:<#>&coupon=<CODE>`)
4. Test default checkout with products and a coupon (e.g. `/checkout`)


strygo:
Do you have screenshots for how this looks on mobile?

sol-loup:
we also had discussed that theme css variables (WP-primitives such as padding, etc) should be transferred over as well. Your site has a padding around the normal website checkout cart view which I'm not seeing here for the SA-Off view

PR #2947, opened 2025-03-14
Fix feed upload error "feed_column_count_mismatch" when product has multiple colors separated by a coma

mshymon:

When rolling out product sync via Feeds, we noticed a fatal upload error for products that have value of a color to be multiple colors written with commas (e.g. color value: Yellow,Blue,Red). This case right leads to fatal error "feed_column_count_mismatch" on feed upload and the product is deleted from the catalog.

The fix is simple what we already fo for other fields like `name` or `description`: wrap it up with quotes by calling `format_string_for_feed`. 
Applying this not only to color but some other fields too. Manually tested feed generation and feed upload success.

### Additional details:
Screenshot shows multi color  value for a single product written via comas that leads to feed sync issue before this fix. 
 
Screenshot 2025-03-14 at 17 10 11

1. Create or update an existing product to have multi color value written via comas, as per screenshot.
2. Trigger feed generation and check for errors.  

### Changelog entry

> Fix - feed upload error "feed_column_count_mismatch" when product has multiple colors separated by a coma.

### DIFF
--- a/includes/fbproductfeed.php
+++ b/includes/fbproductfeed.php
@@ -512,27 +512,27 @@ class WC_Facebook_Product_Feed {
 		static::get_value_from_product_data( $product_data, 'url' ) . ',' .
 		static::format_string_for_feed( static::get_value_from_product_data( $product_data, 'product_type' ) ) . ',' .
 		static::format_string_for_feed( static::get_value_from_product_data( $product_data, 'brand' ) ) . ',' .
-		static::format_price_for_feed(
+		static::format_string_for_feed( static::format_price_for_feed(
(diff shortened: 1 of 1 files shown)


sol-loup:
Do we properly handle comma-based escaping for all fields in the feed system? Was this the only problematic field?

mshymon:
> Do we properly handle comma-based escaping for all fields in the feed system? Was this the only problematic field?

@sol-loup Hey Paul! I went through a number of cases in prod and all I saw was the color, so raised this specifically to stop that issue asap. I was thinking to raise another PR to preemptively add this for some other fields and test them.

PR #2957, opened 2025-03-21
[WAUM][3/n]Add a rest api endpoint to receive whatsapp webhooks

sharunaanandraj:
## Description

We are building a Meta Hosted Connectbridge that will recieve the access tokens and waba info at the end of the onboarding flow and then forward it to the end business.  In order to receive this onboarding information we create a POST REST API in the plugin which will receive the whatsapp waba information and access token and store it in the plugin database.

### Type of change

## Screenshots
https://github.com/user-attachments/assets/59f4037e-e51f-47d8-8bda-bb2593b7f933

## Test instructions
Used postman to call the REST API. Used WC_Logger to check if the params are passed as expected. 

## Changelog entry
creating rest api to receive whatsapp onboarding info

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -96,6 +96,9 @@ class WC_Facebookcommerce extends WooCommerce\Facebook\Framework\Plugin {
 	/** @var WooCommerce\Facebook\Handlers\WebHook webhook handler */
 	private $webhook_handler;
 
+	/** @var WooCommerce\Facebook\Handlers\Whatsapp_WebHook whatsapp webhook handler */
+	private $whatsapp_webhook_handler;
+
 	/** @var WooCommerce\Facebook\Commerce commerce handler */
 	private $commerce_handler;
 
@@ -189,7 +192,6 @@ class WC_Facebookcommerce extends WooCommerce\Facebook\Framework\Plugin {
 
 			$this->heartbeat = new Heartbeat( WC()->queue() );
 			$this->heartbeat->init();
-
 			$this->feed_manager              = new WooCommerce\Facebook\Feed\FeedManager();
 			$this->checkout           		 = new WooCommerce\Facebook\Checkout();
 			$this->product_feed              = new WooCommerce\Facebook\Products\Feed();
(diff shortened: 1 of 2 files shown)


woo-ardsouza (approved):

includes/Handlers/Whatsapp_Webhook.php
```diff
@@ -0,0 +1,72 @@
+<?php
+// phpcs:ignoreFile
```
iodic, line 2: Can you remove these PHPCS ignore comments from the new files you are adding? It will be better to have PHPCS checks sooner rather than later so we can keep the code style cleaner and compliant.

sharunaanandraj: Updated!

includes/Handlers/Whatsapp_Webhook.php
```diff
@@ -0,0 +1,71 @@
...
+class Whatsapp_Webhook {
+
+	/**
+	 * Constructs a new Whatsapp WebHook.
+	 *
+	 * @param \WC_Facebookcommerce $plugin Plugin instance.
+	 *
+	 * @since 2.3.0
+	 */
+	public function __construct( \WC_Facebookcommerce $plugin ) {
```
iodic, line 29: Will the `$plugin` variable be used in the upcoming changes?

class-wc-facebookcommerce.php
```diff
@@ -222,6 +224,7 @@ public function init() {
 			$this->connection_handler = new WooCommerce\Facebook\Handlers\Connection( $this );
 			new WooCommerce\Facebook\Handlers\MetaExtension();
 			$this->webhook_handler    = new WooCommerce\Facebook\Handlers\WebHook( $this );
+			$this->whatsapp_webhook_handler    = new WooCommerce\Facebook\Handlers\Whatsapp_Webhook( $this );
```
iodic, line 227: Can you update the alignment here so that all equals symbols are in line?

The file has the PHPCS ignore comment so the editor isn't linting it. (Please leave the comment in place for now.)

includes/Handlers/Whatsapp_Webhook.php
sharunaanandraj: Yes. This will be used to add logging in future diffs.

class-wc-facebookcommerce.php
sharunaanandraj: Done.

iodic (approved):

PR #2975, opened 2025-03-27
Add GitHub Actions Workflow for JavaScript Unit Testing

devbodaghe:
## Pull Request Description: JavaScript Unit Tests Workflow

This pull request introduces a GitHub Actions workflow for running JavaScript unit tests on Node.js version 16. The workflow is designed to enhance continuous integration by automatically executing tests on all pushes and pull requests. Below are the key features and steps included in this workflow:

- **Concurrency Management**: 
  - Ensures that only one instance of the workflow runs per branch by canceling any in-progress jobs for the same workflow and reference.

- **Node.js Setup**:
  - Configures the environment to use Node.js version 16, as specified in the `package.json` engine requirements.
  - Utilizes caching for npm dependencies to improve efficiency.

- **Linux Compatibility**:
  - Prepares `package.json` for Linux by removing the `fsevents` dependency, which is not required in Linux environments.

- **Dependency Installation**:
  - Implements multiple strategies to install dependencies, with fallback options to handle potential installation issues.

- **Version Verification**:
  - Confirms the versions of Node.js and npm being used to ensure compatibility.

- **Test Execution**:
  - Runs JavaScript unit tests, with a mechanism to continue the workflow even if tests fail, while issuing a warning.

- **Dependency Check**:
  - Skips tests if dependencies fail to install, providing a notification to the user.

- **Cleanup**:
  - Restores the original `package.json` and removes temporary files after the workflow execution.

This setup aims to ensure robust testing and compatibility across different environments, thereby enhancing the reliability and quality of the codebase.

### DIFF
--- a/.github/workflows/js-unit-tests.yml
+++ b/.github/workflows/js-unit-tests.yml
@@ -1,0 +1,83 @@
+name: JavaScript Unit Tests
+
+on: [ push, pull_request ] # Run on all pushes and PRs
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  JSUnitTests:
+    name: JavaScript unit tests - Node ${{ matrix.node }}
+    runs-on: ubuntu-latest
+    env:
+      NODE_ENV: test
+    strategy:
+      matrix:
+        node: [16]  # Using Node 16 as specified in package.json engine requirements
+      fail-fast: false  # Continue with other tests if one fails
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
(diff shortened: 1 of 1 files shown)


.github/workflows/js-unit-tests.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'npm'
+
+      - name: Prepare package.json for Linux
```
iodic, line 30: Could we skip this step and install dependencies in the next step like this: `npm install --no-optional --ignore-scripts`? That should skip `fsevents` when not run a Mac machine.

PR #3003, opened 2025-04-03
[WAUM][6/n] Enable Consent collection setting during whatsapp onboarding flow

sharunaanandraj:
## Description

When a business onboards to whatsapp utility messages and clicks on collect in the consent collection section, we add a setting that will decide if we can show consent collection checkbox to the customers in the checkout flow.

### Type of change

## Changelog entry
Add consent sollection setting for whatsapp utility message integration

## Test Plan
Clicking on "Collect" adds a setting in the woocommerce merchant DB with enabled status

## Screenshots

https://github.com/user-attachments/assets/a81ae020-c428-47c5-9e3e-be4de02a10b4

### DIFF
--- a/includes/AJAX.php
+++ b/includes/AJAX.php
@@ -161,6 +161,9 @@ class AJAX {
 
 
 	public function whatsapp_consent_collection_enable() {
+		if ( ! check_ajax_referer( 'facebook-for-wc-whatsapp-consent-nonce', 'nonce', false ) ) {
+    	wp_send_json_error( 'Invalid security token sent.' );
+		}
 		if(get_option('wc_facebook_whatsapp_consent_collection_setting_status') !== 'enabled') {
 			update_option('wc_facebook_whatsapp_consent_collection_setting_status', 'enabled');
 		}


woo-ardsouza (approved):

includes/AJAX.php
```diff
@@ -157,6 +160,14 @@ public function get_sync_status() {
 	}
 
 
+	public function whatsapp_consent_collection_enable() {
```
iodic, line 163: We might consider adding nonce validation at the beginning of the function, something like this:

```php
if ( ! check_ajax_referer( 'facebook-for-wc-whatsapp-consent-nonce', 'nonce', false ) ) {
    wp_send_json_error( 'Invalid security token sent.' );
    exit;
}
```

sharunaanandraj: Updated

class-wc-facebookcommerce.php
sharunaanandraj, line 1: Note: This file changes are from a previously approved PR merged into this branch
https://github.com/facebook/facebook-for-woocommerce/pull/3004/files

iodic (approved):

PR #3004, opened 2025-04-03
[WAUM][7/n]Add consent collection to customer checkout flow if it is …

sharunaanandraj:
## Description

When consent collection setting is enabled, add checkbox in checkout flow to collect consent for billing and shipping phone numbers

### Type of change

## Changelog entry
Add consent collection checkbox to customer checkout flow in woocommerce

## Test Plan
Enabling consent collection also enables checkbox in customer checkout flow

## Screenshots

https://github.com/user-attachments/assets/21607a26-8b83-4be3-87c8-2d3e13cef019

### DIFF
--- a/class-wc-facebookcommerce.php
+++ b/class-wc-facebookcommerce.php
@@ -169,6 +169,7 @@ class WC_Facebookcommerce extends WooCommerce\Facebook\Framework\Plugin {
 		add_action( 'init', array( $this, 'get_integration' ) );
 		add_action( 'init', array( $this, 'register_custom_taxonomy' ) );
 		add_action( 'add_meta_boxes_product', array( $this, 'remove_product_fb_product_set_metabox' ), 50 );
+		add_action( 'woocommerce_init', array($this, 'add_whatsapp_consent_checkout_fields'));
 		add_filter( 'fb_product_set_row_actions', array( $this, 'product_set_links' ) );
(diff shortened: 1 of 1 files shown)


woo-ardsouza (approved):

class-wc-facebookcommerce.php
```diff
@@ -169,6 +169,7 @@ public function init() {
 		add_action( 'init', array( $this, 'get_integration' ) );
 		add_action( 'init', array( $this, 'register_custom_taxonomy' ) );
 		add_action( 'add_meta_boxes_product', array( $this, 'remove_product_fb_product_set_metabox' ), 50 );
+		add_action( 'woocommerce_init', array($this, 'add_whatsapp_consent_checkout_fields'));
```
iodic, line 172: The effects on this are negligible but we might move the condition here instead of having it inside of the function below.

So we only add the action if the option is set:

```
if ( get_option( 'wc_facebook_whatsapp_consent_collection_setting_status', 'disabled' ) === 'enabled') {
    add_action( 'woocommerce_checkout_fields', array( $this, 'add_whatsapp_consent_checkout_fields' ) );
}
```

iodic (approved):

class-wc-facebookcommerce.php
sharunaanandraj: More processing is required within this function will do this in follow up diffs.

Issue #3008, opened 2025-04-03
Fatal Error


cneary1292:
WordPress caught an error with one of your plugins, Facebook for WooCommerce.

[03-Apr-2025 20:23:45 UTC] PHP Fatal error: Uncaught TypeError: array_multisort(): Argument #1 ($array) must be an array or a sort flag in /home/e8upgxj/public_html/wp-content/plugins/facebook-for-woocommerce/includes/Admin/Enhanced_Catalog_Attribute_Fields.php:148

The error was making it impossible to edit products on woocommerce website. We diasbled the Facebook for WooCommerce and it resolved.

vinkmeta:
Hey, This issue will be fixed today in the new plugin version release 3.4.6

PR #3012, opened 2025-04-04
Fix product attribute sort error

iodic:
## Description

This changeset fixes an attribute sort fatal error which prevents users from editing products.

We now check if there are elements in the attribute array that can be sorted.

This is outlined in issue #3008.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I followed general Pull Request best practices. Meta employees to follow this [wiki]([url](https://fburl.com/wiki/2cgfduwc))
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests and all the new and existing unit tests pass locally with my changes

## Changelog entry

Fix product attribute sort error which prevented product edits in certain scenarios.

## Test Plan

Install the artifact built from this version. The error can manifest in two places:
- In the WooCommerce logs for the plugin after installing.
- It will not let you edit a product's status (published/draft). In this test, the browser console will show a 500 Internal Server Error for the Ajax request.

With this version, you should not see the error.

### DIFF
--- a/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
+++ b/includes/Admin/Enhanced_Catalog_Attribute_Fields.php
@@ -135,6 +135,7 @@ class Enhanced_Catalog_Attribute_Fields {
 			'default'        => 100,
 		);
 
+		$priority = array(); // Initialize priority array.
 		foreach ( $recommended_attributes as $key => $attribute ) {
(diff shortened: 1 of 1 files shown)


vinkmeta (approved):
Looks good!

PR #3016, opened 2025-04-04
[WAUM][8/n] Polling back-end to check if whatsapp onboarding info Received

sharunaanandraj:
## Description
When business completes whatsapp onboarding, meta will send the onboarding info to woocommerce with the whatsapp onboarding details. Once these details are received we have to update the front-end with progress on their onboarding (i.e) let end business know that their onboarding was successful. We do this by initiating a polling session when business click on connect whatsapp button in woocommerce and continuously check if info is received at the back-end. We stop polling when info is received.

Note: The diff to do this via E2E(onboarding and receiving the webhook is not in production yet), for now we send data via postman emulating the data sent by meta.

### Type of change

## Changelog entry
Add polling to check progress update of whatsapp onboarding

## Test Plan
- Initiating onboarding flow kickstarts the polling session
- When info is received polling session stops and response is success.

## Screenshots
https://github.com/user-attachments/assets/8ca00288-e365-4b94-8d81-425eaa6abc0e

### DIFF
--- a/assets/js/admin/whatsapp-connection.js
+++ b/assets/js/admin/whatsapp-connection.js
@@ -10,11 +10,29 @@
 jQuery( document ).ready( function( $ ) {
     // handle the whatsapp connect button click should open hosted ES flow
 	$( '#woocommerce-whatsapp-connection' ).click( function( event ) {
-        // dummy values for app id and config id, will be replaced in upcoming diffs
-        const APP_ID = '18402284156271';
-        const CONFIG_ID = '17502287264684';
+        const APP_ID = '474166926521348'; // WOO_COMMERCE_APP_ID
+        const CONFIG_ID = '1237758981048330'; // WOO_COMMERCE_WHATSAPP_CONFIG_ID
         const HOSTED_ES_URL = `https://business.facebook.com/messaging/whatsapp/onboard/?app_id=${APP_ID}&config_id=${CONFIG_ID}`;
         window.open( HOSTED_ES_URL);
+        updateProgress();
     });
+
+    function updateProgress() {
+        $.post( facebook_for_woocommerce_whatsapp_onboarding_progress.ajax_url, {
+			action: 'wc_facebook_whatsapp_onboarding_progress_check',
+			nonce:  facebook_for_woocommerce_whatsapp_onboarding_progress.nonce
+		}, function ( response ) {
+
+            // check if the response is success (i.e. onboarding is completed)
+            if ( response.success ) {
+                // TODO: if success, update the UI with the onboarding succeeded
+				console.log( 'success', response );
+			} else {
+                console.log('Failure. Checking again in 1 second', response);
+                setTimeout( updateProgress, 1000 );
+            }
+		} );
+
+    }
 
 } );
--- a/assets/js/admin/whatsapp-connection.js
+++ b/assets/js/admin/whatsapp-connection.js
@@ -17,7 +17,7 @@ jQuery( document ).ready( function( $ ) {
         updateProgress(0,1800000); // retry for 30 minutes
     });
 
-    function updateProgress(retryCount = 0, maxRetries = 1800000) {
+    function updateProgress(retryCount = 0, pollingTimeout = 1800000) {
         $.post( facebook_for_woocommerce_whatsapp_onboarding_progress.ajax_url, {
 			action: 'wc_facebook_whatsapp_onboarding_progress_check',
 			nonce:  facebook_for_woocommerce_whatsapp_onboarding_progress.nonce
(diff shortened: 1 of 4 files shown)


assets/js/admin/whatsapp-connection.js
```diff
@@ -10,11 +10,29 @@
 jQuery( document ).ready( function( $ ) {
     // handle the whatsapp connect button click should open hosted ES flow
 	$( '#woocommerce-whatsapp-connection' ).click( function( event ) {
-        // dummy values for app id and config id, will be replaced in upcoming diffs
-        const APP_ID = '18402284156271';
-        const CONFIG_ID = '17502287264684';
+        const APP_ID = '474166926521348'; // WOO_COMMERCE_APP_ID
+        const CONFIG_ID = '1237758981048330'; // WOO_COMMERCE_WHATSAPP_CONFIG_ID
         const HOSTED_ES_URL = `https://business.facebook.com/messaging/whatsapp/onboard/?app_id=${APP_ID}&config_id=${CONFIG_ID}`;
```
woo-ardsouza, line 15: Is this the tech provider config id? Is it the same for all woocommerce businesses?

sharunaanandraj: Yes. This will be the same for all businesses.

includes/AJAX.php
```diff
@@ -168,6 +171,23 @@ public function get_sync_status() {
...
+	/**
+	 * Checks if the onboarding for whatsapp is complete once business has initiated onboarding.
+	 *
+	 * @internal
+	 *
+	 * @since 1.10.0
+	 */
+	public function whatsapp_onboarding_progress_check() {
+		if ( ! check_ajax_referer( 'facebook-for-wc-whatsapp-onboarding-progress-nonce', 'nonce', false ) ) {
+			wp_send_json_error( 'Invalid security token sent.' );
```
woo-ardsouza, line 183: how is this handled?

sharunaanandraj: This will only be handled in case there is an attacker trying to make this AJAX call with invalid nonce. When we do the polling, we should always have the correct nonce sent to the back-end.

woo-ardsouza (approved):

assets/js/admin/whatsapp-connection.js
```diff
@@ -10,11 +10,29 @@
...
+			nonce:  facebook_for_woocommerce_whatsapp_onboarding_progress.nonce
+		}, function ( response ) {
+
+            // check if the response is success (i.e. onboarding is completed)
+            if ( response.success ) {
+                // TODO: if success, update the UI with the onboarding succeeded
+				console.log( 'success', response );
+			} else {
+                console.log('Failure. Checking again in 1 second', response);
+                setTimeout( updateProgress, 1000 );
```
iodic, line 32: I assume this isn't final but to suggest adding a retry limit so we don't end up with runaway process on failure.

To define the function like this:

```js
function updateProgress(retryCount = 0, maxRetries = 20) {
```

And then to check if we are exceeding the retry limit (whichever number of retries makes sense).

sharunaanandraj: Updated.

assets/js/admin/whatsapp-connection.js
```diff
@@ -10,11 +10,33 @@
...
+            if ( response.success ) {
+                // TODO: if success, update the UI with the onboarding succeeded
+				console.log( 'success', response );
+			} else {
+                console.log('Failure. Checking again in 1 second:', response, ', retry attempt:', retryCount, 'maxRetries', maxRetries);
+                if(retryCount >= maxRetries) {
+                    console.log('Max retries reached. Aborting.');
+                    return;
+                }
+                setTimeout( updateProgress(retryCount+1, maxRetries), 1000 );
```
iodic, line 36: I am not sure if this executes properly, a better way would be to wrap the call in an anonymous function:

```js
setTimeout(function() {
    updateProgress(retryCount + 1, maxRetries);
}, 1000);
```

If we leave the call as is, it can result in rapid requests that fire continuously — with the current code, `updateProgress()` is called immediately, not after the 1000ms.

Should we also maybe set a longer delay between retries?

assets/js/admin/whatsapp-connection.js
```diff
@@ -10,11 +10,33 @@
...
-        const APP_ID = '18402284156271';
-        const CONFIG_ID = '17502287264684';
+        const APP_ID = '474166926521348'; // WOO_COMMERCE_APP_ID
+        const CONFIG_ID = '1237758981048330'; // WOO_COMMERCE_WHATSAPP_CONFIG_ID
         const HOSTED_ES_URL = `https://business.facebook.com/messaging/whatsapp/onboard/?app_id=${APP_ID}&config_id=${CONFIG_ID}`;
         window.open( HOSTED_ES_URL);
+        updateProgress(0,1800000); // retry for 30 minutes
     });
 
+    function updateProgress(retryCount = 0, maxRetries = 1800000) {
```
iodic, line 20: Let's consider renaming `maxRetries` to `pollingTimeout` as it indicates more clearly that the var refers to time, rather than the number of retries.

sharunaanandraj: done

assets/js/admin/whatsapp-connection.js
sharunaanandraj: Updated. Made retries after 5 seconds.

PR #3070, opened 2025-04-18
[WAUM][15/n] Redirect to Utility settings page when Finish Button is clicked

sharunaanandraj:
## Description
Upon finish button click on the onboarding flow we need to redirect to the Utility Settings Page

### Type of change

## Changelog entry
WAUM finish button redirect

## Test Plan

https://github.com/user-attachments/assets/164150d6-169d-416f-8bfc-6264bc38529f

### DIFF
--- a/assets/js/admin/whatsapp-finish.js
+++ b/assets/js/admin/whatsapp-finish.js
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package FacebookCommerce
+ */
+
+jQuery( document ).ready( function( $ ) {
+    // handle the whatsapp finish button click
+	$( '#wc-whatsapp-onboarding-finish' ).click( function( event ) {
+        // TODO: call the connect API to create configs and check payment
+        // If error, show banner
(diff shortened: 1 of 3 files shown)


woo-ardsouza (approved):

assets/js/admin/whatsapp-finish.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package FacebookCommerce
+ */
+
+jQuery( document ).ready( function( $ ) {
+    // handle the whatsapp finish button click
+	$( '#wc-whatsapp-onboarding-finish' ).click( function( event ) {
```
iodic, line 12: Note for the upcoming updates, can you handle the redirect destination by setting something like this in the PHP file which contains the button:

```php
$cancel_url = add_query_arg(
    array(
        'page' => self::PAGE_ID,
        'tab'  => self::ID,
        'view' => 'utility_settings'
    ),
    admin_url('admin.php')
);
```

sharunaanandraj: We need to do some AJAX processing calling Meta APIs before we can redirect . We will redirect only if the API Post call is a success, if it is not a success we need to show a banner in the UI in the existing view. So this particular change cannot be moved to the PHP button file.

iodic (approved):

PR #3106, opened 2025-04-28
[WAUM][24/n] Gracefully handle errors during Whatsapp connect API call

sharunaanandraj:
## Description
When we call connect API if an unknown error occurs then the notice was not show since $error_object->error was empty and exception was thrown. Fetch the error->message on API call for unknown errors or fallback to a default error message.

### Type of change

## Changelog entry
Gracefully handle errors during Whatsapp connect API call

## Test Plan
Screenshot 2025-04-28 at 2 26 44 PM

### DIFF
--- a/includes/Admin/Settings_Screens/Whatsapp_Utility.php
+++ b/includes/Admin/Settings_Screens/Whatsapp_Utility.php
@@ -359,14 +359,20 @@ class Whatsapp_Utility extends Abstract_Settings_Screen {
 			<div class="card-item event-config">
 					<div>
 						<div class="event-config-heading-container">
-							<h3><?php esc_html_e( 'Checkbox', 'facebook-for-woocommerce' ); ?></h3>
+							<h3><?php esc_html_e( 'Add WhatsApp option at checkout', 'facebook-for-woocommerce' ); ?></h3>
 							<div class="event-config-status on-status">
 								<?php esc_html_e( 'On', 'facebook-for-woocommerce' ); ?>
-							</div>
 						</div>
-						<p><?php esc_html_e( 'Removing this means you won\'t be able to send messages to your customers.', 'facebook-for-woocommerce' ); ?></p>
 					</div>
-					</div>
+					<span class="consent-update-card-subcontent">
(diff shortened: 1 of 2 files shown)


includes/Admin/Settings_Screens/Whatsapp_Utility.php
```diff
@@ -269,7 +269,7 @@ class="button"
 			</div>
 		</div>
 		<div class="error-notice-wrapper">
-			<div id="payment-method-error-notice fbwa-hidden-element"></div>
+			<div id="payment-method-error-notice"></div>
```
woo-ardsouza, line 272: Does payment-method-error-notice class already have display:none?

includes/Handlers/WhatsAppUtilityConnection.php
```diff
@@ -103,7 +103,7 @@ public static function wc_facebook_whatsapp_connect_utility_messages_call( $waba
 		if ( is_wp_error( $response ) || 200 !== wp_remote_retrieve_response_code( $response ) ) {
 			$error_data    = explode( "\n", wp_remote_retrieve_body( $response ) );
 			$error_object  = json_decode( $error_data[0] );
-			$error_message = $error_object->error->error_user_title;
+			$error_message = $error_object->error->error_user_title ?? $error_object->error->message ?? 'Something went wrong. Please try again later!';
```
woo-ardsouza, line 106: "<?php esc_html_e('Something went wrong. Please try again later!', 'facebook-for-woocommerce' )"?

includes/Admin/Settings_Screens/Whatsapp_Utility.php
sharunaanandraj: payment-method-error-notice is an ID and not a style class. This does not require it to be hidden before hand as the html tag adding the notice is added only in JS file when there is an error.

includes/Handlers/WhatsAppUtilityConnection.php
sharunaanandraj: This error_message is just a default value for logging and will not be shown on the front-end. The front-end text is added via esc_html_e through the localise_script i18

woo-ardsouza (approved):

iodic (approved):

PR #3109, opened 2025-04-29
[WAUM][26/n]Fix the onclick null property when on onboarding view

sharunaanandraj:
## Description
In the onboarding view, the Confirm and Cancel Button are not available causing null error. Fix the null property issue when on onboarding view by checking for property before manipulating the onclick callback.

### Type of change
- Bug fix (non-breaking change which fixes an issue)

## Changelog entry
Fix the onclick null property when on onboarding view

## Test Plan

https://github.com/user-attachments/assets/810c089c-31f7-4c16-8cce-b05e2f25e3bd

### DIFF
--- a/assets/js/admin/whatsapp-consent-remove.js
+++ b/assets/js/admin/whatsapp-consent-remove.js
@@ -22,27 +22,32 @@ jQuery( document ).ready( function( $ ) {
         event.preventDefault();
     });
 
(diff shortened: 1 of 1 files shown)


woo-ardsouza (approved):
lgtm!

iodic (approved):

PR #3117, opened 2025-05-01
[WAUM] Changes to call Event Configs Post API on Saving Event Settings

woo-ardsouza:
## Description
Changes to create or update event configs when the Save button is clicked in the Manage Event view. When the Save button is clicked:
* We call the POST API to create or update the event config object for the selected event with the selected language and selected status option
* We update the database with the created event config id if the status is active. Otherwise, we set the event config in the database to be null.

### Type of change

## Changelog entry

Save function for Manage Events

## Test Plan
1. Select WhatsApp Utility Tab
2. In the Utility Setting View, click Manage for Order Confirmation event
3. Select Option to switch off event, and click Save
4. Validate db has null value for wc_facebook_wa_order_placed_event_config_id
5. Select Option to switch on event, and click Save
6. Validate db has value for wc_facebook_wa_order_placed_event_config_id
7. Validate success API logs in WooCommerce logs

## Screen Recording
https://github.com/user-attachments/assets/11409920-0ea1-4804-9fe9-7f9f511b2bb2

## Logs
Screenshot 2025-04-30 at 8 48 31 PM

### DIFF
--- a/assets/js/admin/whatsapp-events.js
+++ b/assets/js/admin/whatsapp-events.js
@@ -41,4 +41,24 @@ jQuery( document ).ready( function( $ ) {
             }
         });
     });
+
+    $('#woocommerce-whatsapp-save-order-confirmation').click(function (event) {
+        var languageValue = $("#manage-event-language").val();
+        var statusValue = $('input[name="template-status"]:checked').val();
+        console.log('Save confirmation clicked: ', languageValue, statusValue);
+        $.post(facebook_for_woocommerce_whatsapp_events.ajax_url, {
+            action: 'wc_facebook_whatsapp_upsert_event_config',
+            nonce: facebook_for_woocommerce_whatsapp_events.nonce,
+            event: 'ORDER_PLACED',
+            language: languageValue,
+            status: statusValue
+        }, function (response) {
+            //TODO: Add Error Handling
+            let url = new URL(window.location.href);
+            let params = new URLSearchParams(url.search);
+            params.set('view', 'utility_settings');
+            url.search = params.toString();
+            window.location.href = url.toString();
+        });
+    });
 });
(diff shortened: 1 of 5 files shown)


assets/js/admin/whatsapp-events.js
```diff
@@ -41,4 +41,24 @@ jQuery( document ).ready( function( $ ) {
...
+        }, function (response) {
+            //TODO: Add Error Handling
+            let url = new URL(window.location.href);
+            let params = new URLSearchParams(url.search);
+            params.set('view', 'utility_settings');
+            url.search = params.toString();
+            window.location.href = url.toString();
+        });
+    });
 });
```
sharunaanandraj, line 64: Add new line ?

includes/Handlers/WhatsAppUtilityConnection.php
```diff
@@ -122,4 +134,79 @@ public static function wc_facebook_whatsapp_connect_utility_messages_call( $waba
...
+	 * Makes an API call to Whatsapp Utility Event Configs Post API to create or update Event Configs
+	 *
+	 * @param string $event Order Management Event
+	 * @param string $language Language Code
+	 * @param string $status ACTIVE or INACTIVE
+	 * @param string $bisu_token the BISU token received in the webhook
+	 */
+	public static function post_whatsapp_utility_messages_event_configs_call( $event, $language, $status, $bisu_token ) {
+		// TODO: Update Harcoded Integration id to Option value
+		$integration_id = 1486573012315320;
```
sharunaanandraj, line 149: At what point will we get the integration_id ?

sharunaanandraj (approved):
Looks great!

sharunaanandraj:
From the screenrecording I see that wc_facebook_wa_order_placed_event_config_id is initially null, at what point is this set to null ?

includes/Handlers/WhatsAppUtilityConnection.php
woo-ardsouza: The Connection API returns the integration_id

woo-ardsouza:
> From the screenrecording I see that wc_facebook_wa_order_placed_event_config_id is initially null, at what point is this set to null ?

We are saving the event_config_id only if the user selects the "Send order confirmation message". If the user selects "Turn off order confirmation", the value is set as null. I had set the event config as inactive and then active in my test plan.

iodic (approved):

PR #3135, opened 2025-05-03
[WAUM] Minor updates to Manage Events flow

woo-ardsouza:
## Description
This PR updates the following in the Manage Events
* Removes hardcoded integration config id and gets the option value
* Sets the language selector to created event language. If the language is not available, the language is defaulted to English
* Updated the url suffix parameter to {{1}} and removed json encoding for post api query params.

### Type of change

## Changelog entry
Minor updates to Manage Events flow

## Screen Recording
https://github.com/user-attachments/assets/c9e781d9-e944-4f83-a95a-cdcf70f1b5f8

### DIFF
--- a/includes/Admin/Settings_Screens/Whatsapp_Utility.php
+++ b/includes/Admin/Settings_Screens/Whatsapp_Utility.php
@@ -117,9 +117,12 @@ class Whatsapp_Utility extends Abstract_Settings_Screen {
 				),
 			)
 		);
-		$order_placed_event_config_id   = get_option( 'wc_facebook_wa_order_placed_event_config_id', '' );
-		$order_shipped_event_config_id  = get_option( 'wc_facebook_wa_order_shipped_event_config_id', '' );
-		$order_refunded_event_config_id = get_option( 'wc_facebook_wa_order_refunded_event_config_id', '' );
+		$order_placed_event_config_id   = get_option( 'wc_facebook_wa_order_placed_event_config_id', null );
+		$order_placed_language          = get_option( 'wc_facebook_wa_order_placed_language', 'en' );
+		$order_shipped_event_config_id  = get_option( 'wc_facebook_wa_order_shipped_event_config_id', null );
+		$order_shipped_language         = get_option( 'wc_facebook_wa_order_shipped_language', 'en' );
+		$order_refunded_event_config_id = get_option( 'wc_facebook_wa_order_refunded_event_config_id', null );
+		$order_refunded_language        = get_option( 'wc_facebook_wa_order_refunded_language', 'en' );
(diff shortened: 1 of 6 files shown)


includes/Admin/Settings_Screens/Whatsapp_Utility.php
```diff
@@ -117,9 +117,12 @@ public function enqueue_assets() {
 				),
 			)
 		);
-		$order_placed_event_config_id   = get_option( 'wc_facebook_wa_order_placed_event_config_id', '' );
-		$order_shipped_event_config_id  = get_option( 'wc_facebook_wa_order_shipped_event_config_id', '' );
-		$order_refunded_event_config_id = get_option( 'wc_facebook_wa_order_refunded_event_config_id', '' );
+		$order_placed_event_config_id   = get_option( 'wc_facebook_wa_order_placed_event_config_id', null );
+		$order_placed_language          = get_option( 'wc_facebook_wa_order_placed_language', 'en' );
+		$order_shipped_event_config_id  = get_option( 'wc_facebook_wa_order_shipped_event_config_id', null );
+		$order_shipped_language         = get_option( 'wc_facebook_wa_order_shipped_language', 'en' );
```
sharunaanandraj, line 123: Can we delete these settings in wc_facebook_disconnect_whatsapp as well ? We can just add it to the array here https://github.com/facebook/facebook-for-woocommerce/pull/3128/files#r2071803856

includes/Admin/Settings_Screens/Whatsapp_Utility.php
```diff
@@ -511,8 +517,10 @@ public function render_manage_events_view() {
 			<div class="manage-event-card-item">
 				<p><b><?php esc_html_e( 'Select a language', 'facebook-for-woocommerce' ); ?></b></p>
 				<select id="manage-event-language" class="manage-event-selector">
+					<option value="en">English</option>
```
sharunaanandraj, line 520: How are we planning to handle all languages ? Do we have an api to fetch languages for a library template ?

sharunaanandraj (approved):
lgtm!

iodic (approved):

includes/Admin/Settings_Screens/Whatsapp_Utility.php
woo-ardsouza: Synced offline, we will add supported languages as a field in Get <integration_config_id> api

includes/Admin/Settings_Screens/Whatsapp_Utility.php
woo-ardsouza: good catch!

PR #3227, opened 2025-05-22
Change the logic to use transient to avoid state to be lost

francorisso:
## Description

Change the logic to use transient to avoid state to be lost, previously this was loosing the state after adding the logic to avoid calling this multiple times.

### DIFF
--- a/includes/RolloutSwitches.php
+++ b/includes/RolloutSwitches.php
@@ -32,12 +32,6 @@ class RolloutSwitches {
 		self::WHATSAPP_UTILITY_MESSAGING,
 		self::SWITCH_PRODUCT_SETS_SYNC_ENABLED,
 	);
-	/**
-	 * Stores the rollout switches and their enabled/disabled states.
-	 *
-	 * @var array
-	 */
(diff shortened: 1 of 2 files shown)


includes/RolloutSwitches.php
```diff
@@ -103,8 +97,8 @@ public function is_switch_enabled( string $switch_name ) {
 		if ( ! $this->is_switch_active( $switch_name ) ) {
 			return false;
 		}
-
-		return isset( $this->rollout_switches[ $switch_name ] ) ? $this->rollout_switches[ $switch_name ] : true;
+		
+		return get_option( self::SETTINGS_KEY )[$switch_name] === 'yes' ? true : false;
```
vinkmeta, line 101: What if get_option( self::SETTINGS_KEY ) is empty or null ?

vinkmeta: Add a default value to get_option and check if default value is returned then return false else return the value[$switch_name]

vinkmeta (changes requested):

vinkmeta (approved):

PR #3454, opened 2025-06-24
Set stable tag manual workflow

tzahgr:
## Description

Add a workflow to set the readme.txt "Stable tag" to the version based on package.json version.
The workflow:
- Runs on "Deployment" environment 
- Validate the branch matches release/*/publish
- Validate git tag for the new version exist
- Validate svn tag for the new version exist
- Updates "Stable tag" on readme.txt based on package.json version
- Commits to SVN

### Type of change

## Changelog entry

* Add - workflow to set the readme.txt "Stable tag" 

## Test Plan

See successful workflow run:
https://github.com/facebook/facebook-for-woocommerce/actions/runs/15858283545

### DIFF
--- a/.github/workflows/set-stable-tag.yml
+++ b/.github/workflows/set-stable-tag.yml
@@ -1,0 +1,58 @@
+name: Set Stable Tag
+
+on: workflow_dispatch
+
+jobs:
+  set-stable-tag:
+    runs-on: ubuntu-latest
+    environment: Deployment
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Fail if not on a release publish branch
+        run: |
+          echo "Current ref: ${{ github.ref }}"
+          BRANCH="${{ github.ref }}"
+          if [[ ! $BRANCH == release/*/publish ]]; then
+            echo "❌ This workflow can only be run on release/*/publish branches."
+            exit 1
+          fi
+
+      - name: Get NPM Version
+        id: package-version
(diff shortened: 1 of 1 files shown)


.github/workflows/set-stable-tag.yml
```diff
@@ -0,0 +1,58 @@
...
+          cd woocommerce_svn
+          sed -i -E "s/^(Stable tag:)[[:space:]]*[0-9.]+/\1 $NEW_VERSION/" "trunk/readme.txt"
+      
+      - name: Commit to SVN
+        run: |
+          cd woocommerce_svn
+          cat trunk/readme.txt
+          svn status
+          svn diff
+          svn commit -m "Update stable tag for version ${{ steps.package-version.outputs.current-version}}" --username "$SVN_USERNAME" --password "$SVN_PASSWORD" --no-auth-cache --non-interactive
```
vahidkay-meta, line 58: We'd want to commit that to git as well, right?

tzahgr: @vahidkay-meta not necessarily.
For Stable tag I think we should have one source of truth on the SVN.
See:
https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#how-the-readme-is-parsed

Only the trunk version of Stable tag is used to figure out the SVN tag used in production, so I would even consider setting the Git version of it as 0.0.0 or X.X.X to make it clearer.

Regardless, even if we decide to sync it, we can add this ability later to the workflow.

PR #3522, opened 2025-07-21
Deprecate FB Product Sets tab and migrate legacy product sets

mshymon:
## Description

This PR is a follow up project after we shipped automatic WooCommerce product category to Meta catlog product sets syncing. Ther are two main changes here:

1/ Remove all code regarding the Product Set tab, so users do not see it any more and there is no backend that supports it.
2/ Migrate previously manually created product sets by updating the sets filter from static retailer IDs to a dynamic filter. This is needed because otherwise legacy created product sets will get out of date quickly.
3/ Show a banner to users explaining what happened to Product Sets tab. (TODO)

### Type of change

## Changelog entry

Deprecate FB Product Sets tab and migrate legacy product sets.

## Test Plan

Upgrade your pluign verison to this one and after the upgrade the Product Sets tab will be gone and if you navigate to Comerce Manager you should see previosly created product sets to exist but with update filter (see screenshots).

### Before

Tab "Product Sets" is present
Screenshot 2025-07-21 at 15 05 10

Product Set filter
Screenshot 2025-07-21 at 15 20 35

### After

Tab "Product Sets" is gone
Screenshot 2025-07-21 at 16 52 53

Product Set filter (updated)
Screenshot 2025-07-21 at 15 35 23


mshymon:
Clsoing this one in favor of https://github.com/facebook/facebook-for-woocommerce/pull/3534 (that one has auto-gen zip file)

PR #3545, opened 2025-08-08
changelog: add referrer url

chc421:
## Description

This PR focus on adding referrer url to the event object.
gk wc_parambuilder_enable
This change enablement depended on [D80536437](https://www.internalfb.com/diff/D80536437)

### Type of change

## Checklist

## Changelog entry

Add referrer url

## Test Plan

Tested manually

Verify with log
pxl 1064514242455849
ubd 1187694782652822


facebook-commerce-events-tracker.php
```diff
@@ -1019,6 +1019,11 @@ public function inject_lead_event() {
 		 * @param bool  $send_now optional, defaults to true
 		 */
 		protected function send_api_event( Event $event, bool $send_now = true ) {
+			if ( isset( $_SERVER['HTTP_REFERER'] ) ) {
```
jarretth: Let's not do this by duplicating the event data. Where do we construct the original event? Let's add this code there. 
Ideally we want to minimize how deep we're looking at superglobals like $_SERVER as well. Lets keep the side effects contained and rely on a stored referrer where we create the event

vahidkay-meta: Should this be done only for Server events? 
If not, you can move it to includes>Events>Event.php file. That way it'll be added to all events.

vahidkay-meta (changes requested):

facebook-commerce-events-tracker.php
chc421: @jarretth I agree with @vahidkay-meta . Moving to Event.php would be much more cleaner.

includes/Events/Event.php
```diff
@@ -95,6 +95,7 @@ protected function prepare_data( $data ) {
 				'user_data'        => array(),
 			)
 		);
+		$this->data['referrer_url'] = isset( $_SERVER['HTTP_REFERER'] ) ? wc_clean( wp_unslash( $_SERVER['HTTP_REFERER'] ) ) : '';
```
vahidkay-meta: Please wrap it in an if, like this:
```
if ( isset( $_SERVER['HTTP_REFERER'] ) {
$this->data['referrer_url'] = wc_clean( wp_unslash( $_SERVER['HTTP_REFERER'] ) );
}
```

includes/Events/Event.php
```diff
@@ -96,6 +96,11 @@ protected function prepare_data( $data ) {
 			)
 		);
 
+		$this->data['referrer_url'] = '';
```
vahidkay-meta: You don't need this line

vahidkay-meta (approved):

PR #3549, opened 2025-08-11
changelog: add parambuilde

chc421:
## Description

Add parambuilder [library](https://github.com/facebook/capi-param-builder/tree/main/php) to improve fbc, fbp coverage. 

### Type of change

## Checklist

## Changelog entry

Add parambuilder to improve fbc, fbp

## Test Plan

manual test

## Screenshots

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -51,6 +51,9 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		/** @var bool whether the pixel should be enabled */
 		private $is_pixel_enabled;
 
+		/** @var \FacebookAds\ParamBuilder|null parameter builder instance */
+		private $param_builder = null;
+
 
 		/**
 		 * Events tracker constructor.
@@ -69,6 +72,42 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 			$this->tracked_events = array();
 
 			$this->add_hooks();
+			$this->initialize_param_builder();
+		}
+
+		/**
+		 * Initialize the Facebook CAPI Parameter Builder
+		 *
+		 * @since 3.0.0
+		 */
+		private function initialize_param_builder() {
+			try {
+				// Initialize the Facebook CAPI Parameter Builder with the site domain
+				$site_url = $_SERVER['HTTP_HOST'];
+				$this->param_builder = new \FacebookAds\ParamBuilder( array( $site_url ) );
+
+				// Process the request to get any cookies that need to be set
+				$cookie_to_set = $this->param_builder->processRequest(
+					$_SERVER['HTTP_HOST'],
+					$_GET,
+					$_COOKIE,
+					$_SERVER['HTTP_REFERER'] ?? null
+				);
+
+				// If there are cookies to set, set them
+				if ( ! empty( $cookie_to_set ) ) {
+					foreach ($cookie_to_set as $cookie) {
+						$res = setcookie(
+							$cookie->name,
+							$cookie->value,
+							time() + $cookie->max_age,
+							'/',
+							$cookie->domain);
+					}
+				}
+			} catch ( \Exception $exception ) {
+				facebook_for_woocommerce()->log( 'Error initializing CAPI Parameter Builder: ' . $exception->getMessage() );
+			}
 		}
 
 
@@ -1023,6 +1062,17 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 
 			if ( $send_now ) {
 				try {
+					// Get fbc and fbp values from the parameter builder
+					$fbc = $this->param_builder->getFbc();
+					$fbp = $this->param_builder->getFbp();
+
+					// $event_data = $event->get_data();
+					if (!empty($fbc)) {
+						$event['user_data']['fbc'] = $fbc;
+					}
+					if (!empty($fbp)) {
+						$event['user_data']['fbp'] = $fbp;
+					}
 					facebook_for_woocommerce()->get_api()->send_pixel_events( facebook_for_woocommerce()->get_integration()->get_facebook_pixel_id(), array( $event ) );
 				} catch ( ApiException $exception ) {
 					facebook_for_woocommerce()->log( 'Could not send Pixel event: ' . $exception->getMessage() );
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
     "woocommerce/action-scheduler-job-framework": "2.0.0",
     "composer/installers": "~1.0",
     "woocommerce/grow": "dev-compat-checker",
-    "firebase/php-jwt": "^6.10"
+    "firebase/php-jwt": "^6.10",
+    "facebook/capi-param-builder-php": "^1.0"
   },
   "require-dev": {
     "composer/installers": "^1.7.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "4e80c194ac3fde3776fb76150346aee6",
+    "content-hash": "0aae699fc138c58de959d423e499f785",
     "packages": [
         {
             "name": "composer/installers",
@@ -158,6 +158,40 @@
             "time": "2021-09-13T08:19:44+00:00"
         },
         {
+            "name": "facebook/capi-param-builder-php",
+            "version": "1.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/facebookincubator/capi-param-builder.git",
+                "reference": "0b92b4ebf3287a1ef9e5384532cd9d981b53435f"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/facebookincubator/capi-param-builder/zipball/0b92b4ebf3287a1ef9e5384532cd9d981b53435f",
+                "reference": "0b92b4ebf3287a1ef9e5384532cd9d981b53435f",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=7.4"
+            },
+            "require-dev": {
+                "phpunit/phpunit": "^8.5"
+            },
+            "type": "library",
+            "autoload": {
+                "psr-4": {
+                    "FacebookAds\\": "php/capi-param-builder/src/"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "description": "Parameter builder SDK for Conversion API events",
+            "support": {
+                "issues": "https://github.com/facebookincubator/capi-param-builder/issues",
+                "source": "https://github.com/facebookincubator/capi-param-builder/tree/v1.0.0"
+            },
+            "time": "2025-05-09T00:30:02+00:00"
+        },
+        {
             "name": "firebase/php-jwt",
             "version": "v6.10.0",
             "source": {
@@ -267,12 +301,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/woocommerce/grow.git",
-                "reference": "92a72d51e81b2ec62eb74d3e950a1e55cac63469"
+                "reference": "3bc225916439d7b93d4f5ecb209c94fb47e76172"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/woocommerce/grow/zipball/92a72d51e81b2ec62eb74d3e950a1e55cac63469",
-                "reference": "92a72d51e81b2ec62eb74d3e950a1e55cac63469",
+                "url": "https://api.github.com/repos/woocommerce/grow/zipball/3bc225916439d7b93d4f5ecb209c94fb47e76172",
+                "reference": "3bc225916439d7b93d4f5ecb209c94fb47e76172",
                 "shasum": ""
             },
             "require-dev": {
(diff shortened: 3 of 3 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -69,6 +72,43 @@ public function __construct( $user_info, $aam_settings ) {
...
+		 */
+		private function initialize_param_builder() {
+			try {
+				// Initialize the Facebook CAPI Parameter Builder with the site domain
+				$site_url = isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
+				$this->param_builder = new \FacebookAds\ParamBuilder( array( $site_url ) );
+
+				// Process the request to get any cookies that need to be set
+				$cookie_to_set = $this->param_builder->processRequest(
+					isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '',
```
jarretth:
```suggestion
				$site_url = isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
				$this->param_builder = new \FacebookAds\ParamBuilder( array( $site_url ) );

				// Process the request to get any cookies that need to be set
				$cookie_to_set = $this->param_builder->processRequest(
					$site_url,
```

facebook-commerce-events-tracker.php
```diff
@@ -69,6 +72,43 @@ public function __construct( $user_info, $aam_settings ) {
...
 			$this->add_hooks();
+			$this->initialize_param_builder();
+		}
+
+		/**
+		 * Initialize the Facebook CAPI Parameter Builder
+		 *
+		 * @since 3.0.0
+		 */
+		private function initialize_param_builder() {
```
jarretth: Lets separate out initialize_param_builder and the call to collect and set cookies

facebook-commerce-events-tracker.php
```diff
@@ -69,6 +72,43 @@ public function __construct( $user_info, $aam_settings ) {
...
+					isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '',
+					$_GET,
+					$_COOKIE,
+					isset( $_SERVER['HTTP_REFERER'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_REFERER'] ) ) : null
+				);
+
+				// Set cookie from parambuilder
+				if ( ! empty( $cookie_to_set ) ) {
+					foreach ( $cookie_to_set as $cookie ) {
+						setcookie(
```
vahidkay-meta: Have you tested this? Are you sure you can set a cookie here?

facebook-commerce-events-tracker.php
```diff
@@ -1023,6 +1092,16 @@ protected function send_api_event( Event $event, bool $send_now = true ) {
 
 			if ( $send_now ) {
 				try {
+					// Get fbc and fbp values from the parameter builder
+					$fbc = $this->param_builder->getFbc();
+					$fbp = $this->param_builder->getFbp();
```
vahidkay-meta: This shouldn't be done here. Please move it to includes>Events>Event.php

facebook-commerce-events-tracker.php
chc421: I tested this before adding the parambuilder client js. _fbp updated correctly. Is there a preferred place to set cookie?

vahidkay-meta: So on WordPress the cookies should be set before any headers are sent. 
That you can check by manually testing this.

chc421: I moved them to class-wc-facebookcommerce.php during plugin initialization.

class-wc-facebookcommerce.php
```diff
@@ -966,6 +985,79 @@ public function use_enhanced_onboarding(): bool {
...
+		$this->initialize_param_builder();
+		$this->set_param_builder_cookies();
+	}
+
+	/**
+	 * Initializes the Facebook CAPI Parameter Builder with the site domain.
+	 */
+	public function initialize_param_builder() {
+		try {
+			$site_url = isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
```
vahidkay-meta: You can use [get_site_url()](https://developer.wordpress.org/reference/functions/get_site_url/) method.

class-wc-facebookcommerce.php
```diff
@@ -966,6 +985,79 @@ public function use_enhanced_onboarding(): bool {
...
+				$site_url,
+				$_GET,
+				$_COOKIE,
+				isset( $_SERVER['HTTP_REFERER'] ) ? 
+					sanitize_text_field( wp_unslash( $_SERVER['HTTP_REFERER'] ) ) : 
+					null
+			);
+
+			if ( ! empty( $cookie_to_set ) ) {
+				foreach ( $cookie_to_set as $cookie ) {
```
vahidkay-meta: Have you tested this? Please take a look at the console when you test it to ensure the cookies are being set

facebook-commerce-events-tracker.php
```diff
@@ -177,6 +183,50 @@ public function inject_base_pixel_noscript() {
...
+			);
+		}
+
+		/**
+		 * Adds fbc and fbp to the user_data array if available from param_builder.
+		 *
+		 * @param array $user_data
+		 * @return array
+		 */
+		private function add_fbc_fbp_to_user_data( $user_data ) {
```
vahidkay-meta: Instead of doing it here, please make this change in includes>Events>Event.php
In there there's a method to retrieve fbp & fbc.

vahidkay-meta (changes requested):

class-wc-facebookcommerce.php
chc421: Here is how I tested it. I disabled parambuilder client js and I clear all cookies and reload the web.
I can see language token `AQ` was appended at the end. According parambuilder [doc](https://docs.google.com/document/d/1mZSy6xy-ql9JjznPB80GGS_lopHBsjzzTtGLx4inK2I/edit?tab=t.0#bookmark=id.3aqv47pgp9ws), this is a signature of cookie is being set by the php library.
Screenshot 2025-08-19 at 11 36 03 AM

facebook-commerce-events-tracker.php
```diff
@@ -176,6 +179,29 @@ public function inject_base_pixel_noscript() {
 			}
 		}
 
+		/**
+		 * Enqueues the Facebook CAPI Param Builder script.
+		 */
+		public function enqueue_capi_param_builder_script() {
+			if ( ! $this->is_pixel_enabled() ) {
```
chc421: @vahidkay-meta I wasn't sure if this is needed, I saw other inject methods had this.

chc421:
@vahidkay-meta we are planning to put these changes behind gk wc_parambuilder_enable, I think it might make more sense to merge referrer url first?

facebook-commerce-events-tracker.php
```diff
@@ -102,10 +109,14 @@ private function is_pixel_enabled() {
 		 */
 		private function add_hooks() {
 
+			add_action( 'init', array( $this, 'inject_param_builder' ) );
```
vahidkay-meta: Since both inject_param_builder and init_param_builder are added to init hook, they could be run in a random order. Is that expected?

facebook-commerce-events-tracker.php
```diff
@@ -149,6 +160,40 @@ private function add_hooks() {
...
+		/**
+		 * Sets the Parameter Builder cookies using the Facebook CAPI Parameter Builder.
+		 */
+		public function inject_param_builder() {
+			$param_builder = facebook_for_woocommerce()->get_param_builder();
+			if ( ! $param_builder ) {
+				return;
+			}
+
+			$site_url = isset( $_SERVER['HTTP_HOST'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
```
vahidkay-meta: let's use this instead: get_site_url()

facebook-commerce-events-tracker.php
vahidkay-meta: I'd change it with: facebook_for_woocommerce()->get_connection_handler()->is_connected()

includes/Events/Event.php
```diff
@@ -127,6 +135,17 @@ protected function prepare_user_data( $data ) {
 			$this->data['user_data']['country'] = $country;
 			unset( $this->data['user_data']['cn'] );
 		}
+		// Add parambuilder information to user data.
+		if ( null != $this->param_builder ) {
+			$fbc = $this->param_builder->getFbc();
```
vahidkay-meta: Instead of this, please update the methods for fbp & fbc: get_browser_id & get_click_id

vahidkay-meta (changes requested):

facebook-commerce-events-tracker.php
chc421: From what I tested, it didn't matter which one went first. Client side parambuilder takes time to load, it will always be server side parambuilder finish loading and then client side.

includes/Events/Event.php
```diff
@@ -284,6 +295,8 @@ protected function get_browser_id() {
 			$_SESSION['_fbp'] = $fbp;
 		} elseif ( isset( $_SESSION['_fbp'] ) ) {
 			$fbp = $_SESSION['_fbp']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		} elseif ( isset( $this->param_builder ) ) {
+			$fbp = $this->param_builder->getFBP();
 		}
```
chc421: @vahidkay-meta fbp can be retrieved from the parambuilder library and address initial empty payload issue.

facebook-commerce-events-tracker.php
vahidkay-meta: Why should these two be in different classes? Can't you first initialize it and then inject it in the same method? 
Why should there be 2 different methods/hooks for it?

includes/Events/Event.php
```diff
@@ -284,6 +295,8 @@ protected function get_browser_id() {
 			$_SESSION['_fbp'] = $fbp;
 		} elseif ( isset( $_SESSION['_fbp'] ) ) {
 			$fbp = $_SESSION['_fbp']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		} elseif ( isset( $this->param_builder ) ) {
+			$fbp = $this->param_builder->getFBP();
```
vahidkay-meta: It should be something like this: 
$fbp = ! empty( $_COOKIE['_fbp'] ) ? wc_clean( wp_unslash( $_COOKIE['_fbp'] ) ) : '';
if ( ! $fbp ) {
$fbp = $this->param_builder->getFBP();
} elseif ( isset( $_SESSION['_fbp'] ) ) {
			$fbp = $_SESSION['_fbp']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
}

if ($fbp) {
  $_SESSION['_fbp'] = $fbp;
}

vahidkay-meta (changes requested):

class-wc-facebookcommerce.php
```diff
@@ -179,6 +186,7 @@ public function __get( $key ) {
 	 * @internal
 	 */
 	public function init() {
+		add_action( 'init', array( $this, 'init_param_builder' ) );
```
iodic: Hi, can we add a priority here as well, so we can ensure proper execution? (No race conditions or issues with cookie setting.)

Something like: `add_action( 'init', array( $this, 'init_param_builder' ), '5' );`

facebook-commerce-events-tracker.php
```diff
@@ -101,11 +108,15 @@ private function is_pixel_enabled() {
 		 * @since 2.2.0
 		 */
 		private function add_hooks() {
+			add_action( 'init', array( $this, 'param_builder_server_setup' ) );
```
iodic: Can we add priority here as well? Like this: `add_action( 'init', array( $this, 'param_builder_server_setup' ). 15 );`.

facebook-commerce-events-tracker.php
```diff
@@ -176,6 +186,58 @@ public function inject_base_pixel_noscript() {
...
+			if ( ! $param_builder ) {
+				return;
+			}
+
+			try {
+				$cookie_to_set = $param_builder->getCookiesToSet();
+
+				if ( ! empty( $cookie_to_set ) ) {
+					foreach ( $cookie_to_set as $cookie ) {
+						setcookie(
```
iodic, line 203: Have you considered adding more params to this call?

Like this:
```php
setcookie(
	$cookie->name,
	$cookie->value,
	array(
		'expires'  => time() + $cookie->max_age,
		'path'     => '/',
		'domain'   => $cookie->domain,
		'secure'   => is_ssl(),
		'httponly' => true,
		'samesite' => 'Lax',
	)
);
```
Adding these params will result in better handling by more browsers.

chc421: Thank you for providing feedback. 
If that's necessary we should bring awareness to the parambuilder team.
We are following the instruction from the parambuilder lib [doc](https://github.com/facebook/capi-param-builder/tree/main/php#api-usage) and wasn't considering adding more params to this call.

chc421: @hongj-src do you think parambuilder will be benefit from these options?

facebook-commerce-events-tracker.php
```diff
@@ -101,11 +108,15 @@ private function is_pixel_enabled() {
 		 * @since 2.2.0
 		 */
 		private function add_hooks() {
+			add_action( 'init', array( $this, 'param_builder_server_setup' ), 5 );
```
iodic: The priorities got flipped: this one should be `15` and we should set `init_param_builder` to `5`. Can you update them?

iodic (changes requested):
Can you make the update to the action priorities?

facebook-commerce-events-tracker.php
```diff
@@ -176,6 +186,62 @@ public function inject_base_pixel_noscript() {
 			}
 		}
 
+		/**
+		 * Sets the Parameter Builder cookies using the Facebook CAPI Parameter Builder.
+		 */
+		public function param_builder_server_setup() {
```
iodic: One more note: can we rename this method to `param_builder_set_cookies`? That will make its name more closely aligned with what it does. :)

class-wc-facebookcommerce.php
```diff
@@ -837,6 +856,26 @@ public function use_enhanced_onboarding(): bool {
...
 		return true;
 	}
+
+	/**
+	 * Initializes the Facebook CAPI Parameter Builder with the site domain.
+	 */
+	public function init_param_builder() {
+		try {
+			$site_url = get_site_url();
+			$this->param_builder = new \FacebookAds\ParamBuilder( array( $site_url ) );
```
hongj-src, line 886: cmiiw, seems get_site_url() seems return the full url. This field we'd recommend to return etld+1, so the recommended cookie would benefit the most. Not sure if this also could apply to woocommerce website. 

Example like: my website is www.example.com. But current webpage is on subdomain.example.com or homepage.subdomain.example.com. The constructor for ParamBuilder recommended to input etld+1(in this example is example.com). As a result, even I'm in a subdomain(eg. test.example.com), my cookie will persist.

chc421: I saw other approaches here
https://github.com/facebook/capi-param-builder/blob/07e36aad065669fb4b73cb94165cfaed6416e20b/php/examples/local/FacebookParamManager.php#L21-L27
which one is preferred?

facebook-commerce-events-tracker.php
hongj-src: That's great question! In demo, se didn't intentionally set secure, httponly and samesite. The goal for this API is set cookies and serve the cookie for pixel and CAPI use cases. 
I have very limited knowledge on woocommerce, so I just roughly explain why we added them in demo. 
For some use cases, the fbc/fbp may be updated via Javascript(eg. advertiser website install both parambuilder + meta pixel). If this case apply, the possible preferred httponly is false.
samesite for lax is good for our use case, I'll update in demo too(mark as optional). secure is totally optional, more up to the website settings.

PR #3617, opened 2025-09-19
Pb fix option 3

chc421:
## Description

Overview:
The CAPI Parambuilder library works with both client-side and server-side integration and can potentially improve fbc and fbp coverage and matchability.

In this PR, we are adding
1. ParamBuilder client library (facebook-commerce-events-tracker.php)
2. ParamBuilder server library (facebook-commerce-events-tracker.php)
3. Enhance fbp missing for CAPI in initial events (Event.php)

Design doc: https://docs.google.com/document/d/1GH0dtJWjUrJlG8BUSzfyeoNlC9eK9-IdJBn3mp4shhM/edit?tab=t.0

### Type of change

## Checklist

- [X] I have commented my code, particularly in hard-to-understand areas, if any.
- [X] I have confirmed that my changes do not introduce any new PHPCS warnings or errors. 
- [X] I have checked plugin debug logs that my changes do not introduce any new PHP warnings or FATAL errors.
- [X] I followed general Pull Request best practices. Meta employees to follow this [wiki]([url](https://fburl.com/wiki/2cgfduwc)).
- [X] I have added tests (if necessary) and all the new and existing unit tests pass locally with my changes.
- [X] I have completed dogfooding and QA testing, or I have conducted thorough due diligence to ensure that it does not break existing functionality.
- [X] I have updated or requested update to plugin documentations (if necessary). Meta employees to follow this [wiki]([url](https://fburl.com/wiki/nhx73tgs)).

## Changelog entry

Add client and server parambuilder.

## Test Plan

Manual test with following cases:
1. **Incognito mode: new visitor (no fbclick)**
Screenshot 2025-09-30 at 2 47 57 PM

Screenshot 2025-09-30 at 3 28 48 PM

Screenshot 2025-09-30 at 3 31 00 PM

2. **Incognito mode: new visitor (with fbclick)**
Screenshot 2025-09-30 at 2 57 50 PM

Screenshot 2025-09-30 at 2 58 11 PM

3.  **Admin page**
Screenshot 2025-09-30 at 2 59 03 PM

## Screenshots
### Before
Screenshot 2025-09-30 at 3 37 06 PM

### After
Screenshot 2025-09-30 at 3 34 43 PM

D83659301

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -207,9 +207,6 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 			add_action( 'woocommerce_process_shop_order_meta', array( $this, 'inject_purchase_event' ), 20 );
 			add_action( 'woocommerce_checkout_update_order_meta', array( $this, 'inject_purchase_event' ), 30 );
 			add_action( 'woocommerce_thankyou', array( $this, 'inject_purchase_event' ), 40 );
-			add_action( 'woocommerce_payment_complete', array( $this, 'inject_purchase_event' ), 50 );
-			add_action( 'woocommerce_order_status_processing', array( $this, 'inject_purchase_event' ), 60 );
-			add_action( 'woocommerce_order_status_completed', array( $this, 'inject_purchase_event' ), 70 );
 
 			// Lead events through Contact Form 7
 			add_action( 'wpcf7_contact_form', array( $this, 'inject_lead_event_hook' ), 11 );
@@ -522,7 +519,13 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		 */
 		public function send_search_event() {
 
-			$this->send_api_event( $this->get_search_event() );
+			$event = $this->get_search_event();
+
+			if ( null === $event ) {
+				return;
+			}
+
+			$this->send_api_event( $event );
 		}
 
 
@@ -545,6 +548,10 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 				$product_ids  = array();
 				$contents     = array();
 				$total_value  = 0.00;
+
+				if ( empty( $wp_query->posts ) ) {
+					return null;
+				}
 
 				foreach ( $wp_query->posts as $post ) {
 
@@ -596,6 +603,10 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 		public function actually_inject_search_event() {
 
 			$event = $this->get_search_event();
+
+			if ( null === $event ) {
+				return;
+			}
 
 			$this->send_api_event( $event );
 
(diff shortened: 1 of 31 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -68,9 +73,50 @@ public function __construct( $user_info, $aam_settings ) {
...
+
+			return self::$param_builder;
+		}
+
+		/**
+		 * Initializes the CAPI server side Parameter Builder and sets cookies as needed.
+		 */
+		public function param_builder_server_setup() {
+			try {
+				$cookie_to_set = $this->get_param_builder()->getCookiesToSet();
```
vahidkay-meta: get_param_builder is a static method. It should be called with self:: and not $this.

facebook-commerce-events-tracker.php
```diff
@@ -176,6 +224,30 @@ public function inject_base_pixel_noscript() {
...
+		 */
+		public function param_builder_client_setup() {
+			// Client js setup
+			if ( ! facebook_for_woocommerce()->get_connection_handler()->is_connected() ) {
+				return;
+			}
+
+			wp_enqueue_script(
+				'facebook-capi-param-builder',
+				'https://capi-automation.s3.us-east-2.amazonaws.com/public/client_js/capiParamBuilder/clientParamBuilder.bundle.js',
```
vahidkay-meta, line 256: Is this something that's gonna live forever ( long-term )? 
Some users tend to dwell on a version for a long time. So if there's a chance that in future the team decides to shut this s3 down and replace it with another URL, that needs to be addressed now. 
Either we need commitment from that team that they will inform us at least 6 months in advance, or we need to handle this ourselves

includes/Events/Event.php
```diff
@@ -284,11 +284,15 @@ protected function get_click_id() {
 	 */
 	protected function get_browser_id() {
```
vahidkay-meta, line 288: Can you do the same for fbc? 
Was that skipped intentionally?

facebook-commerce-events-tracker.php
chc421: @hongj-src iirc this url should be a long term url?

includes/Events/Event.php
chc421: Thanks for pointing this. When I was testing, I found that it was already handled by https://github.com/facebook/facebook-for-woocommerce/blob/main/includes/Events/Event.php#L264. I don't think parambuilder would provide more than what is already there, but I can add the logic if there are cases that parambuilder can handle but the existing implementation cannot.

vahidkay-meta: Yes, let's add that since it's going to help improve the backend coverage

facebook-commerce-events-tracker.php
hongj-src: This S3 link will be used for at least 6 months + usage. We are considering move to a CDN, but the task got delayed. Even our CDN is ready, we'll maintain both for certain period. We should be good use the s3 link

facebook-commerce-events-tracker.php
```diff
@@ -68,9 +73,50 @@ public function __construct( $user_info, $aam_settings ) {
 			$this->aam_settings   = $aam_settings;
 			$this->tracked_events = array();
 
+			$this->param_builder_server_setup();
```
vahidkay-meta, line 76: $this is not correct.
It should be self

chc421: param_builder_server_setup() is not a static method, do you want me to change its signature?

vahidkay-meta: You're right. My bad

Issue #3653, opened 2025-10-03
Avoid loading checkout inside full-page iframe or allow proper URL navigation


jitman:
The plugin currently handles the /fb-checkout route by rendering the WooCommerce checkout page inside a full-screen <iframe> rather than performing a standard redirect. This approach causes several UX, technical, and compatibility issues.

https://github.com/facebook/facebook-for-woocommerce/blob/19c0808a1a7c994404a5f0970d67515c375eacbf/includes/Checkout.php#L216

## Steps to Reproduce
1. Install and activate the Facebook for WooCommerce plugin, set up with Meta Commerce account
2. Build a test checkout URL formatted as: /fb-checkout?products={...}&coupon={...}.
3. Upon code inspection, observe that the page loads an HTML wrapper with an <iframe> pointing to the checkout URL.

## Current Behavior

- The entire checkout experience is embedded inside an <iframe>.
- The browser’s address bar remains fixed at /fb-checkout, even as users progress through the checkout flow.
- Navigating elsewhere or refreshing causes confusion because the URL does not reflect the current page state.

## Expected Behavior

- Users should be redirected to the normal WooCommerce checkout URL after the cart is rebuilt, not shown within an <iframe>.
- Alternatively, if the iframe is required for specific integrations, the plugin should:
- Allow URL updates (pushState or redirect) when navigating to checkout or order confirmation.
- Provide a filter or setting to disable iframe wrapping.

## Issues with Current Implementation

1. Session and Cookie Problems - Some browsers restrict cookies and local storage in iframe contexts, leading to lost sessions or failed payments.
2. Payment Gateway Incompatibility - Payment methods could fail or are blocked in iframes.
3. Analytics and Conversion Tracking - Conversion tracking scripts may not fire correctly in iframes.
4. Poor UX - URL never changes, confusing users and breaking bookmarks, refreshes, and the back button.
5. Accessibility and SEO - Screen readers and accessibility tools struggle with iframe-wrapped pages.

## Proposed Solutions

1. Preferred: Replace the iframe output with a direct redirect to `wc_get_checkout_url()` after rebuilding the cart and applying coupons.
2. Alternative: Introduce a filter or option to disable iframe embedding while keeping the cart-rebuilding logic intact.
3. Minimal: Update browser history (via window.history.pushState) to reflect the actual checkout URL inside the iframe.

## Environment

- WordPress: 6.8.3
- WooCommerce: 10.2.2
[...]

PR #3656, opened 2025-10-06
Make iframe in /fb-checkout optional and default to direct redirect

ukilla:
## Description

This change makes the /fb-checkout iframe optional and defaults to using a top-level redirect to the standard WooCommerce checkout page.
Previously, the plugin always rendered the checkout inside a full-screen iframe, causing UX, session, payment gateway, and analytics issues. With this update:
- Classic redirect is now the default behavior for /fb-checkout.
- The iframe remains available as a legacy option via the facebook_commerce_checkout_use_iframe filter.
- Cart rebuilding and coupon application via query parameters remain unchanged.
- A JavaScript fallback ensures proper navigation if headers are already sent.
No additional dependencies are required. This update improves browser history handling, UX consistency, payment gateway compatibility, and analytics tracking.
Fixes/addresses [Issue #3653 ](https://github.com/facebook/facebook-for-woocommerce/issues/3653)

### Type of change

## Checklist

- [X] I have commented my code, particularly in hard-to-understand areas, if any.
- [X] I have confirmed that my changes do not introduce any new PHPCS warnings or errors. 
- [X] I have checked plugin debug logs that my changes do not introduce any new PHP warnings or FATAL errors.
- [X] I followed general Pull Request best practices. Meta employees to follow this [wiki]([url](https://fburl.com/wiki/2cgfduwc)).
- [X] I have added tests (if necessary) and all the new and existing unit tests pass locally with my changes.
- [X] I have completed dogfooding and QA testing, or I have conducted thorough due diligence to ensure that it does not break existing functionality.
- [] I have updated or requested update to plugin documentations (if necessary). Meta employees to follow this [wiki]([url](https://fburl.com/wiki/nhx73tgs)).

## Changelog entry

/fb-checkout now defaults to a classic redirect; iframe support is optional via filter.

## Test Plan

Tested the changes using the setup checkout URL provided by Facebook Commerce Manager. Verified that the cart rebuilds correctly, coupons are applied, and users are redirected to the standard WooCommerce checkout page.


ajello-meta:
The reason `/fb-checkout` was implemented as an iframe and not a standard redirect to the checkout page was to ensure that the URL and it's parameters persist and load an identical cart when being opened in another window. When buyers open the seller's checkout in Meta's in-app browser, many of them choose to "open in external browser", if the URL does not persist, then the cart is emptied (as with the standard redirect).

PR #3713, opened 2025-11-18
Adding Language Override Feeds for Products

rafael-curran:
## Description

This adds language override support for catalogs, enabling sites to upload translations of products. Supported localization plugins in this release are WPML and Polylang. 

### Type of change

## Checklist

## Changelog entry

Adding Language Override Feeds for WPML and Polylang Product Translations

### DIFF
--- a/includes/Integrations/WPML.php
+++ b/includes/Integrations/WPML.php
@@ -191,6 +191,47 @@ class WPML extends Abstract_Localization_Integration {
 	}
 
 	/**
+	 * Get the plugin-specific language identifier for a given locale
+	 *
+	 * Converts a full locale (e.g., 'es_ES') to WPML's language code.
+	 *
+	 * @param string $locale Full locale code (e.g., 'es_ES', 'fr_FR')
+	 * @return string|null WPML language code or null if not found
+	 */
+	protected function get_plugin_language_identifier( string $locale ): ?string {
+		$wpml_languages = apply_filters( 'wpml_active_languages', null );
+
+		if ( is_array( $wpml_languages ) ) {
+			foreach ( $wpml_languages as $code => $language_data ) {
+				$language_locale = $language_data['default_locale'] ?? $code;
+				if ( $language_locale === $locale ) {
+					return $code;
+				}
+			}
+		}
+
+		// Fallback: use the locale as the code
+		return $locale;
+	}
+
+	/**
+	 * Check if a product is in a specific language
+	 *
+	 * @param int    $product_id Product ID to check
+	 * @param string $language_identifier WPML language code
+	 * @return bool True if product is in the specified language
+	 */
+	protected function is_product_in_language( int $product_id, string $language_identifier ): bool {
+		$product_language = apply_filters( 'wpml_post_language_details', null, $product_id );
+
+		if ( $product_language && isset( $product_language['language_code'] ) ) {
+			return $product_language['language_code'] === $language_identifier;
+		}
+
+		return false;
+	}
+
+	/**
 	 * Get products from the default language
 	 *
 	 * Uses WPML's API to find products that are in the default language.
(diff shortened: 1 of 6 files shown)


includes/fbwpml.php
```diff
@@ -141,6 +141,24 @@ public function wpml_support() {
...
+						return FB_WPML_Language_Status::VISIBLE === $status;
+					}
+				)
+			);
+
+			// Check if user has legacy multi-language setup (2+ languages visible)
+			$has_legacy_setup = $visible_languages_count >= 2;
+
+			// Don't show the section at all for new/eligible users
+			if ( ! $has_legacy_setup ) {
```
vahidkay-meta, line 158: How do you know if it's an old user, or someone who installs wpml after this work added two languages?

includes/Jobs/GenerateProductFeed.php
```diff
@@ -91,14 +91,31 @@ protected function process_items( array $items, array $args ) {
...
+			// When localization plugin is active AND language override feeds are enabled,
+			// only retrieve default language products
+			$default_language = $integration->get_default_language();
+			if ( $default_language ) {
+				$query_args['lang'] = $default_language;
+			}
+		} else {
+			// When no localization plugin is active OR language override feeds are disabled,
+			// retrieve all products
+			$query_args['lang'] = 'all';
```
vahidkay-meta, line 115: Why are you adding this? 
In the current code, that query arg is not there. 
If we add the same query arg in the current code ( without your changes ), how would it behave differently?

includes/Integrations/WPML.php
```diff
@@ -0,0 +1,434 @@
+<?php
+declare( strict_types=1 );
```
vahidkay-meta, line 2: How will this affect the rest of the plugin code?

includes/Integrations/WPML.php
```diff
@@ -0,0 +1,434 @@
...
+			/** Use WPML filter to check if this product is in the default language */
+			$product_language = apply_filters( 'wpml_post_language_details', null, $product_id );
+
+			if ( $product_language && isset( $product_language['language_code'] ) ) {
+				// Only include products that are in the default language (compare WPML codes)
+				if ( $product_language['language_code'] === $default_language_code ) {
+					$default_language_products[] = $product_id;
+
+					// Stop when we have enough products
+					if ( count( $default_language_products ) >= $limit ) {
```
vahidkay-meta: What if you found 50 products in a secondary language? Low probability, but possible?

includes/Jobs/GenerateProductFeed.php
rafael-curran: If no lang is specified in the query parameters, Polylang adds the current language as a filter in [ln 178 here](https://github.com/polylang/polylang/blob/e650089b81a39c4ad7e5bedfd4344c54e917b2c9/include/query.php#L4). By adding lang = 'all', I'm forcing all products to be returned, and then we filter for only default lang products in ProductValidator. The current behavior of the plugin, when using Polylang, depends on the current language set in the admin panel, which can vary and thus impact the set of products which are synced to Facebook. This is a bug. Without the language feed changes, adding this parameter+validation alone essentially fixes that bug, by only syncing default language products.

includes/fbwpml.php
rafael-curran: I'm hiding these settings as part of this feature for users with a valid setup (<2 langs selected), so they won't be able to do this.

includes/Integrations/WPML.php
rafael-curran: This only enforces types on methods used in the current file.

includes/Integrations/WPML.php
rafael-curran: Updated to remove the hardcoded limits to be safe, matching the main feed generation in fbproductfeed.php

PR #3835, opened 2026-01-23
Fix pixel event tracking by isolating JS execution context

cshing-meta:
## Description

Pixel events were failing when other plugins caused JS errors because wc_enqueue_js() collects all inline JS into a shared execution context. If any script in that queue throws an error, subsequent pixel events never fire, causing a gap between browser pixel events and server-side events.

This PR replaces wc_enqueue_js() with wp_enqueue_script() + wp_localize_script() to fire events via an external JS file in an isolated execution context. 

Refer to gap 1 in technical plan: https://docs.google.com/document/d/1Yry02lgbnS_Uaj3ggQ-yruJss2J4FTwOMd7QvpHGzLU/edit?tab=t.0 

gap-1-logic-flow

### Type of change

## Checklist

## Changelog entry

One liner entry to be surfaced in changelog.txt

## Test Plan
1. Run `./vendor/bin/phpunit tests/Unit/FacebookCommercePixelIsolatedExecutionTest.php` and all unit tests should pass.
2. Run `npm run test:js -- --testPathPattern="pixel-events"` and all jest tests should pass.
3. All E2E tests should pass.
4. Manual Testing for the following scenarios: 
Visit product page -> ViewContent + PageView events should fire
Click "Add To Cart" on product page with AJAX enabled only -> No events should fire (same state prior to change)
Click "Add To Cart" with AJAX and redirect enabled -> AddToCart + PageView event should fire
Inject JS error via mu-plugin and visit product page -> pixel events still fire despite error

### DIFF
--- a/includes/fbutils.php
+++ b/includes/fbutils.php
@@ -73,6 +73,11 @@ if ( ! class_exists( 'WC_Facebookcommerce_Utils' ) ) :
 				return;
 			}
 
+			// Check if isolated pixel execution is enabled
+			$is_isolated_execution_enabled = facebook_for_woocommerce()->get_rollout_switches()->is_switch_enabled(
+				\WooCommerce\Facebook\RolloutSwitches::SWITCH_ISOLATED_PIXEL_EXECUTION_ENABLED
+			);
+
 			// Separate events by type
 			$legacy_events   = array();
 			$isolated_events = array();
@@ -85,14 +90,16 @@ if ( ! class_exists( 'WC_Facebookcommerce_Utils' ) ) :
 				}
 			}
 
-			// Handle isolated execution events (event data arrays)
-			foreach ( $isolated_events as $event ) {
-				WC_Facebookcommerce_Pixel::enqueue_event(
-					$event['name'],
-					$event['params'],
-					$event['method'] ?? 'track',
-					$event['eventId'] ?? ''
-				);
+			// Handle isolated execution events (event data arrays) - only if switch is enabled
+			if ( $is_isolated_execution_enabled ) {
+				foreach ( $isolated_events as $event ) {
+					WC_Facebookcommerce_Pixel::enqueue_event(
+						$event['name'],
+						$event['params'],
+						$event['method'] ?? 'track',
+						$event['eventId'] ?? ''
+					);
+				}
 			}
 
 			// Handle legacy events (JS code strings) - combine into single script tag
--- a/tests/Unit/FacebookCommercePixelIsolatedExecutionTest.php
+++ b/tests/Unit/FacebookCommercePixelIsolatedExecutionTest.php
@@ -617,24 +617,18 @@ class FacebookCommercePixelIsolatedExecutionTest extends AbstractWPUnitTestWithO
 	 * Enable the isolated pixel execution rollout switch.
 	 */
 	private function enable_isolated_pixel_execution_switch(): void {
-		update_option(
-			'wc_facebook_for_woocommerce_rollout_switches',
-			array(
-				\WooCommerce\Facebook\RolloutSwitches::SWITCH_ISOLATED_PIXEL_EXECUTION_ENABLED => 'yes',
-			)
-		);
+		$options = get_option( 'wc_facebook_for_woocommerce_rollout_switches', array() );
+		$options[ \WooCommerce\Facebook\RolloutSwitches::SWITCH_ISOLATED_PIXEL_EXECUTION_ENABLED ] = 'yes';
+		update_option( 'wc_facebook_for_woocommerce_rollout_switches', $options );
 	}
 
 	/**
 	 * Disable the isolated pixel execution rollout switch.
 	 */
 	private function disable_isolated_pixel_execution_switch(): void {
-		update_option(
-			'wc_facebook_for_woocommerce_rollout_switches',
-			array(
-				\WooCommerce\Facebook\RolloutSwitches::SWITCH_ISOLATED_PIXEL_EXECUTION_ENABLED => 'no',
-			)
-		);
+		$options = get_option( 'wc_facebook_for_woocommerce_rollout_switches', array() );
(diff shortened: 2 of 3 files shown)


assets/js/frontend/pixel-events.js
```diff
@@ -0,0 +1,138 @@
...
+    function fireStaticEvents() {
+        var events = data.staticEvents;
+
+        if (!events || !Array.isArray(events)) {
+            return;
+        }
+
+        for (var i = 0; i < events.length; i++) {
+            try {
+                fireEvent(events[i]);
```
vahidkay-meta, line 117: Shouldn't you remove the events that are fired from the list? 
In case they don't have an event id, they'd be fired again and again,  no?

vahidkay-meta: Unless you're ensuring that this fireStaticEvents or fireEvent only happens once per each event

facebook-commerce-pixel-event.php
```diff
@@ -69,6 +93,164 @@ public function __construct( $user_info = [] ) {
...
+	/**
+	 * Adds an event to the static events array for immediate frontend execution.
+	 * Events are stored as DATA, not executable code.
+	 *
+	 * @param string $event_name The name of the event to track.
+	 * @param array  $params     Event parameters.
+	 * @param string $method     The fbq method to use (track, trackCustom, etc.).
+	 * @param string $event_id   Optional event ID for deduplication.
+	 */
+	public static function add_static_event( $event_name, $params, $method = 'track', $event_id = '' ) {
```
vahidkay-meta: Why do you call it "static event"?
Shouldn't this just be something like: enqueue_event?

facebook-commerce-pixel-event.php
```diff
@@ -69,6 +93,164 @@ public function __construct( $user_info = [] ) {
...
+	/**
+	 * Prepares event parameters for pixel tracking.
+	 *
+	 * Extracts event_id, unwraps custom_data, and applies build_params().
+	 *
+	 * @param array  $params     Raw event parameters.
+	 * @param string $event_name The name of the event.
+	 * @return array ['params' => array, 'event_id' => string]
+	 */
+	private static function prepare_event_params( $params, $event_name ) {
```
vahidkay-meta, line 222: This is mostly done when we create a new event. Check Event.php constructor
Let's not create additional logic here.

assets/js/frontend/pixel-events.js
cshing-meta: Good catch! fireStaticEvents() is only called once on page load via init() so in practice this shouldn't cause duplicates. But to be safe, we can add a safeguard to clear the events after firing.

facebook-commerce-pixel-event.php
cshing-meta: Good Point! enqueue_event has a clearer intent; I'll update it.

facebook-commerce-pixel-event.php
cshing-meta: This is not additional logic. The logic here already exists in build_events below; I simplify refactored it into a separate method to make it easier to understand. 
Also, the Event.php constructor is mainly for CAPI events, and the data structures are different between both CAPI and Pixel. E.g. CAPI requires server-side data (IP, user agent) whereas Pixel events don't.

assets/js/frontend/pixel-events.js
vahidkay-meta: Let's add that. In case events don't have an Id, they could be fired multiple times and we don't want that.

includes/fbutils.php
```diff
@@ -60,13 +60,44 @@ class WC_Facebookcommerce_Utils {
...
+		 * Supports both legacy (JS code string) and isolated execution (event data array) formats.
+		 * - Legacy format (switch OFF): Outputs inline <script> tag with JS code
+		 * - Isolated format (switch ON): Uses WC_Facebookcommerce_Pixel::enqueue_event()
+		 *
 		 * @since 3.1.6
 		 */
 		public static function print_deferred_events() {
 			$deferred_events = static::load_deferred_events();
 
-			if ( ! empty( $deferred_events ) ) {
```
vahidkay-meta, line 68: Your code should be gated by the rollout switch here.

facebook-commerce-pixel-event.php
```diff
@@ -292,16 +479,44 @@ public function get_event_script( $event_name, $params, $method = 'track' ) {
...
+				} else {
+					// Queue event for this page's external script.
+					self::enqueue_event( $event_name, $event_params, $method, $event_id );
+				}
 			} else {
-				WC_Facebookcommerce_Utils::enqueue_inline_js( $code );
+				// Legacy execution: Use enqueue_inline_js for inline script.
+				$code = $this->get_event_code( $event_name, self::build_params( $params, $event_name ), $method );
+
+				if ( $is_deferred ) {
```
vahidkay-meta, line 514: You're changing the logic here. previously, is_add_to_cart also was being checked

vahidkay-meta (changes requested):

assets/js/frontend/pixel-events.js
cshing-meta: It was already added on line 124.

cshing-meta: It was already added on line 124

includes/fbutils.php
cshing-meta: Isolated execution uses arrays and the legacy/existing approach uses strings for events, so it's guaranteed they won't conflict with each other. I'll add a rollout switch gate here as extra precaution

facebook-commerce-pixel-event.php
cshing-meta: The logic is still the same. `is_add_to_cart` is still here, it's just encapsulated in the `$is_deferred` variable on line 495 - `$is_deferred = $is_redirect && $is_add_to_cart;`. This was refactored so that we could reuse the same condition in both the isolated + legacy code paths.

PR #3917, opened 2026-05-05
E2E test expansion

iodic:
- Increase Cross-browser and Mobile Coverage
- Test Events with Variable and Grouped Products
- Privacy Sandbox: Remove Filtering and Add Tests
- Document and Clarify Event Field Contracts
- Remove WP User Login Requirement
- Add Theme Variations in Tests
- Cover Ajax AddToCart and Applicable Events
- Introduces new E2E coverage for signal consent hold/release behavior and replaces the previous consent test implementation


.github/workflows/product-creation-tests.yml
```diff
@@ -600,15 +603,125 @@ jobs:
...
+          sudo apt-get install -y opera-stable
+
+          OPERA_BIN="$(command -v opera-stable || command -v opera || true)"
+          if [ -z "$OPERA_BIN" ]; then
+            echo "❌ Opera binary not found after install" >&2
+            exit 1
+          fi
+          echo "OPERA_EXECUTABLE_PATH=$OPERA_BIN" >> $GITHUB_ENV
+
+      - name: Verify Opera browser setup
```
vahidkay-meta, line 717: What about Safari?

vahidkay-meta:
**Safari/WebKit on mobile:** The `ios-iphone-wp-customer` project currently uses Chromium with iPhone 13 device emulation (viewport + user-agent), but not the WebKit engine. Playwright ships a built-in WebKit browser that closely matches Safari's rendering engine — worth adding a separate `safari-ios-wp-customer` project with `browserName: 'webkit'` to catch WebKit-specific quirks. The CI step would just need `npx playwright install webkit`.

.github/workflows/product-creation-tests.yml
ukilla: Safari is now covered. 
Added safari-ios-wp-customer using Playwright webkit + iPhone 13 profile, and updated CI to install WebKit, verify WebKit setup, and run that project in plugin-events.

PR #3939, opened 2026-05-28
Prevent duplicate AddToCart Pixel fires from repeated AJAX fragment execution

cshing-meta:
## Description

This fixes a duplicate AddToCart Pixel event issue on the add-to-cart flows. Some third-party plugins can cause AJAX fragments or related frontend handlers to execute the same injected Pixel script more than once. When that happened, the same AddToCart event could call fbq() twice in the browser.

### Type of change

Added browser-side deduplication using the event’s eventID. Every AddToCart event already has a unique ID. When the browser is about to send the Pixel event, we now check whether that exact ID has already been sent on the page.

If it has not been sent yet:

mark this eventID as sent
send the Pixel event

If it has already been sent:
do nothing
skip the duplicate Pixel event

This fixes the issue because the duplicate was happening in the browser. WooCommerce AJAX was causing the same AddToCart Pixel script to run twice. Both executions had the same eventID, so the new guard lets the first one fire and blocks the second one.

## Checklist

## Changelog entry

One liner entry to be surfaced in changelog.txt

## Test Plan

Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce. 
Please also list any relevant details for your test configuration.

1. Install a few third party plugins (e.g, Side Cart, FunnelCart, etc). Go to the Shop page and click Add To Cart. Verify that only 1 AddToCart pixel gets fired.

### Before

https://github.com/user-attachments/assets/280a5019-80ab-4ef6-a1f7-216640be0310

### After

https://github.com/user-attachments/assets/b7093525-70fd-40f3-93b7-30714f07d441

### DIFF
--- a/facebook-commerce-pixel-event.php
+++ b/facebook-commerce-pixel-event.php
@@ -958,9 +958,15 @@ JS;
 		if ( ! empty( $event_id ) ) {
 			$event = sprintf(
 				"/* %s Facebook Integration Event Tracking */\n" .
+				"window.wcFacebookPixelFiredEvents = window.wcFacebookPixelFiredEvents || {};\n" .
+				"if (!window.wcFacebookPixelFiredEvents[%s]) {\n" .
+				"window.wcFacebookPixelFiredEvents[%s] = true;\n" .
 				"fbq('set', 'agent', '%s', '%s');\n" .
-				"fbq('%s', '%s', %s, %s);",
+				"fbq('%s', '%s', %s, %s);\n" .
+				'}',
 				WC_Facebookcommerce_Utils::get_integration_name(),
+				wp_json_encode( $event_id ),
+				wp_json_encode( $event_id ),
 				Event::get_platform_identifier(),
 				self::get_pixel_id(),
 				esc_js( $method ),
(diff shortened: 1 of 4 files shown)


facebook-commerce-pixel-event.php
```diff
@@ -958,9 +958,15 @@ public static function build_event( $event_name, $params, $method = 'track' ) {
 		if ( ! empty( $event_id ) ) {
 			$event = sprintf(
 				"/* %s Facebook Integration Event Tracking */\n" .
+				"window.wcFacebookPixelFiredEvents = window.wcFacebookPixelFiredEvents || {};\n" .
+				"if (!window.wcFacebookPixelFiredEvents[%s]) {\n" .
+				"window.wcFacebookPixelFiredEvents[%s] = true;\n" .
 				"fbq('set', 'agent', '%s', '%s');\n" .
```
iodic: I think it would be better to move this line above the check (961). That way, we prevent the agent registration being skipped.

iodic (changes requested):

facebook-commerce-pixel-event.php
cshing-meta: @iodic Thanks for reviewing! Updated the order, could you review?

iodic (approved):
LGTM

facebook-commerce-pixel-event.php
iodic: Thank you, looking good!

PR #3951, opened 2026-06-17
Enhance cookie handling and inline script conditions

jaygidwitz:
adjust previous pull request which didn't work due to the cookie VALUE being different each request

ParamBuilder is generating a new _fbp on every request instead of reusing the existing one.

### DIFF
--- a/facebook-commerce-events-tracker.php
+++ b/facebook-commerce-events-tracker.php
@@ -193,8 +193,13 @@ if ( ! class_exists( 'WC_Facebookcommerce_EventsTracker' ) ) :
 				} else {
 					$cookie_to_set = $param_builder->getCookiesToSet();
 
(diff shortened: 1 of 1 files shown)


facebook-commerce-events-tracker.php
```diff
@@ -193,8 +193,13 @@ public function param_builder_server_setup() {
 				} else {
 					$cookie_to_set = $param_builder->getCookiesToSet();
 
-					if ( ! headers_sent() ) {
+					if ( ! headers_sent() && ! empty( $cookie_to_set ) ) {
```
vahidkay-meta, line 196: Why is this needed? If $cookie_to_set is empty, the foreach will not work. Will it?

vahidkay-meta (changes requested):

PR #3964, opened 2026-06-29
ci(e2e): fix flaky Global Setup login + reduce shared-asset contention

vahidkay-meta:
## Summary
Two E2E reliability fixes.

### 1. Fix flaky Global Setup login (root cause of the `#wpcontent` timeouts)
`global-setup.js` gated login on a **1-second** `#user_login` `isVisible` check. On a slow/contended CI runner the `/wp-admin/` → `wp-login.php` form renders later than 1s, so login was **skipped entirely**, and the run then waited 60s for the admin `#wpcontent` wrapper that never appears — the Global Setup `TimeoutError` seen on `product-crud`, `performance-sync`, etc.

Now the login flow:
- explicitly waits for the login form (up to 60s) instead of racing a 1s check,
- drops the flaky `networkidle` wait and relies on the real post-login signal (`#wpcontent`/`#wpadminbar`),
- retries the whole flow up to 3× for transient hiccups.

### 2. Reduce contention on the shared Facebook test asset
All shards in a run select the **same** FB asset set (the selector is `github.run_number`, constant within a run), so they pound one catalog/pixel concurrently. To reduce that:
- `max-parallel: 4` on the matrix,
- merge `plugin-events` back into a single shard,
- raise its timeout to 90m.

## Test plan
Opening this PR runs the full E2E suite; expecting the Global Setup timeouts to stop and the matrix to run ≤4 shards at a time.

### DIFF
--- a/.github/workflows/product-creation-tests.yml
+++ b/.github/workflows/product-creation-tests.yml
@@ -125,14 +125,17 @@ jobs:
     if: ${{ needs.check-secrets.outputs.secrets-configured == 'true' }}
     runs-on: ubuntu-latest
     # Fail fast instead of letting a stuck step (e.g. a hung browser install) run
-    # until GitHub's 6-hour default job timeout. The plugin-events shards run many
-    # headed browser projects sequentially, so they get a larger budget.
-    timeout-minutes: ${{ startsWith(matrix.test-group, 'plugin-events') && 75 || 45 }}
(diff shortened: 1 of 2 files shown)


vahidkay-meta:
Closing — these E2E reliability changes were moved onto #3960's branch (chore/dependabot-changelog-label) instead of a separate PR.

PR #3995, opened 2026-07-29
test(e2e): cover Search event on classic + block theme projects

vahidkay-meta:
## Summary

Closes gap #6 from the "Reducing manual regression scope after the E2E expansion" review. The Search and *Search - No Results* event tests were **skipped** on the `chromium-wp-customer-classic-theme` and `chromium-wp-customer-block-theme` projects because those themes render no header search UI. Per the reliability gate, **a skipped configuration is not coverage evidence** — so the manual Search step couldn't be retired for the classic/block architectures.

This enables real Search coverage on both theme architectures.

### Investigation (why not just "pick a theme with search")
I inspected the default header of every stock WP theme and the popular third-party FSE themes:
- **Classic:** Twenty Thirteen and Twenty Eleven ship a persistent, load-visible `get_search_form()` in `header.php`. Twenty Twenty hides search behind a toggle; Twenty Fifteen/Sixteen/Seventeen/Twenty-One have none.
- **Block/FSE:** *none* — stock (Twenty Twenty-Two → Five) **and** popular third-party (Ollie, Frost, Tove, Aino, Wabi, Pendant, Blockbase, Raft) all ship header = title + nav, no search.

So the classic half is a clean theme swap; the block half genuinely needs injection.

### Changes
- **Classic** → switch project theme `twentytwentyone` → **`twentythirteen`** (native header search, no injection).
- **Block** → keep `twentytwentyfive`; a mu-plugin (installed **only** for the block-theme project, removed in `afterAll`) injects a `core/search` block into the header template part via the `get_block_template` / `get_block_file_template` filters (covers file- and DB-based parts). Harmless on classic themes (no block template parts to match).
- Remove the classic/block project-name skips; generalize the brave-only hard-fail to `PROJECTS_REQUIRING_SEARCH` (brave + classic + block) so a missing search UI on these configs **fails loudly instead of silently skipping**.
- The responsive/mobile skip (search behind a hamburger on `android-pixel` / `safari-ios`) is intentionally left unchanged — a genuinely different fixture limitation.

`submitSearch` already forces `post_type=product`, so both the native and injected core search forms fire the plugin's WooCommerce Search event the same way the Storefront product-search form does.

### Test plan / verification
Because the tests now **hard-fail** (not skip) on classic + block, CI is the verification: if Twenty Thirteen's native search or the injected block search doesn't render/fire the Search event, the `Search` and `Search - No Results` tests on those two projects will go red. Watching the first run on:
- `E2E Tests - * - plugin-events` (classic-theme + block-theme projects).

Local checks: `node --check` on all changed JS, `php -l` on the generated mu-plugin, and a runtime `require()` of the helper barrel all pass.

### DIFF
--- a/tests/e2e/events-test.spec.js
+++ b/tests/e2e/events-test.spec.js
@@ -13,6 +13,8 @@ const {
   removeJsErrorSimulatorMuPlugin,
   installSingleSearchRedirectBlockerMuPlugin,
   removeSingleSearchRedirectBlockerMuPlugin,
+  installBlockThemeHeaderSearchMuPlugin,
+  removeBlockThemeHeaderSearchMuPlugin,
   getVisibleSearchInput,
   createVariableProductEventFixture,
   createGroupedProductEventFixture,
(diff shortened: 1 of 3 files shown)


marwanwanoo482-design:
وش ذا اختراق ؟

في الأربعاء، ٢٩ يوليو ٢٠٢٦، ٦:٠٣ م github-actions[bot] <
***@***.***> كتب:

PR #4017, opened 2026-08-27
ci(e2e): harden shared catalog test execution

jczhuoMeta:
## Description

Hardens the shared Meta catalog E2E pipeline against cross-shard contention and slow catalog propagation.

- Spreads the 12 E2E shards across three Meta test asset sets and caps each run at three active shards.
- Reserves unique WordPress product and category ID ranges per run/shard to prevent remote catalog collisions.
- Drains pending sync jobs before catalog validation and shares one drain across concurrent validators.
- Removes redundant explicit drains from callers.
- Avoids suite-wide retry replay, runs product CRUD with one worker, reduces the PR-only performance batch, and gives slower shards more appropriate timeouts.

This draft changes only test infrastructure and test code; it does not alter production behavior or user data handling.

### Type of change

- Tweak (test infrastructure reliability)

## Checklist

- [x] No PHP production files are changed.
- [x] I added focused tests and all JavaScript unit tests pass locally.
- [ ] Full E2E validation is pending.

## Changelog entry

N/A — test infrastructure only.

## Test Plan

- [x] Run `npm run test:js -- --runInBand` (107 tests).
- [x] Run `node --check` on every changed JavaScript file.
- [x] Parse the workflow YAML and verify all 12 test/version combinations, balanced asset assignments, and unique shard indexes.
- [x] Verify the branch is one commit directly on latest `main` and its diff contains no Topics API or Privacy Sandbox changes.
- [x] Run `git diff --check origin/main...HEAD`.
- [ ] Run the full GitHub E2E matrix against the real Meta test assets.

## Screenshots

Not applicable.

### DIFF
--- a/.github/workflows/product-creation-tests.yml
+++ b/.github/workflows/product-creation-tests.yml
@@ -120,7 +120,7 @@ jobs:
           echo "Resolved versions — WP: $WP_VERSION, WC: $WC_VERSION"
 
   e2e-tests:
-    name: E2E Tests - ${{ matrix.version-label }} - ${{ matrix.test-group }}
+    name: E2E Tests - ${{ matrix.version-label }} - ${{ matrix.test-group }} - Asset ${{ matrix.asset-index }}
     needs: [check-secrets, resolve-versions]
     if: ${{ needs.check-secrets.outputs.secrets-configured == 'true' }}
     # Run every shard on the standard GitHub-hosted runner. We previously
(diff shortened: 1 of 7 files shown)


ukilla:
Since this overlaps with #4015 and #4018, all three add a sync drain and asset lanes to the same workflow and `sync.js`.

Maybe we can use #4015 as the base for both this and #4018. Its catalog detection falls back to sequential when fewer than three catalogs are configured, where this one hard-fails, and it already serializes drains and waits for batch completion. Once it lands, the parts worth carrying over on top of it are the shared drain and Jest coverage from here, and the product-set membership polling and its Jest tests from #4018.

PR #4016, opened 2026-08-26
ci(e2e): fix every known cause of the red E2E pipeline

jczhuoMeta:
## Description

`main`'s nightly E2E has been red continuously. This is the single PR fixing **every** known cause. Each was identified from real CI logs, not inspection. Supersedes #4020 and #4022, both folded in and closed.

### Deterministic

**1. Obsolete Topics API test.** Chrome removed `document.browsingTopics()`, so `events-test.spec.js:1341` failed 3/3 attempts on *every* nightly run 08-26 → 08-31. Removes the test and its launch flags; narrows the Privacy Sandbox gate to the surviving Protected Audience test.

### Wall clock — the 45-minute job timeouts

**2. `performance-sync`** and **3. `variable-product-depth`** both used `test.describe.serial`, which re-runs the **entire block** when any test fails. Caught in the log:

| 03:48:50 | batch test **passes** (13.7m) |
| 03:54:13 | variable-product test **fails** (3m) |
| 03:54:14 | batch test **restarts** ← whole block retried |

3 × ~18m against a 45m budget. Both files already run `--workers=1`, so declaration order holds without `.serial`, and neither has mutable state at describe scope. This took `performance-sync` from a **45m timeout to a 23m pass**. No spec in `tests/e2e/` uses `describe.serial` after this.

### Flakes no retry could recover

**4. `Check WooCommerce logs`** re-reads the same **cumulative** log file every attempt, so once Meta wrote a 500 into it (`items_batch`, 42.6s, error code 1 / subcode 99) no retry could clear it — all three attempts failed identically. A 5xx is now reported but tolerated; **a 4xx still fails**, because that's the plugin sending a malformed request.

**5. `Exclude product from sync`** asserted on a *single probe* that the item was gone, while the sync validator polls until an item *appears*. Exclusion propagates asynchronously, so any lag failed instantly — and the retry just rebuilt the product and raced again. Now polls until absent with a 2-minute deadline; each probe is one API call, so a still-present item returns immediately.

**6. The `Select all` button** only renders after WooCommerce finishes an AJAX round-trip loading the attribute's terms, but was waited on with a 5s timeout. Under CI contention that regularly expired — the most common flake in this shard.

### Visibility

[...]

PR #4020, opened 2026-09-01
ci(e2e): report every browser project instead of aborting on the first failure

jczhuoMeta:
## Description

The `Run Plugin & Events tests` step runs **twelve** browser projects back to back in a single `run:` block. GitHub executes step scripts with `shell: /usr/bin/bash -e`, so the **first non-zero exit aborted the whole step** and every project after it was silently skipped — reported as neither passed nor failed.

This is not hypothetical:

| Where | First failing invocation | Projects silently skipped |
|---|---|---|
| `main` (nightly, 08-28 → 08-31) | #5 `chromium-privacy-sandbox` (obsolete Topics API test) | edge, firefox, brave, android-pixel, safari-ios, opera |
| PR #4014 (08-26) | #1 `plugin-level-tests` (flaky Select2 locator) | **all 11** `events-test` project runs |

So since 2026-08-27 the nightly job has not executed roughly half of the browser coverage it claims to run, and nothing in the CI output says so.

This change collects per-project results and fails **once at the end**, listing exactly which projects failed. A single flaky project can no longer erase the rest of the shard's coverage.

### Type of change

## Notes for the reviewer

- **No test, assertion, or product code changes.** Only the step's error handling.
- The step still exits non-zero whenever any project fails — this does not make CI more permissive.
- `set +e` is scoped to after the `cd`, so a failed `cd` still aborts immediately.
- The two `node` gates (capability probe and no-skip gate) are captured explicitly via `note_rc $?`, since heredocs can't route through the `run_project` wrapper.
- The opera invocation keeps its `set -o pipefail` inside the wrapped subshell, so `tee` cannot mask a failure.
- **Expect this to surface previously-hidden failures.** That is the point, but the first run may look worse than before because projects that were being skipped will now report honestly.
- **Related:** this guarantees the full tail runs on every job. Observed full-step durations are 68–76m against the `timeout-minutes: 90` cap, so the headroom is now thinner. If a run hits the cap, the follow-up is a timeout bump — deliberately not bundled here.
- **Independent of #4016.** Both are cut against `main` and both touch this step, so whichever lands second needs a trivial rebase. #4016 removes the Topics test that is currently the trigger on `main`; this PR makes any such trigger stop hiding the projects behind it.

## Checklist

- [x] No PHP files are changed.
- [x] All existing JavaScript unit tests pass locally.

## Changelog entry

N/A — CI infrastructure only.

## Test Plan

- [x] Parse `.github/workflows/product-creation-tests.yml` as YAML and extract the step script.
- [x] `bash -n` on the extracted step script.
- [x] Behavioural check of the harness: a failing project does not stop later projects; `pipefail` still surfaces the left-hand exit code; heredoc gates are captured; an all-green run records no failures.
- [x] `npm run test:js -- --runInBand` (215 tests).
- [x] `git diff --check`.
- [ ] CI: confirm the six previously-unreachable browser projects actually report a result.

### DIFF
--- a/.github/workflows/product-creation-tests.yml
+++ b/.github/workflows/product-creation-tests.yml
@@ -827,19 +827,55 @@ jobs:
           FIREFOX_EXECUTABLE_PATH: ''
         run: |
           cd ${{ env.PLUGIN_PATH }}
-          npx playwright test plugin-level-tests.spec.js --project=chromium-wp-admin --workers=1
-          xvfb-run --auto-servernum npx playwright test events-test.spec.js --headed --project=chromium-wp-customer --workers=1
-          xvfb-run --auto-servernum npx playwright test events-test.spec.js --headed --project=chromium-wp-customer-classic-theme --workers=1
-          xvfb-run --auto-servernum npx playwright test events-test.spec.js --headed --project=chromium-wp-customer-block-theme --workers=1
-          xvfb-run --auto-servernum npx playwright test events-test.spec.js --headed --project=chromium-privacy-sandbox-wp-customer --workers=1
+
+          # This step runs a dozen browser projects back to back. The step shell is
+          # `bash -e`, so the first non-zero exit used to abort everything after it:
+          # a single failing project silently skipped the remaining projects and
+          # reported them as neither passed nor failed. Collect failures instead and
+          # fail once at the end, so every project always reports its own result.
+          set +e
+          failed_projects=()
+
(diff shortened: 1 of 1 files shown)


jczhuoMeta:
Folded into #4016, which is now the single PR fixing every known E2E CI breakage. The failure-collecting harness is carried over verbatim, with the Privacy Sandbox `--grep` reconciled against the Topics removal in the same PR.

PR #4030, opened 2026-09-09
Use delegated access tokens for Shops management URLs

jczhuoMeta:
## Description

Updates the connected Shops management flow to request a short-lived delegated access token from the Commerce Extension token endpoint and use it to build the Commerce Partner Hub Overview URL. The flow uses the Commerce Partner Integration ID stored with the active connection.

For backward compatibility, the existing business-configuration lookup remains available as a fallback when the integration ID is unavailable, the new request fails, or its response does not contain a usable token. A successful primary or fallback request clears stale invalid-connection state. Failed Stefi requests do not mark the connection invalid; the legacy Graph API framework remains responsible for doing so when it receives a real authentication error.

The change also redacts the legacy management URI from safe response serialization so credentials embedded in that URI are not written to debug logs. Focused tests cover the primary path, fallback behavior, connection-state handling, iframe rendering, and log redaction.

### Type of change

## Checklist

- [x] I have checked plugin debug logs and found no new PHP warnings or fatal errors.
- [x] I followed general pull request best practices.
- [x] I added tests and the relevant unit tests pass locally.
- [x] I completed dogfooding and fallback-path QA testing.
- [x] Documentation updates are not required because this does not change the user interface or setup steps.

## Changelog entry

Use delegated access tokens when opening the Shops management experience.

### Automated

- [x] Focused PHP unit suites covering the change (`ShopsTest`, `MetaExtensionTest`, `FBEConfigurationReadResponseTest`) — green in CI on PHP 7.4 and 8.4.
- [x] JavaScript unit tests: 220 tests.
- [x] Full PHP unit suite on PHP 7.4 and PHP 8.4.
- [x] Integration tests on PHP 7.4 and PHP 8.4.
- [x] PHP coding standards on changed files.
- [x] PHP syntax checks on every changed PHP file.
- [x] `git diff --check`.

### Manual

1. On a connected test store, open **Marketing > Facebook > Shops > Overview**.
2. Confirm the Commerce Extension token request succeeds and the Overview page loads from Commerce Partner Hub.
3. Confirm the iframe URL does not contain the stored long-lived access token.
4. In a local test environment, make the new endpoint return `404`, then refresh the settings page.
5. Confirm the legacy management URL is requested immediately afterward and Overview still loads.
6. Restore the endpoint, reload, and confirm the primary path succeeds again.
7. Review plugin debug logs and confirm no access tokens, new PHP warnings, or fatal errors are present.

## Updates

- Rebased onto latest `main` (includes #4035); kept both the merchant-token-cleanup iframe tests and the new token-endpoint tests.
- Fixed a `ShopsTest` teardown filter leak (teardown now removes test filters before resetting the cached API client).
- Review feedback: legacy path uses the shared `get_api()` client; token-endpoint transport moved into `Finalize\Client::request_commerce_extension_token()` returning a `CommerceExtensionToken\Response` per the codebase Response pattern.

## Screenshots

Not applicable; there is no user-interface change.


ukilla:
`MetaExtension.php:149-151` sets `wc_facebook_connection_invalid` when the new endpoint returns 401 and the legacy fallback throws for *any* reason (the catch is on `\Throwable`, so timeouts and 5xx count too). The events tracker reads that transient before every CAPI send (`facebook-commerce-events-tracker.php:1790`), so a Stefi (new token endpoint) 401 plus a network hiccup would stop server-side events for up to 24 hours.

Maybe we should drop the manual `set_transient` and let the API framework's own auth-error handling in the legacy path set that flag, as it already does for real Graph auth errors.

jczhuoMeta:
Addressed in 5e67a4c4d. Removed the manual `wc_facebook_connection_invalid` write from the Stefi/fallback orchestration. The legacy API framework remains the sole owner of that state, so a Stefi 401 followed by a legacy transport failure no longer disables CAPI; a real Graph OAuth error still marks the connection invalid. Added a regression assertion for the 401 + transport-failure case, and the focused MetaExtension suite passes (14 tests, 86 assertions).

includes/Handlers/MetaExtension.php
```diff
@@ -96,34 +99,143 @@ public static function generate_iframe_splash_url( $is_connected, $plugin, $exte
...
+	 * Generates the management URL using the legacy Graph API endpoint.
+	 *
+	 * @param string $external_business_id External business ID.
+	 * @param string $access_token Long-lived business integration system user token.
+	 *
+	 * @return string
+	 * @throws \Exception If the legacy endpoint does not return a management URL.
+	 */
+	private static function generate_legacy_iframe_management_url( $external_business_id, $access_token ) {
+		$api        = new API( $access_token );
```
vahidkay-meta: Why isn't it using new API and not facebook_for_woocommerce()->get_api() ?

includes/Handlers/MetaExtension.php
```diff
@@ -96,34 +99,143 @@ public static function generate_iframe_splash_url( $is_connected, $plugin, $exte
...
+	 * @throws \Exception If the endpoint request fails or does not return a delegated access token.
+	 */
+	private static function generate_iframe_management_url_with_commerce_extension_token( $external_business_id, $commerce_partner_integration_id, $access_token ) {
+		$endpoint = sprintf(
+			'%s/commerce-partner-integrations/%s/commerce-extension-token',
+			self::BASE_STEFI_ENDPOINT_URL,
+			rawurlencode( $commerce_partner_integration_id )
+		);
+
+		$response = wp_safe_remote_post(
```
vahidkay-meta: wouldn't it be possible to re-use this: finalize_install in Client.php ?

includes/Handlers/MetaExtension.php
jczhuoMeta: Good catch — switched to the factory.

`generate_legacy_iframe_management_url()` now calls `facebook_for_woocommerce()->get_api( $access_token )` instead of constructing `new API( $access_token )` directly, so the legacy fallback shares the same API instance and configuration as the rest of the plugin.

includes/Handlers/MetaExtension.php
jczhuoMeta: Yes — done.

The inline `wp_safe_remote_post()` call has moved out of `MetaExtension` and into `CommerceIntegration\Finalize\Client` as `request_commerce_extension_token()`, sitting alongside `finalize_install()`. Both now share the same `Exception` type and the same `get_http_failure_reason()` status mapping, so error handling is consistent across the two endpoints.

`MetaExtension` just calls the client:

```php
$client         = new CommerceIntegrationClient();
$token_response = $client->request_commerce_extension_token( $access_token, $commerce_partner_integration_id );
```

The endpoint constant and the raw HTTP/JSON handling are gone from `MetaExtension.php`.

PR #4035, opened 2026-09-10
Tweak - Stop writing redundant legacy connection options

jczhuoMeta:
## Description

The current Commerce Extension flow issues one system-user access token, but the plugin still duplicated it into the legacy merchant-token option and continued writing several retired FBE and Messenger options.

This PR removes those internal writes and makes the canonical access token the sole connection signal for the Shops management iframe. It also centralizes the retired option names so disconnect and uninstall continue deleting stale values.

### Changes

- Remove `merchant_access_token` from the Commerce Extension finalize-install payload and the legacy `settings/update` schema and mapping.
- Require exact alignment between the generated finalize-install payload and its request schema.
- Gate the management iframe on `Connection::is_connected()` rather than the legacy merchant-token option.
- Stop internal connection flows from writing the retired merchant token, system-user ID, FBE status flags, Messenger flag, and attribute-sync timestamp.
- Keep the public legacy token/system-user setters deprecated but functional until a breaking release, preserving compatibility for external callers.
- Keep the merchant-token crash-reporting fallback for existing legacy installations; the option is no longer used for authentication.
- Delete all retired option rows during both disconnect and uninstall.
- Update the reset-settings E2E test to verify non-rendered retired options through WordPress rather than waiting for fields that do not exist on `options.php`.

Existing stored values are not migrated or deleted while a store remains connected.

### Privacy and compliance

No new user-data processing is introduced. The change reduces credential duplication by stopping new Commerce Extension connections from persisting a second copy of the access token.

### Type of change

- Tweak (non-breaking cleanup with a user-visible routing fix)

## Checklist

- [x] I have checked plugin debug logs that my changes do not introduce any new PHP warnings or fatal errors.
- [x] I followed general pull request best practices.
- [x] I added or updated tests for the changed behavior.
- [x] Documentation updates are not required because this does not change setup steps or user-facing copy.

## Changelog entry

Tweak - Stop persisting redundant legacy connection state and use the canonical access token for Shops management.

### Automated

- [x] Full PHP unit suite: 2,004 tests, 143,389 assertions.
- [x] Focused PHP tests: 40 tests, 212 assertions across Shops rendering, finalize-install, settings update, both cleanup paths, and the legacy crash-reporting fallback.
- [x] JavaScript unit tests: 220 tests.
- [x] PHPCS on all changed production PHP files.
- [x] PHP syntax checks on all changed PHP files.
- [x] JavaScript syntax check for the updated E2E test.
- [x] `git diff --check`.

### Manual verification

1. Disconnect and reconnect a test store through the Commerce Extension iframe.
2. Confirm `settings/finalize-install` succeeds without `merchant_access_token` in its payload.
3. Reload **Marketing > Facebook > Shops** and confirm the management iframe appears using the canonical access token.
4. Disconnect or use **WooCommerce > Status > Tools > Reset connection settings**.
5. Confirm the active connection options and all retired options are removed.

## Screenshots

Not applicable; there is no visual change.


ukilla:
The plugin-events E2E test is failing here because on disconnect this PR deletes the deprecated options which is fine, but the "Reset connection settings" E2E test (`tests/e2e/plugin-level-tests.spec.js:541-558`) still reads `wc_facebook_merchant_access_token` and `wc_facebook_system_user_id` on `options.php`, so both plugin-events shards time out.  

We can drop those two from the `fbOptions` list and assert they're gone via `execWP(get_option(...))`, as the test already does for `wc_facebook_page_access_token`

jczhuoMeta:
Addressed in 58a6039a6. The reset-settings E2E test no longer waits for merchant/system-user fields on `options.php`; it seeds all six retired options and verifies their deletion directly with `get_option()`. I also rebased onto current `main`, preserved the landed webhook deletion and finalize-install normalization, and added unit coverage for both cleanup paths.

PR #4042, opened 2026-09-17
Gate Value Optimization behind a rollout switch, and fix its net revenue calculation

vahidkay-meta:
## Summary

Value Optimization was held back by a hardcoded constant on the events tracker:

```php
/** @var bool disable VO while product is not GA */
const IS_VO_ENABLED = false;
```

Turning it on therefore meant shipping a new plugin release. This replaces the constant with a server-controlled rollout switch (`woocommerce_vo_switch`), fixes a bug in the net revenue arithmetic that the switch makes reachable, and adds a few Local-environment helper scripts used while testing it.

## 1. Gate VO behind a rollout switch

- `includes/RolloutSwitches.php` — add `SWITCH_VALUE_OPTIMIZATION_ENABLED` and register it in `ACTIVE_SWITCHES`.
- `facebook-commerce-events-tracker.php` — remove `IS_VO_ENABLED`; the Purchase path now calls a new private `is_value_optimization_enabled()` helper, which returns `false` if the plugin or the switches handler is not available yet (it is untyped and unset early in bootstrap).
- `tests/Unit/RolloutSwitchesTest.php` — cover the constant, membership in `ACTIVE_SWITCHES`, and the enabled / disabled / absent responses.

Both directions were exercised against a live store:

| Response | Stored option | `is_switch_enabled()` |
| --- | --- | --- |
| `{"switch":"woocommerce_vo_switch","enabled":""}` | `no` | `false` |
| `{"switch":"woocommerce_vo_switch","enabled":"1"}` | `yes` | `true` |

## 2. Fix double-subtraction of shipping tax

`WC_Abstract_Order::get_total_tax()` is set as cart tax **plus** shipping tax — `set_total_tax( get_cart_tax() + get_shipping_tax() )`, called from both `set_cart_tax()` and `set_shipping_tax()` (`abstract-wc-order.php:1041,1053`). The calculation subtracted shipping tax a second time:

```diff
 $order_value_excluding_tax_including_discounts = $order->get_total()
     - $order->get_total_tax()
-    - $order->get_shipping_total()
-    - $order->get_shipping_tax();
+    - $order->get_shipping_total();
```

Net revenue came out low by the shipping tax, and an order whose shipping tax pushed the result to zero or below dropped `net_revenue` from the payload entirely, because of the `$net_profit > 0` guard.

Worked example — item 100, shipping 10, cart tax 20, shipping tax 2, COGS 40: the old formula reported 58, the corrected one reports 60.

This was dormant while VO was pinned off by the constant; gating it on a switch makes the path reachable, so it is fixed here rather than left behind the new switch.

`tests/Unit/Events/FacebookCommerceEventsTrackerTest.php` gains four regression tests that drive `inject_purchase_event()` and assert on `get_tracked_events()`:

| Case | Expectation |
| --- | --- |
| shipping tax 2, COGS 40 | `net_revenue` = 60 (58 before) |
| no shipping tax, COGS 40 | `net_revenue` = 60 — base case, both formulas agree |
| COGS unavailable | no `net_revenue`, `value` still present |
| COGS above net order value | no `net_revenue` |

A shared helper builds the order and injects an anonymous COGS stub by reflection, so the tests do not require a Cost of Goods integration to be installed.

## 3. Local environment helper scripts

`bin/lib/local-wp.sh` resolves a site from Local's `sites.json` and exposes `wp_local`; Local's bundled `wp-cli.phar` needs the site's generated `php.ini`, because `wp-config.php` sets `DB_HOST` to `localhost` and only that `php.ini` knows the mysqld socket. `bin/get_values_for_test_env_secrets.sh` reads the connection options as JSON, and `bin/reset_plugin_options.sh` clears them so onboarding can be replayed as a new install.

`wc_facebook_merchant_access_token` is excluded from the secrets helper: the option is deprecated and no longer read anywhere, only written (`Connection.php:734`).

These are macOS + Local specific. Happy to split them into their own PR if that is preferred.

## Behaviour

No change for merchants until the switch is turned on. It reports disabled by default, so `net_revenue` stays off Purchase events exactly as it is today.

One caveat worth recording: `RolloutSwitches::is_switch_enabled()` treats an active switch that is *missing* from an otherwise populated response as **enabled** (`RolloutSwitches.php:141`). That is deliberate for older plugins talking to a newer backend, but it is the opposite of what this feature wants. It is unreachable while the switch is always reported, and a test pins the behaviour with a comment so the risk is visible if that ever changes.

## Test instructions

1. Check out this branch on a store connected to Meta with a pixel configured.
2. Confirm the default is off:
   - `wp option delete wc_facebook_for_woocommerce_rollout_switches`
   - `wp transient delete _wc_facebook_for_woocommerce_rollout_switch_flag_<plugin-version>`
   - Load any WP Admin page to trigger the refetch, then check the stored option contains `"woocommerce_vo_switch":"no"`.
   - Place an order **as a guest or non-admin** — Purchase tracking is skipped for users with `manage_woocommerce`.
   - The Purchase event's `custom_data` should contain `value` but no `net_revenue`.
3. Turn the switch on:
   - `wp option patch update wc_facebook_for_woocommerce_rollout_switches woocommerce_vo_switch yes`
   - Set a Cost of Goods value on the product. `calculate_cogs_for_products()` bails if any line item has a cost of zero.
   - Place another order, with shipping **and** shipping tax, as a guest.
   - `custom_data` should now include `net_revenue`, equal to `total − total_tax − shipping_total − COGS`.
4. Turn it back off and confirm `net_revenue` disappears again.

Two notes when testing against the live response rather than patching the option: `RolloutSwitches::init()` caches switches in a transient keyed by plugin version for 60 minutes, and the plugin's `log()` discards everything unless debug mode is enabled while the store is connected.

## Unit tests

```
composer test-unit -- --filter 'RolloutSwitchesTest|FacebookCommerceEventsTrackerTest'
```


facebook-commerce-events-tracker.php
```diff
@@ -1324,6 +1321,28 @@ private function is_purchase_trackable_order( $order ) {
...
+		 * @return bool
+		 */
+		private function is_value_optimization_enabled(): bool {
+			$plugin = facebook_for_woocommerce();
+
+			if ( ! $plugin || ! $plugin->get_rollout_switches() ) {
+				return false;
+			}
+
+			return (bool) $plugin->get_rollout_switches()->is_switch_enabled(
```
ukilla, line 1341: When the stored VO entry is missing, it defaults to enabled. After an upgrade, cached switches from the previous version lack this key, so the shared handler returns true before the next admin-triggered refresh. Should we default VO to disabled when the entry is missing and add a test covering this upgrade scenario?

facebook-commerce-events-tracker.php
```diff
@@ -1470,7 +1489,7 @@ public function inject_purchase_event( $order_id ) {
 				'event_id'    => $event_id,
 			);
 
-			if ( self::IS_VO_ENABLED ) {
+			if ( $this->is_value_optimization_enabled() ) {
```
ukilla, line 1492: We have a bug in this exposed path: `get_total_tax()` already includes shipping tax, which is subtracted again below. I think we can remove the extra shipping-tax deduction and add a Purchase-payload regression test.

bin/get_values_for_test_env_secrets.sh
```diff
@@ -0,0 +1,57 @@
...
+# Exits non-zero if any option is empty; the JSON is still written to stdout so
+# callers can inspect what was found. Warnings go to stderr.
+#
+set -euo pipefail
+
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/lib/local-wp.sh"
+
+OPTION_KEYS=(
+	wc_facebook_access_token
+	wc_facebook_merchant_access_token
```
ukilla: Since this option `wc_facebook_merchant_access_token` is deprecated and current onboarding no longer populates it, I think we can remove it from required validation.

facebook-commerce-events-tracker.php
vahidkay-meta: The switch is already 100% on. So it being defaul=true doesen't introduce an issue. 
The only reason I added it behind a killswitch is that if it starts causing issues we could shut it down.

PR #4049, opened 2026-09-18
Move the Commerce Partner Integration client out of the Finalize namespace

jczhuoMeta:
## Description

`API\CommerceIntegration\Finalize\Client` is named after one of its operations, but every endpoint on it is a sub-path of the same Meta endpoint:

```
/commerce-partner-integrations/finalize-install
/commerce-partner-integrations/{id}/commerce-extension-token
```

So it is already **one client per endpoint, one method per operation** — mirroring the server-side schema. Only the name suggests otherwise, which makes it read as though finalize had grown unrelated responsibilities such as token minting.

This moves `Client` and `Exception` up to the `CommerceIntegration` namespace. **Responses stay grouped per operation**, because each returns a genuinely different shape — it is the transport that is shared, not the payload.

Grouping by server endpoint is the more stable axis: the endpoint contract changes less often than plugin internals, and when a new sub-path is added there is an obvious home for it.

### Type of change

- Tweak (non-breaking change which fixes code modularity, linting or phpcs issues)

## Changes

| From | To |
|---|---|
| `API\CommerceIntegration\Finalize\Client` | `API\CommerceIntegration\Client` |
| `API\CommerceIntegration\Finalize\Exception` | `API\CommerceIntegration\Exception` |

- `Finalize\Response` is unchanged and still lives beside the other per-operation responses; the client now imports it explicitly.
- Consumer imports updated in `Plugin\Settings\Handler` and `Handlers\MetaExtension`.
- Test moved alongside the class.

No behaviour change. Both classes are internal to the plugin with no external consumers, so no deprecation shim is required — unlike the public accessors retired in #4044 / #4045.

## Checklist

- [x] I have completed dogfooding and QA testing.
- [x] Documentation updates are not required — internal namespace only.

## Changelog entry

Dev - Group the Commerce Partner Integration API client by endpoint rather than by operation.

### Automated

- `php -l` clean on all changed files.
- `vendor/bin/phpcs` clean, 5/5 production files.
- Repo-wide grep confirms no remaining references to the old namespaces.

### Runtime autoload verification

Autoloading is PSR-4 (`WooCommerce\Facebook\` → `includes`), so the move must resolve at the new paths. Verified by bootstrapping WordPress on a connected store:

```
WooCommerce\Facebook\API\CommerceIntegration\Client                            OK
WooCommerce\Facebook\API\CommerceIntegration\Exception                         OK
WooCommerce\Facebook\API\CommerceIntegration\Finalize\Response                 OK
WooCommerce\Facebook\API\CommerceIntegration\CommerceExtensionToken\Response   OK
instantiate + methods: finalize_install, request_commerce_extension_token
```

### Manual

1. On a connected store, load **Marketing → Facebook → Shops** and confirm the management iframe renders (exercises `request_commerce_extension_token()`).
2. Re-onboard a store and confirm finalize-install still succeeds (exercises `finalize_install()`).

## Screenshots

N/A — no UI change.


jczhuoMeta:
Folded into #4048 — the rename is most legible alongside the two reads that PR adds, and keeping them together avoids a conflict on a moved file.

PR #4054, opened 2026-09-22
Stop the Attribute Mapping tab requesting a script that is never built

vahidkay-meta:
## Description

`Marketing > Facebook > Attribute Mapping` enqueues `assets/build/admin/product-attributes.js`. That file has no webpack entry, so it has never existed in a build, and the browser gets a 404 on every visit to the tab.

The 404 is not inert. WordPress's default rewrite routes requests for missing files under `wp-content` through `index.php`, so the request renders the site's 404 template — and `inject_base_pixel()` fires a **PageView CAPI event for a missing JavaScript file**.

## Steps to reproduce

1. `Marketing > Facebook > Attribute Mapping`
2. Network tab — `product-attributes.js?ver=3.7.7` returns 404

## The fix

The handle is registered with no source. It is kept rather than dropped because it is load-bearing in two ways that are easy to miss:

- **It pulls in select2.** This screen lives under Marketing, so its screen ID is not in `wc_get_screen_ids()` and WooCommerce does not enqueue `wc-enhanced-select` here on its own. The dependency on that handle is the only reason the mapping dropdowns are enhanced selects.
- **The notice dismissal script hangs off it** via `wp_add_inline_script()`, which is dropped if the handle goes away.

A source-less handle still resolves its dependencies and still prints its inline scripts, so both keep working with no request for the missing file.

## Why not just add the webpack entry

Building `assets/js/admin/product-attributes.js` as it stands would introduce two regressions, because it has drifted from the inline script `render()` prints:

- Both bind `.add-new-mapping` directly, so one click would append **two** rows.
- Both bind a delegated `change` on `.wc-attribute-search`, and the file's handler renames the Meta field select to `wc_facebook_field_mapping[<attribute>]` while the row's attribute select keeps `[<index>]`. `process_form_submission()` pairs the two arrays by key, so the pair no longer matches and **new mappings are silently not saved**.

The file has not been touched since the feature landed (apart from a copyright header sweep). It is removed here — leaving it in the tree is an invitation to "fix" the 404 by adding the entry and shipping both of the above.

## Testing

`tests/Unit/Admin/Settings_Screens/Product_AttributesTest.php` pins the handle carrying no source, keeping its dependencies, still carrying the inline dismissal script, enqueueing only a stylesheet that exists on disk, and doing nothing on other tabs.

## Adjacent, not fixed here

`ajax_dismiss_notice()` is unreachable: it calls `is_current_screen_page()`, which reads `$_GET['page']` and so is always false over `admin-ajax.php`, and its `check_ajax_referer( 'dismiss_facebook_attributes_notice', 'nonce' )` does not match the nonce the inline script sends (`fb_dismiss_attribute_notice` as `security`). The notice is hidden client-side and its transient is already deleted at render time, so nothing is visibly broken. Left alone to keep this change to the 404. Happy to fix separately.

Firing PageView on 404 pages at all is a broader signal-quality question — it happens for any bad URL on the store, not just this one. Also left for a separate discussion.


includes/Admin/Settings_Screens/Product_Attributes.php
```diff
@@ -83,9 +83,12 @@ public function enqueue_assets() {
 
 		wp_enqueue_style( 'woocommerce_admin_styles' );
 
+		// This screen has no bundled script of its own: the behavior it needs is printed inline by
+		// render() and by the inline script below. The handle is registered without a source so it
+		// can still carry those dependencies and the inline script.
 		wp_enqueue_script(
```
ukilla: The issue here is that passing `false` as $src to `wp_enqueue_script()` skips registration entirely, since core only calls $wp_scripts->add() when there is an actual source. Because of this the handle isn't enqueued and `wp_add_inline_script()` has nothing to attach to, which is what the three failing tests show.

We can register the alias first, then enqueue it:

```suggestion
		wp_register_script(
			'facebook-for-woocommerce-product-attributes',
			false,
			array( 'jquery', 'jquery-tiptip', 'wc-enhanced-select' ),
			\WC_Facebookcommerce::PLUGIN_VERSION,
			true // Load in footer
		);
		wp_enqueue_script( 'facebook-for-woocommerce-product-attributes' );
```

With this change there is no request for the missing file, the inline dismissal script is printed, and the mapping dropdowns are still enhanced selects. As-is, the 404 goes away but the inline script is dropped.

ukilla (changes requested):

PR #4056, opened 2026-09-22
Create product sets during onboarding instead of waiting for the heartbeat

vahidkay-meta:
## Description

Onboarding runs the product set sync inline with an empty catalog ID, so the sets it should create are never created — two failing Graph calls per category instead, and the sets only appear when a later daily heartbeat runs.

## Steps to reproduce

1. Disconnect (or Reset settings) and connect again through the Shops tab
2. `WooCommerce > Status > Logs > facebook_for_woocommerce` — requests to `graph.facebook.com/v21.0//product_sets` (empty catalog segment) return 400 `Unsupported get request`
3. `Commerce Manager > Catalog > Sets` — no new sets until the daily sync runs

## Cause

`WC_Facebookcommerce_Integration::get_product_catalog_id()` memoizes on first read:

```php
if ( ! is_string( $this->product_catalog_id ) ) {
    $value                    = get_option( self::OPTION_PRODUCT_CATALOG_ID, '' );
    $this->product_catalog_id = is_string( $value ) ? $value : '';
}
```

`apply_installation_settings()` reads it to decide whether the update warrants a sync — pinning `''` on a first connection — and then `update_settings()` stores the new ID with a plain `update_option()`, which never touches that in-memory copy. `maybe_trigger_feed_uploads()` runs the product set sync in the same request, `ProductSetSync` reads the catalog ID back through the integration, and gets the stale empty string.

Same code in 3.7.6 (`settings/update`); 3.7.7 moves onboarding to the finalize-install route, which inherits it. The product sync itself is unaffected — it is a background job, so it picks up the stored value on a later request.

## The fix

**1. Keep the stored row and the in-memory copy in step.** `update_settings()` routes the catalog ID through `update_product_catalog_id()` instead of writing the option directly, so anything running later in the request sees the catalog the merchant just connected.

**2. Do not spend the day's sync on a run that cannot do anything.** `sync_all_product_sets()` returns before setting its 24 hour transient when no catalog is known.

This second part is what made the bug look intermittent — "only on the first onboarding of the day". The failing inline run set the transient on its way in, so the heartbeat that would otherwise have created the sets skipped for the rest of the window. With the guard, a catalog arriving later in the day is picked up by the next heartbeat rather than waiting it out. It is also a backstop against any other path reaching this with no catalog.

## Testing

- `HandlerCatalogIdTest` (new) — after `handle_update()` the integration reports the new catalog ID even though the old one was read first; a recording stand-in for the sync handler asserts the value visible **at sync time**; and an unchanged catalog ID still does not trigger a sync.
- `ProductSetSyncTest` — two cases for the guard: no Graph work is attempted without a catalog, and the daily transient is left unspent so the next run proceeds once a catalog exists. The existing cases now set a catalog in `setUp()`, which is the normal state for product set sync anyway.


ukilla:
The fix works on **WP 7.1.1/WC 11.1.1**, but the main issue is the performance overhead:

**1. The sync now does work inside the finalize-install request.** Before, the loop failed fast. Now it's two Graph calls per category, inline: 8 s → 13 s here, tested with 22 categories, minutes for a store with a few hundred categories. The token and the install are already committed on Meta's side before the loop runs, so if a host or proxy cuts the request, the store is connected server-side while the browser lands in the `.catch` in `Shops.php`, which only logs to the console. The merchant keeps seeing the onboarding splash. The product sync already avoids this by going to a background job.

**2. It works around the memo instead of removing it.** `wc_facebook_product_catalog_id` is autoloaded, so `get_option()` is already served from the options cache and the memo on the integration buys nothing. Keeping it and special-casing one key in `update_settings()` leaves the other writers inconsistent (`map_params_to_options()` writes the pixel ID with a direct `update_option()`, `clear_integration_options()` deletes the catalog row without touching the memo).

**Suggested shape of the fix**

Maybe we should drop the memo, so every writer stays in sync and the special case in `update_settings()` isn't needed:

```php
public function get_product_catalog_id() {
	$value = get_option( self::OPTION_PRODUCT_CATALOG_ID, '' );
	return apply_filters( 'wc_facebook_product_catalog_id', is_string( $value ) ? $value : '', $this );
}
```

And queue the set sync instead of running it inline. Action Scheduler is already in use for the heartbeat:

```php
as_enqueue_async_action( 'facebook_for_woocommerce_sync_product_sets', array(), 'facebook-for-woocommerce', true );
```

with the hook bound to `sync_all_product_sets()`. The 24 h guard keeps working inside the job, onboarding returns in about a second, and a fresh request sidesteps the memo problem on its own.

Issues with the tests: the guard only covers `sync_all_product_sets()`, the category create/edit/delete hooks still call Graph directly with whatever catalog is stored; and the new tests drive `handle_update()` rather than `handle_finalize_install()`, where the 3.7.7 path actually lives (the finalize client is injectable via the constructor, so a stub would cover it).

